PR #104, opened 2018-06-03
Protocol Paused

### LINKED ISSUE #97: Dynamically detect protocol is paused, and put up "protocol paused" modal
Currently the protocol explorer doesn't know when the protocol is paused.  When users use the explorer during protocol pause, all transactions fail.

Not able to detect paused protocol

Protocol explorer can detect when the protocol is paused, warn users, and stop them from doing actions that will trigger a transaction.

ericxtang:
Add a check for protocol paused status, and put up a modal if it's paused.  

* Add an API call to get protocol status
* Add graphql schema and query to get the status
* Add a view for the modal

Used mockProtocol to turn on/off the modal

Closes https://github.com/livepeer/livepeerjs/issues/97

**Checklist:**


packages/explorer/src/views/ImportantMessage/index.js
```diff
@@ -1,18 +1,21 @@
...
-        href="https://forum.livepeer.org/t/protocol-paused-for-bug-fix-upgrade-5-26-18-9-00pm-edt/265"
-        target="_blank"
-      >
-        forum post
-      </a>
-    </p>
-  </BasicModal>
-)
+const ImportantMessage = ({ protocol }) => {
+  console.log('Protocol paused: ' + protocol.data.paused)
```
jozanza: Let's remove this

packages/graphql-sdk/src/types/Protocol.js
```diff
@@ -0,0 +1,11 @@
+const Protocol = `
+    type Protocol {
+        "Protocol paused"
+        paused: Boolean!
+
+        "Protocol id"
+        id: String!
+    }
```
jozanza: The formatting in this block is using 3 or 4 space indents. Let's make sure this type def is using 2 spaces for indentation. It would be helpful to add a comment on the line above the type declaration as well.

jozanza (changes requested):
Looks really good. I think we'll just want to add a couple tests to cover these updates. We need to

- [ ] Add a couple unit tests in `packages/graphql-sdk/src/resolvers/Protocol.test.js` for testing the Protocol type's resolvers.
- [ ] Add a snapshot test in `packages/graphql-sdk/src/index.test.js` for "ProtocolQuery"

The snapshot test will probably look something like this:
```js
/**
 * Account
 */
test('ProtocolQuery', async t => {
  const res = await graphql(schema, ProtocolQuery, null, { livepeer }, {})
  t.snapshot(res)
})
```
You'll need to also add a `getProtocolPaused()` method somewhere in the `rpc` namespace of the mock-sdk (`packages/graphql-sdk/src/mock-sdk.js`)

And last, since you added a new method to the sdk, we should update the generated documentation. Just run `yarn docs` from the project root and add/commit all the updated files.

PR #113, opened 2018-06-14
Player: ENS based channel names

### LINKED ISSUE #110: Broadcaster channel name

I am unable to configure a human-readable name as the Broadcaster in media.livepeer.org, which means that my viewers are unable to build awareness of my name and brand, as a broadcaster.

When a viewer loads up a stream, e.g. [Livepeer TV](https://media.livepeer.org/channels/0x0dDB225031cCB58fF42866f82D907F7766899014), they are told that the Broadcaster is `0x0dDB225031cCB58fF42866f82D907F7766899014`. My conjecture is that this information has limited meaning or relevance to a viewer.

See below.

image

**Do you have any personal experiences dealing with this problem? (optional)**

I have the problem that I don't feel like my viewers "get it" as well as they might.

Create a mechanism to store a "channel name" associated with the broadcaster address, with this information signed and stored in the Blockchain.

It's unlikely that it will change often, although the solution ought to allow unlimited changes by the key holder(s).

Then, when the channel is loaded by the viewer, it should show the "human-readable" name instead of the hex code. Like so:

image

Keep up the good work, it's very nice.

yondonfu:

Allows a channel to either be an ENS name OR an ETH address. If the channel string is not an ETH address, the player will treat it as an ENS name and try to look up the corresponding ETH address to be used to fetch jobs. If the channel string is an ETH address, the player will just fetch the jobs.

If the channel string is a valid ENS name, the player will display both the ENS name and the corresponding ETH address.

- Add a method in `@livepeerjs/sdk` that looks up the ETH address for a provided ENS name
- Updated the schema in `@livepeerjs/graphql-sdk `to add a ENSName type which stores an ENS name and has a nested Account type. Introduced ENSNameQuery which will fetch an Account associated with an ENS name and ENSNameBroadcasterQuery which will fetch a Broadcaster that can include a list of jobs associated with an ENS name
- Update the Channel view in `@livepeerjs/player` to use ENSNameBroadcasterQuery to fetch an associated ETH address and its jobs for an ENS name if the channel string is not an ETH address. If the channel string is an ETH address, the player uses the same JobsQuery that it is currently using

A detailed description of how you tested your code changes. Include details of your testing environment, and the tests you ran to see how your change affects other areas of the code, etc.

I updated the tests in `@livepeerjs/graphql-sdk` for the newly introduced queries.

I registered an ENS name `appleapple.test` on Rinkeby and tested using the player using the url `http://localhost:3000/channels/appleaple.test` to observe whether if the player would not only display the ENS name and associated ETH address, but also if the player would continue to play live streams associated with the ETH address.

Fixes #110

**Screenshots (optional):**
screen shot 2018-06-14 at 4 12 37 pm

**Checklist:**


yondonfu:
Updated the GraphQL schema
- Removed the ENSName type
- Added an optional `ensName: String` field to the `Account` type
- Updated the `account` resolver to accept either an `id` (ETH address) and `ensName` - if `ensName` is used, then it will attempt to lookup the address for the name
- Using a single `AccountBroadcasterQuery` in the player Channel view which will pass in the `ensName` parameter if the URL query parameter is not an ETH address and will pass in the `id` parameter if the URL query parameter is an ETH address

PR #154, opened 2018-07-17
Fix hardcoded MerkleMine proof size for transaction data in explorer

yondonfu:

Properly construct the ABI encoded Merkle proof size for the transaction data used to generate tokens.

- Instead of using a hard coded Merkle proof size (currently 704 bytes), construct the ABI encoded Merkle proof size based on the provided hex encoded proof by taking the proof size and left padding with zeros until we have 32 bytes

I ran the explorer locally and hardcoded the variable `defaultAddress` here https://github.com/livepeer/livepeerjs/blob/f3094301de6e1cf0f203c8affa0731b06ad8a605/packages/explorer/src/views/Mining/index.js#L24 to 0xd3bbe001d3044e78b068244aae95bafe75744233 which has an associated proof of size 640 which is different from the currently hardcoded 704. I logged the transaction data created when clicking "Claim Token" while using this default address to observe that the proof size was properly encoded in the transaction data.

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/Mining/enhance.js
+++ b/packages/explorer/src/views/Mining/enhance.js
@@ -6,6 +6,21 @@ import {
   connectToasts,
 } from '../../enhancers'
 
+const encodeProofSize = proof => {
+  // Assume hex encoded proof
+  const proofSize = proof.length / 2
+
+  let res = proofSize.toString('16')
+  let len = res.length
+
+  while (len < 64) {
+    res = '0' + res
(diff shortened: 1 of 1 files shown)


dob (approved):
Looks good to me. @ericxtang can you give a sanity check look, and then if good merge and deploy this fix to the explorer which will fix the merkleMine page?

ericxtang:
🚢

PR #164, opened 2018-07-31
fix blank page issue in mist browser

### LINKED ISSUE #155: Explorer not loading in Mist browser - displaying blank page
The explorer is not loading in Ethereum's Mist browser. It just shows a blank page.

I expected the same page that you see in Chrome or Firefox at https://explorer.livepeer.org/token or https://explorer.livepeer.org/transcoders to load. However the page just remains blank.

**To Reproduce (required)**
1. Launch Ethereum's Mist browser
2. Click on the search bar browse icon
3. Type https://explorer.livepeer.org/token
4. Wait for page to load but observe blank screen

**Device (required):**
 - Device: Macbook pro
 - OS: OS X
 - Browser Mist
 - Version 0.11.0

 
**Screenshots (optional)**
screen shot 2018-07-20 at 12 43 46 pm

PumpkingWok:

I have edited two js files for resolve the issue of get a blank webpage on mist browser, for example https://explorer.livepeer.org/token or https://explorer.livepeer.org/transcoders.

The issue was caused by native web3 command in mist (window.web3), because Mist does not ship its own version of web3.js in browser anymore since 0.9 version.In the future mist will provide a new ethereum provider i think.

In the code, i have changed from window.web3 to window.web3.version because in this way the control fails and the browser always will use infura. The issue with checking only on window.web3 is that it is exists in native mist browser but provide only an object (CurrentProvider) without attributes like version or version.network and for this reason the browser obtain a blank page because it is breaks when it tries to access to 'network' parament in window.web3.version, window.web3.version is undefined in this case.

- The webpage is loaded correctly now on Mist browser.
- With these changes, Mist will use web3 injection provided by infura like other browsers without metamask installed.

I have tested this in the same environment requested in Fix #155 , OS X and Mist 0.11.

Yes, Fix #155 issue.

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/explorer/src/components/BasicNavbar.js
+++ b/packages/explorer/src/components/BasicNavbar.js
@@ -24,7 +24,7 @@ const BasicNavbar = ({ onSearch, currentRound, toasts, coinbase, history }) => {
   const myAccountAddress = coinbase.data.coinbase
   const notAuthenticated = !myAccountAddress
   const { host } = window.livepeer.config.eth.currentProvider
-  const networkName = !window.web3
+  const networkName = !window.web3.version
     ? /infura/.test(host)
       ? host.split('.')[0].replace(/(http|https):\/\//, '')
       : 'Custom RPC'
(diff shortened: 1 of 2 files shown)


ericxtang:
Thanks @PumpkingWok!  We'll look into this soon.

PR #169, opened 2018-08-02
LIP-8 Interface Support

### LINKED ISSUE #147: Add support for new unbonding interface in the explorer

[LIP-8](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-8.md) introduces the concept of partial unbonding to the protocol allowing delegators to unbond a portion of their bonded stake which can be withdrawn after the unbonding period. These unbonding tokens are managed by an "unbonding lock" structure which can also be used to rebond towards a delegator's current delegate prior to the unbonding lock's withdraw round.

Adding full support for partial unbonding within the explorer will require some upfront UI/UX design work to determine the optimal way to present the partial unbonding workflow to end users. However, as a first step before tackling the UI/UX design, we can add support for the new unbonding interface of an upgraded BondingManager contract within the SDK so the contract can be upgraded first without breaking the existing explorer experience.

The SDK currently uses the old unbonding interface where a user can call the `unbond()` and `withdrawStake()` contract functions without any parameters.

**SDK**

Remove references to a delegator's `Unbonding` state.

Remove references to a delegator's `withdrawRound`.

Add the SDK `getUnbondingLocks()` method
- Stub for future implementation with support for all of a delegator's unbonding locks
- For now, get the delegator's `nextUnbondingLockId` and if there is an active unbonding lock with ID `nextUnbondingLockId - 1` return it, else return null.

Update the SDK `unbond()` method
- Query the BondingManager for the account's current bonded amount
- Call `BondingManager.unbond()` with the account's current bonded amount

Update the SDK `withdrawStake()` method
- Call the SDK `getUnbondingLocks()` method. If the result is null, throw
- If the returned unbonding lock's withdraw round < current round, throw
- Use the unbonding lock ID returned by `getUnbondingLocks()` to call `BondingManager.withdrawStake()`

**Explorer**

In the delegator card view, if the SDK `getUnbondingLocks()` returns a valid unbonding lock, display the delegator's status as `Unbonding` if the lock's withdraw round < current round, else display the delegator's status as `Unbonded`. Use the lock's withdraw round to display the delegator's withdraw round.

Note: These changes do not handle the case where an account is used in another context outside of the explorer. 

This is just a first step towards supporting the partial unbonding workflow within the explorer by allowing the SDK to use the new unbonding interface of an upgraded BondingManager contract.

yondonfu:

Adds support for the new BondingManager contract interface for partial unbonding introduced in [LIP-8](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-8.md).

Assume that a delegator's latest unbonding lock ID is `nextUnbondingLockId > 0 ? nextUnbondingLockId - 1 : 0`. For the time being until a full product design is created for partial unbonding, the explorer will only allow users to fully unbond (which is the current experience).

- **sdk**: Add the `getDelegatorUnbondingLock()` function.
- **sdk**: Update the `getDelegator()` function to also return `nextUnbondingLockId` and `withdrawAmount`. `withdrawAmount` is the amount stored by the delegator's latest unbonding lock. `withdrawRound` is now the withdraw round stored by the delegator's latest unbonding lock.  In order to hide the notion of partial unbonding and unbonding locks from the user for the time being, `getDelegator()` will return a status of `Unbonding` if the delegator's most recent unbonding lock has a withdraw round greater than the current round.
- **sdk**: Update the `unbond()` function to use the delegator's `bondedAmount` fully unbond to call `BondingManager.unbond(bondedAmount)`.
- **sdk**: Update the `withdrawStake()` function to use the delegator's latest unbonding lock ID to call `BondingManager.withdrawStake(latestUnbondingLockId)`.
- **sdk**: Update the `BondingManager` ABI in `sdk/etc/BondingManager.json`.
- **graphql-sdk**: Update the `Delegator` schema.
- **explorer**: Display the withdraw amount (if non-zero) for an unbonding/unbonded delegator

Note: These changes do not account for the scenario where users use a key outside of the explorer - if users directly interact with the contracts or use keys with the Livepeer node CLI, they will be able to partially unbond and create multiple unbonding locks. The explorer will then not be able to account for multiple unbonding locks. These changes are just meant to allow the explorer to still be usable in the interim by making the assumption that the user is just using the explorer and the hope is that a full partial unbonding experience later on will solve this issue.

- Updated **graphql-sdk** tests
- Updated **sdk** tests
- Manually tested the explorer changes with contracts deployed to Rinkeby

Fixes #147 

**Checklist:**


packages/sdk/src/index.js
```diff
@@ -1999,16 +2003,21 @@ export default async function createLivepeerSDK(
...
-      }
+      const {
+        status,
+        withdrawAmount,
+        nextUnbondingLockId,
+      } = await rpc.getDelegator(tx.from)
 
-      if (status != DELEGATOR_STATUS.Unbonding) {
-        throw new Error('Delegator is not in the unbonding state')
+      if (status != DELEGATOR_STATUS.Unbonded && withdrawAmount !== '0') {
```
Randy1Burrell: @yondonfu, this line might throw off the result if status and DELEGATOR_STATUS.Unbonded are not already the same data types. Why not `if (status !== DELEGATOR_STATUS.Unbonded && withdrawAmount !== '0')` instead?

ericxtang (approved):
🚢 after merge

PR #187, opened 2018-08-20
Check for web3 object before checking property on web3 object

### LINKED ISSUE #185: Explorer not loading on non web3 enabled browsers
The explorer is displaying a blank page (not loading at all) on mobile browsers and non web3 enabled desktop browsers. It is possibly because of this PR, which would throw an exception if web3 isn't injected into the page: https://github.com/livepeer/livepeerjs/pull/164/files

Explorer should load

**To Reproduce (required)**
1. Visit explorer.livepeer.org on mobile safari or brave

**Device (required):**
 - Reproduced on mobile iOS and desktop browsers.
 
**Additional context**
The above referenced PR got Livepeer working within the Mist browser, but I think it's more important to roll it back (or just undo the two line change) and fix this issue, prior to worrying about Mist support.

yondonfu:

Check for injected web3 object in the explorer before checking for property on the web3 object. At the moment, the explorer tries to check for a web3 object with the `version` property. See #164 for more details on the rationale before the current behavior in order to allow the explorer to be used within the Mist browser. The problem with the current implementation is that if a user navigates to the explorer with a non-web3 enabled browser (i.e. Chrome without Metamask), there will be no injected web3 object at all. The explorer will try to check for the `version` property for an undefined object resulting in an error. 

Now, the explorer will first check for the injected web3 object and only check for the `version` property if the web3 object is present.

- **explorer**: Check for injected web3 object before checking property on web3 object

Fixes #185 

**Checklist:**

### DIFF
--- a/packages/explorer/src/components/BasicNavbar.js
+++ b/packages/explorer/src/components/BasicNavbar.js
@@ -24,7 +24,12 @@ const BasicNavbar = ({ onSearch, currentRound, toasts, coinbase, history }) => {
   const myAccountAddress = coinbase.data.coinbase
   const notAuthenticated = !myAccountAddress
   const { host } = window.livepeer.config.eth.currentProvider
-  const networkName = !window.web3.version
+  // Test if web3 is injected
+  // For Mist compatability we also check if the web3 object has the `version` property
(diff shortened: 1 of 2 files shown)


dob (approved):
:ship:

PR #207, opened 2018-09-14
Use pending stake when fully unbonding

### LINKED ISSUE #204: sdk: Use pending stake return value when fully unbonding

The SDK currently implements full unbonding by querying for a delegator's bonded amount and then passing the value as a parameter to the `BondingManager.unbond()` [function](https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L1879). This will not actually fully unbond the delegator if it has unclaimed rewards for past rounds because `BondingManager.unbond()` will automatically claim those rewards to update the delegator's bonded amount before executing the unbonding logic.

The SDK should instead use the delegator's [pending stake](https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L1120) as the value passed as a parameter to the `BondingManager.unbond()` function.

The SDK passes the delegator's bonded amount as a parameter to the `BondingManager.unbond()` function.

yondonfu:

When fully unbonding, the SDK will use the delegator's pending stake as the parameter for `BondingManager.unbond()`.

- **sdk**: Modify `unbond()` to use the delegator's pending stake instead of bonded amount

I tested this change on Rinkeby.

Fixes #204 

**Checklist:**

### DIFF
--- a/packages/sdk/src/index.js
+++ b/packages/sdk/src/index.js
@@ -1870,13 +1870,13 @@ export default async function createLivepeerSDK(
      * // }
      */
     async unbond(tx = config.defaultTx): Promise<TxReceipt> {
-      const { status, bondedAmount } = await rpc.getDelegator(tx.from)
+      const { status, pendingStake } = await rpc.getDelegator(tx.from)
       // Can only unbond successfully when not already "Unbonded"
       if (status === DELEGATOR_STATUS.Unbonded) {
         throw new Error('This account is already unbonded.')
       } else {
         return await utils.getTxReceipt(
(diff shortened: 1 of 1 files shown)


Randy1Burrell (commented):
The changes looks good. What about the changes to the documentation?

yondonfu:
Oops, accidentally marked the box that indicates that documentation changes are required.

Though, the documentation will need to be changed once partial unbonding is introduced in the explorer - at the moment, the SDK only allows full unbonding which is the same as the previous behavior (so it shouldn't require any documentation changes yet).

PR #209, opened 2018-09-15
Add transcoder address to job info

### LINKED ISSUE #190: Add transcoder address to job info in SDK

The SDK function `getJob()` currently will not return a transcoder address in the result object if the transcoder did not already submit an on-chain claim to store the transcoder address on-chain. In this case, `getJob()` should call the contract function [electActiveTranscoder](https://github.com/livepeer/protocol/blob/master/contracts/bonding/BondingManager.sol#L651) in order to compute the assigned transcoder address to be returned with the result object.

The SDK function `getJob()` currently will not return a transcoder address in the result object if the transcoder did not already submit an on-chain claim to store the transcoder address on-chain.

Randy1Burrell:

The SDK function getJob() currently will not return a transcoder address in the result object if the transcoder did not already submit an on-chain claim to store the transcoder address on-chain. This PR calls the contract function [electActiveTranscoder](https://github.com/livepeer/protocol/blob/master/contracts/bonding/BondingManager.sol#L651) in order to compute the assigned transcoder address to be returned with the result object.

Fixes #190 
**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/sdk/src/index.js
+++ b/packages/sdk/src/index.js
@@ -1627,12 +1627,19 @@ export default async function createLivepeerSDK(
      */
     async getJob(id: string): Promise<Job> {
       const x = await JobsManager.getJob(id)
+      const { hash } = await rpc.getBlock(x.creationBlock)
+      const transcoder = await BondingManager.electActiveTranscoder(
+        x.maxPricePerSegment,
+        hash,
+        x.creationRound,
+      )
       return {
         id: toString(id),
         streamId: x.streamId,
         transcodingOptions: utils.parseTranscodingOptions(x.transcodingOptions),
         transcoder: x.transcoderAddress,
         broadcaster: x.broadcasterAddress,
+        transcoder: transcoder,
       }
     },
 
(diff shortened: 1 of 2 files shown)


yondonfu (commented):
Change looks good. Should we add a test in `index.test.js` to check that the shape of the return object from `getJob()` matches the expected schema?

Randy1Burrell:
@yondonfu good suggestion. I just added the test.

packages/sdk/src/index.js
```diff
@@ -1627,6 +1627,15 @@ export default async function createLivepeerSDK(
      */
     async getJob(id: string): Promise<Job> {
       const x = await JobsManager.getJob(id)
+      // Computer transcoder's address if none was returned
+      if (!x.transcoderAddress) {
```
yondonfu: If the job does not yet have a transcoder address attached to it `x.transcoderAddress` will actually be the null/empty address 0x0000000000000000000000000000000000000000

So we could change this check to `x.transcoderAddress === EMPTY_ADDRESS` (where EMPTY_ADDRESS is a constant already defined within the SDK)

yondonfu (changes requested):

packages/sdk/src/index.js
Randy1Burrell: @yondonfu I made the change.

yondonfu (approved):
🚢

PR #215, opened 2018-09-24
fix #206 , @livepeer/sdk should work out of the box with both import and require

### LINKED ISSUE #206: `@livepeer/sdk` works with import but not require out of the box
```javascript
// works
import LivepeerSDK from '@livepeer/sdk'
console.log(typeof LivepeerSDK === 'function') // true

// doesn't work
const { LivepeerSDK } = require('@livepeer/sdk')
console.log(typeof LivepeerSDK === 'function') // false
```

this happens because the function in question is actually called `createLivepeerSDK` but exported this way

```javascript
export default async function createLivepeerSDK(
  opts: LivepeerSDKOptions,
): Promise<LivepeerSDK> {
```

this is apparently not picked up by `require` , so in order to get it to work now we gotta use this

```javascript
const LivepeerSDK = require('@livepeer/sdk').default // works fine
```

both `require` and `import` should work out of the box

**To Reproduce (required)**
1. run code in the description section.

**Device (required):**
 - Device: Laptop (browser, webpacked)
 - OS: Ubuntu
 - Browser : Chrome, firefox

**Possible enhancement**

- rename `createLivepeerSDK` to `LivepeerSDK` to avoid confusion
- or mention this explicitly in the docs.

ya7ya:
this exports `LivepeerSDK` and `createLivepeerSDK` so we can support import and require out of the box.

Fixes #206

(this has backward compatibility so we don't break applications out there, thanks to @Randy1Burrell for pointing that out :+1: )

### DIFF
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -12,9 +12,9 @@ A module for interacting with Livepeer's smart contracts. The SDK is a core depe
 
 ## Table of Contents
 
-* [Installation](#installation)
-* [Usage](#usage)
-* [API](#api)
+- [Installation](#installation)
+- [Usage](#usage)
+- [API](#api)
 
 <!-- hide-on-docup-stop -->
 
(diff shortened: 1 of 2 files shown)


Randy1Burrell (approved):
LGTM! Great work @ya7ya!

PR #219, opened 2018-09-26
Update state requirements for withdrawal

yondonfu:

Update state requirements for withdrawal.

This PR is in response to a bug reported by users where they unbonded, waited through the unbonding period, but also bonded more tokens. Then after the unbonding period completes, the user should be able to withdraw the unbonded tokens, but is unable to do so because at the moment there is a requirement at SDK/UI level that the user has to be unbonded in order withdraw. Since the user bonded more tokens it would not be in the unbonded state after the unbonding period completes which blocks the withdrawal.

- **explorer**: Instead of requiring the user to be in the unbonded state to withdraw, require the user to not be in the unbonding state to withdraw. This should cover the scenario described above.
- **sdk**: Update error message handling for `withdrawStake()`

Note: There is no longer an explicit unbonding state in the protocol. In the SDK/UI, the unbonding state is maintained however - a delegator is considered as being in the unbonding state here if its unbonding lock with the ID `nextUnbondingLockId - 1` has a non-zero withdraw round and the current round is less than that value.

Tested locally.

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/Account/index.js
+++ b/packages/explorer/src/views/Account/index.js
@@ -52,7 +52,6 @@ const AccountView: React.ComponentType<AccountViewProps> = ({
   const { delegateAddress, status } = me.data.delegator
   const isBonded = status === 'Bonded'
   const isBonding = status === 'Pending'
-  const isUnbonded = status === 'Unbonded'
   const isMyDelegate = accountAddress === delegateAddress
   const canRebond = isMyDelegate && (isBonded || isBonding)
   const canBond = !!userAddress
--- a/packages/explorer/src/views/AccountDelegating/index.js
+++ b/packages/explorer/src/views/AccountDelegating/index.js
@@ -70,13 +70,12 @@ const AccountDelegating: React.ComponentType<AccountDelegatingProps> = ({
   const earnedFees = hasUnclaimedRounds ? MathBN.sub(pendingFees, fees) : '0'
   const hasFees = fees !== '0'
   const isUnbonding = status === 'Unbonding'
-  const isUnbonded = status === 'Unbonded'
   const isBonding = status === 'Pending'
   const isBonded = status === 'Bonded'
   const roundsUntilUnbonded = isUnbonding
     ? MathBN.sub(withdrawRound, lastInitializedRound)
     : ''
-  const canWithdraw = isUnbonded && withdrawAmount !== '0'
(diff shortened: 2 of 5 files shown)


packages/explorer/src/views/Account/index.js
```diff
@@ -52,7 +52,6 @@ const AccountView: React.ComponentType<AccountViewProps> = ({
   const { delegateAddress, status } = me.data.delegator
   const isBonded = status === 'Bonded'
   const isBonding = status === 'Pending'
-  const isUnbonded = status === 'Unbonded'
```
yondonfu, line 55: Variable not used

packages/explorer/src/views/AccountDelegating/index.js
```diff
@@ -70,13 +70,12 @@ const AccountDelegating: React.ComponentType<AccountDelegatingProps> = ({
   const earnedFees = hasUnclaimedRounds ? MathBN.sub(pendingFees, fees) : '0'
   const hasFees = fees !== '0'
   const isUnbonding = status === 'Unbonding'
-  const isUnbonded = status === 'Unbonded'
```
yondonfu, line 73: Variable not used

dob (commented):
Ok, so the resulting behavior here would be that a user can withdraw if `nextUnbondingLockId - 1`'s withdrawBlock is in the past?

Though if they believe they've waited the 7 day waiting period of their first unbondingLock, but have a newer outstanding one, they still will run into a weird state, right? (I know we need much deeper UI updates to handle multiple unbondingLocks).

yondonfu:
> Ok, so the resulting behavior here would be that a user can withdraw if nextUnbondingLockId - 1's withdrawBlock is in the past?

That is already the current behavior. The change this PR makes is that now if a user's `nextUnbondingLockId - 1` unbonding lock is valid (non-zero withdraw block) and its withdraw block is in the past, the user can use it to withdraw even if the user is not in the unbonded state (i.e. if the user bonded more tokens before withdrawing). Previously, the user could only withdraw if the user was in the unbonded state.

> Though if they believe they've waited the 7 day waiting period of their first unbondingLock, but have a newer outstanding one, they still will run into a weird state, right? 

Yes - the explorer needs more involved UI updates to support this type of management.

dob (approved):
Got it. Thanks!

PR #227, opened 2018-10-01
[READY] livepeer/player: specify stream source in url

### LINKED ISSUE #182: Specify Stream Source via URL
Adding a 'special' channel requires a codebase change. Maintaining this list of hard-coded addresses is unwieldy, especially now that we are pulling directly from the broadcaster rather than the p2p network. This makes it difficult for folks to experiment with broadcasting on their own nodes without also running a livepeerjs component and making address changes.

We maintain a list of hard coded broadcaster (CDN) addresses to pull from.

Specify the broadcaster or CDN URL via the path or query string. For example, `media.livepeer.org/ethaddress?source=http://cdn.address` . When the latest stream ID is looked up, the final playback URL becomes `http://cdn.address/streamid.m3u8`

For long running jobs, maybe we could also specify a full path such as `media.livepeer.org?fullsource=http://cdn.address/streamid.m3u8` which would bypass the blockchain lookup and improve responsiveness.

ya7ya:
this should allow the user to add `source` query to their player url to specify the stream source. this should close #182 

- detect query string and parse it.
- generate a proper m3u8 url using the stream source + manifestId

I ran a local lp node, and used specified an external source (cdn) . it works out of the box.

fixes #182 

**Checklist:**

### DIFF
--- a/packages/player/src/views/Channel/index.js
+++ b/packages/player/src/views/Channel/index.js
@@ -172,34 +172,34 @@ class Channel extends Component {
     }
 
     if (address === process.env.REACT_APP_LIVEPEER_TV_ADDRESS.toLowerCase()) {
-      return this.setState({
+      this.setState({
         live: true,
         url: `${
           process.env.REACT_APP_LIVEPEER_TV_STREAM_ROOT_URL
         }/${manifestId}.m3u8`,
       })
-    }
-    if (
+    } else if (
       address === process.env.REACT_APP_CRYPTO_LIVEPEER_TV_ADDRESS.toLowerCase()
     ) {
-      return this.setState({
+      this.setState({
         live: true,
         url: `${
           process.env.REACT_APP_CRYPTO_LIVEPEER_TV_STREAM_ROOT_URL
         }/${manifestId}.m3u8`,
       })
-    }
-    if (address === process.env.REACT_APP_INGEST2_ADDRESS.toLowerCase()) {
-      return this.setState({
+    } else if (
+      address === process.env.REACT_APP_INGEST2_ADDRESS.toLowerCase()
+    ) {
+      this.setState({
         live: true,
         url: `${
           process.env.REACT_APP_INGEST2_STREAM_ROOT_URL
         }/${manifestId}.m3u8`,
       })
+    } else {
+      let url = `${process.env.REACT_APP_STREAM_ROOT_URL}/${manifestId}.m3u8`
+      this.setState({ url })
     }
-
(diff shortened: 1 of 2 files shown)


packages/player/src/views/Channel/index.js
```diff
@@ -145,45 +150,61 @@ class Channel extends Component {
 
   async componentWillReceiveProps(nextProps) {
     const { data } = nextProps.account
+    const location = nextProps.location
     const address = data.id.toLowerCase()
     const [latestJob] = data.broadcaster.jobs
+    const manifestId = latestJob.streamId.substr(0, 68 + 64)
```
Randy1Burrell, line 156: Why not use a constant here, rather than adding two numbers? 
Maybe this could work:

```
const manifestIdLen = 132
const manifestId = latestJob.streamId.substr(0, manifestIdLen)

```

or

```
const manifestId = latestJob.streamId.substr(0, 132)
```
Maybe it's my OCD talking but IMO the first example can improve code readability.

packages/player/src/views/Channel/index.js
```diff
@@ -145,45 +150,61 @@ class Channel extends Component {
...
       if (nextProps.loading === false) this.setState({ live: false })
       return
     }
-    const manifestId = latestJob.streamId.substr(0, 68 + 64)
+
+    if (location.search) {
+      let queryObject = parseQs(location.search)
+      if (queryObject && queryObject.source) {
+        console.log('using source qs')
+        return this.setState({
```
Randy1Burrell, line 167: # Just for clarity here. 

I'm guessing a return statement was added here, in case there are actual environment variables set it doesn't mess with the stream location pulled from the query string. But why remove the rest of the return statements even though they don't affect the app?

> @ya7ya Please note. The code is correct, however, I just want to get an understanding of your programming style.

ya7ya: @Randy1Burrell , I removed the rest of the return statements and exchanged them for `else if` because of codeclimate recommendation. 

When i added that return the codeclimate test failed with "too many return statements" so I got rid of it.

packages/player/src/views/Channel/index.js
ya7ya: @Randy1Burrell That line was here before me 😄 so i didn't technically change it , i think the `64` is the signature length. whoever coded it (git blame :D ) was probably making a point of "add signature length 64 to the actual hash length" .

Randy1Burrell: @ya7ya thanks for clearing this up.

Randy1Burrell (approved):
Good to go!!

PR #232, opened 2018-10-02
Some proposed changes to Livepeer Explorer landing

chrishobcroft:
Removed the video tutorial, edit some copy, changed some routing.

It makes some changes to explorer.livepeer.org landing page, to help new explorers in Livepeer to begin to engage with the protocol more easily.

Specifically, it promotes the section where a user can enter any Ethereum address (including their own), to begin to explore livepeer's protocol.

It also rewords some copy, to make it more accessible to a wider audience - the 2.6 million who have no idea about bonding and inflationary reward yet - but they will soon, maybe!

- Remove the video (for now, pending a new video to be put in it's place to help new entrants joined the community)

- Rebrand "Explorer Protocol Activity" to "Explore Your Account"

   - Adapt the copy to make it friendlier, and easier to engage with. Also, to seek to personalise it somewhat, and make the user feel at the centre of this.

- Rebrand "Play the Delegation Game" to "BOND your LPT to a Node"

   - Someone should totally gamify this, but not us. At least, not yet.
   - For now, I propose to lead them down the garden path, to the specific call-to-action that we want them to take. To BOND.
   - GET TOKEN is finished now. Remove from everywhere, let's tidy this up :)

I tested them in my mind's eye, tweaked some text about, fit @jozanza's pattern. It'll be fine. Can we push them to staging first and then have a look?

No.

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/Landing/index.js
+++ b/packages/explorer/src/views/Landing/index.js
@@ -22,19 +22,15 @@ const Landing = ({ history, ...props }) => (
         <h3 style={{ letterSpacing: 8 }}>Protocol Explorer</h3>
         <br />
         <br />
-        <Video src="/static/media/landing.mp4" />
-        <br />
-        <br />
         <div style={{ display: 'flex' }}>
           <InfoBox>
             <h2>
               <Icon use="explore" />&nbsp;
-              {`Explore Protocol Activity`}
+              {`Explore Your Account`}
             </h2>
             <p>
-              The account view shows the recent Livepeer smart contract
-              transactions for any Ethereum address. It also gives you deep
-              insight into important protocol metrics.
+              Check your balance of Livepeer Tokens (LPT),
+              and learn what you can do with them.
             </p>
             <br />
             <form
@@ -50,7 +46,7 @@ const Landing = ({ history, ...props }) => (
                 name="address"
                 type="search"
                 pattern="^0x[a-fA-F0-9]{40}$"
-                placeholder="Enter an ETH account address"
+                placeholder="Enter your ETH account address"
                 onKeyDown={e => {
                   if (e.keyCode !== 13 || !e.target.value) return
                   document.getElementById('account-search-button').click()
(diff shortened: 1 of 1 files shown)


packages/explorer/src/views/Landing/index.js
```diff
@@ -22,19 +22,15 @@ const Landing = ({ history, ...props }) => (
...
         <br />
-        <Video src="/static/media/landing.mp4" />
-        <br />
-        <br />
         <div style={{ display: 'flex' }}>
           <InfoBox>
             <h2>
               <Icon use="explore" />&nbsp;
-              {`Explore Protocol Activity`}
+              {`Explore Your Account`}
```
Randy1Burrell, line 29: @chrishobcroft Great job. Just a few comments/questions.

> Wouldn't it be better to say "Explore an account" ?
- Reasoning: A user will know that they can also view others accounts. Ex. friends' account number they have and might be able to inform that friend of current unbonded tokens they don't know about.
- The description would make it clear that logging into a web3 enabled wallet is not mandatory for viewing an account.

packages/explorer/src/views/Landing/index.js
```diff
@@ -50,7 +46,7 @@ const Landing = ({ history, ...props }) => (
                 name="address"
                 type="search"
                 pattern="^0x[a-fA-F0-9]{40}$"
-                placeholder="Enter an ETH account address"
+                placeholder="Enter your ETH account address"
```
Randy1Burrell, line 49: @chrishobcroft same question as above.

packages/explorer/src/views/Landing/index.js
chrishobcroft: Hmm. Well. Good question @Randy1Burrell.

When I think about priorities, I compare the following objectives:

**- Objective 1:** Helping a token holder become aware that they have LPT

vs.

**- Objective 2:** Helping a token holder to check whether their friend has LPT

For me, Objective 1 is higher priority, as it is the shortest route for any keyholder to check whether they have LPT themselves.

Objective 2 can also be met using this process, but is more likely to happen once a token holder has established that _they_ have LPT in the first place, and hence should be lower priority.

What do you think?

packages/explorer/src/views/Landing/index.js
chrishobcroft: Same answer as above.

Randy1Burrell (approved):

Randy1Burrell (approved):
Merging this but it will be overwritten as soon as frontend-modification features are merged.

Randy1Burrell:
Thanks @chrishobcroft for contributing this.

PR #239, opened 2018-10-08
[READY] Added method to estimate gas in sdk

### LINKED ISSUE #174: Use eth_estimateGas RPC call to set default gas limit for transactions

Instead of hardcoding default gas limits within the graphql-sdk, we can use the `eth_estimateGas` RPC call to set the provided gas for transactions (users will still always be able to modify the gas limit using their wallet UI).

Ethereum transactions submitted from the explorer currently leverage graphql-sdk mutations which invoke the relevant contract functions defined in the SDK. The graphql-sdk hardcodes the default gas limit provided for many transactions (ex. https://github.com/livepeer/livepeerjs/blob/master/packages/graphql-sdk/src/resolvers/Mutation.js#L55). While this works fine in most cases, in the event that the gas usage increases for a particular contract function changes, the hardcoded default gas limit would need to be modified as well. An example of this happening in practice is #168.

Randy1Burrell:

Enables users to estimate gas to be used to run LP SDK methods on the EVM. 

- Method for getting amount of gas to be used for running a method
- 

- Added automated test to test return value of changes

Fixes #174 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/graphql-sdk/src/resolvers/Mutation.js
+++ b/packages/graphql-sdk/src/resolvers/Mutation.js
@@ -4,6 +4,7 @@ import {
   TransactionSubmitted,
 } from './Subscription'
 
+const BN = require('bn.js')
 /** Typedefs */
 
 type GQLContext = {
@@ -80,7 +81,7 @@ export async function claimEarnings(
   const { utils, config } = ctx.livepeer
   const { eth } = config
   const { endRound } = args
-  const gas = await ctx.livepeer.rpc.estimategas(
+  const gas = await ctx.livepeer.rpc.estimateGas(
     'BondingManager',
     'claimEarnings',
     [endround],
@@ -159,8 +160,16 @@ export async function unbond(
   args,
   ctx: GQLContext,
 ): Promise<TxReceipt> {
-  const gas = await ctx.livepeer.rpc.estimategas('BondingManager', 'unbond', [
-    10,
+  const { pendingStake, bondedAmount } = await ctx.livepeer.rpc.getDelegator(
+    ctx.config.defaultTx.from,
+  )
+  const totalStake =
+    new BN(pendingStake).cmp(new BN(bondedAmount)) < 0
+      ? bondedAmount
+      : pendingStake
+
+  const gas = await ctx.livepeer.rpc.estimateGas('BondingManager', 'unbond', [
+    totalStake,
   ])
   return await ctx.livepeer.rpc.unbond({
     ...ctx.config.defaultTx,
--- a/packages/graphql-sdk/src/resolvers/Mutation.js
+++ b/packages/graphql-sdk/src/resolvers/Mutation.js
@@ -4,9 +4,6 @@ import {
   TransactionSubmitted,
 } from './Subscription'
 
-/** Constants **/
-const gasRate = 1.5
-
 /** Typedefs */
 
 type GQLContext = {
@@ -33,12 +30,11 @@ export async function approve(
   const { type, amount } = args
   switch (type) {
     case 'bond':
-      const gas =
-        (await ctx.livepeer.rpc.estimateGas(
-          'BondingManager',
-          'approveTokenBondAmount',
-          [amount],
-        )) * gasRate
+      const gas = await ctx.livepeer.rpc.estimateGas(
+        'LivepeerToken',
+        'approve',
+        [ctx.livepeer.config.defaultTx.from, amount],
+      )
       return await ctx.livepeer.rpc.approveTokenBondAmount(amount, {
         gas: gas,
       })
@@ -61,12 +57,10 @@ export async function bond(
   ctx: GQLContext,
 ): Promise<TxReceipt> {
   const { to, amount } = args
-  const gas =
-    (await ctx.livepeer.rpc.estimateGas(
-      'BondingManager',
-      'bondApprovedTokenAmount',
-      [amount, to],
-    )) * gasRate
+  const gas = await ctx.livepeer.rpc.estimateGas('BondingManager', 'bond', [
+    amount,
+    to,
+  ])
   return await ctx.livepeer.rpc.bondApprovedTokenAmount(to, amount, {
     gas: gas,
   })
--- a/packages/graphql-sdk/src/resolvers/Mutation.js
+++ b/packages/graphql-sdk/src/resolvers/Mutation.js
@@ -4,6 +4,9 @@ import {
   TransactionSubmitted,
 } from './Subscription'
 
+/** Constants **/
+const gasRate = 1.5
+
 /** Typedefs */
 
 type GQLContext = {
@@ -30,8 +33,14 @@ export async function approve(
   const { type, amount } = args
   switch (type) {
     case 'bond':
+      const gas =
+        (await ctx.livepeer.rpc.estimateGas(
+          'BondingManager',
+          'approveTokenBondAmount',
+          [amount],
+        )) * gasRate
       return await ctx.livepeer.rpc.approveTokenBondAmount(amount, {
-        gas: 60000,
+        gas: gas,
       })
       break
     default:
@@ -52,8 +61,14 @@ export async function bond(
   ctx: GQLContext,
 ): Promise<TxReceipt> {
   const { to, amount } = args
+  const gas =
+    (await ctx.livepeer.rpc.estimateGas(
+      'BondingManager',
+      'bondApprovedTokenAmount',
+      [amount, to],
+    )) * gasRate
   return await ctx.livepeer.rpc.bondApprovedTokenAmount(to, amount, {
-    gas: 750000,
+    gas: gas,
   })
 }
 
--- a/packages/graphql-sdk/src/resolvers/Mutation.js
+++ b/packages/graphql-sdk/src/resolvers/Mutation.js
@@ -33,7 +33,7 @@ export async function approve(
       const gas = await ctx.livepeer.rpc.estimateGas(
         'LivepeerToken',
         'approve',
-        [ctx.livepeer.config.defaultTx.from, amount],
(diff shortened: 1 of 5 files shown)


packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -4,6 +4,9 @@ import {
   TransactionSubmitted,
 } from './Subscription'
 
+/** Constants **/
+const gasRate = 1.5
```
yondonfu: Since we're using this multiplier every time we estimate the gas usage of a transaction, perhaps we can move the gas multiplier calculation logic to the SDK method?

Also, I think the gas multiplier could be smaller - x1.2?

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -30,8 +33,14 @@ export async function approve(
   const { type, amount } = args
   switch (type) {
     case 'bond':
+      const gas =
+        (await ctx.livepeer.rpc.estimateGas(
+          'BondingManager',
+          'approveTokenBondAmount',
```
yondonfu: The corresponding LivepeerToken contract function name (as defined in the contract ABI) should be "approve". Also, the function signature is `approve(uint256 spender, uint256 amount)` so the arguments here will need to be updated as well

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -30,8 +33,14 @@ export async function approve(
   const { type, amount } = args
   switch (type) {
     case 'bond':
+      const gas =
+        (await ctx.livepeer.rpc.estimateGas(
+          'BondingManager',
```
yondonfu: The approve function is defined in the LivepeerToken contract

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -52,8 +61,14 @@ export async function bond(
   ctx: GQLContext,
 ): Promise<TxReceipt> {
   const { to, amount } = args
+  const gas =
+    (await ctx.livepeer.rpc.estimateGas(
+      'BondingManager',
+      'bondApprovedTokenAmount',
```
yondonfu: The corresponding BondingManager contract function (as defined in the contract ABI) should be "bond"

packages/sdk/README.md
```diff
@@ -71,7 +71,17 @@ The following section details the rpc API's function signatures and typedefs.
 - [module~exports](#moduleexports)
   - [default](#default)
 - [livepeer~rpc](#livepeerrpc)
+  - [getENSName](#getensname)
+  - [getENSAddress](#getensaddress)
+  - [getBlock](#getblock)
   - [getEthBalance](#getethbalance)
+  - [getUnbondingPeriod](#getunbondingperiod)
+  - [getUnbondingPeriod](#getunbondingperiod-1)
```
yondonfu: Looks like an extra "getUnbondingPeriod" here?

packages/sdk/README.md
```diff
@@ -71,7 +71,17 @@ The following section details the rpc API's function signatures and typedefs.
...
+  - [getNumActiveTranscoders](#getnumactivetranscoders)
+  - [getNumActiveTranscoders](#getnumactivetranscoders-1)
```
yondonfu: Extra "getNumActiveTranscoders"?

packages/sdk/README.md
```diff
@@ -71,7 +71,17 @@ The following section details the rpc API's function signatures and typedefs.
...
+  - [getMaxEarningsClaimsRounds](#getmaxearningsclaimsrounds)
+  - [getMaxEarningsClaimsRounds](#getmaxearningsclaimsrounds-1)
```
yondonfu: Extra "getMaxEarningsClaimsRounds"?

packages/graphql-sdk/src/resolvers/Mutation.js
Randy1Burrell: I had it in the SDK originally but I then I had trail of thought that it might be best to take it out and allow anyone using rpc.estimateGas to determine their own rate of increase. I will update accordingly.

Randy1Burrell:
@yondonfu I have updated the code!

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -33,12 +30,11 @@ export async function approve(
...
-      const gas =
-        (await ctx.livepeer.rpc.estimateGas(
-          'BondingManager',
-          'approveTokenBondAmount',
-          [amount],
-        )) * gasRate
+      const gas = await ctx.livepeer.rpc.estimateGas(
+        'LivepeerToken',
+        'approve',
+        [ctx.livepeer.config.defaultTx.from, amount],
```
yondonfu: The first parameter here should actually be the address of the BondingManager contract (can probably fetch it from the SDK?) and not the address submitting the transaction - the transaction should approve the BondingManager contract to transfer `amount` tokens on behalf of the account submitting the transaction.

yondonfu (commented):
Could we also add gas estimation to the rest of the mutations in `graphql-sdk/src/resolvers/Mutation.js` i.e. `claimEarnings`, `unbond`, etc. ?

packages/graphql-sdk/src/resolvers/Mutation.js
Randy1Burrell: @yondonfu Updated. I didn't add any gas calculation to the `sendTransaction` function in `graphql-sdk/src/resolvers/Mutation.js` because it's not a part of Livepeer core and `estimateGas` does not calculate gas estimation for any function that it cannot find an ABI for. Later I will expand the estimateGas function within the SDK to include these different methods and update accordingly, but for now they or excluded.

packages/sdk/src/index.js
```diff
@@ -1843,6 +1843,54 @@ export async function createLivepeerSDK(
...
+    ): Promise<number> {
+      tx.value = tx.value ? tx.value : '0'
+      const gasRate = 1.2
+      const contractABI = config.abis[contractName]
+      const methodABI = utils.findAbiByName(contractABI, methodName)
+      const encodedData = utils.encodeMethodParams(methodABI, methodArgs)
+      return (
+        toNumber(
+          await config.eth.estimateGas({
+            to: tx.from,
```
yondonfu: Just realized there needs to be a few changes here:
- `to` should be the address of the contract that the function is being called on
- `from` should be `tx.from` (specifying the address of the account submitting the transaction)
- Since `gasRate` is a decimal, we should make sure to floor the result of multiplying by `gasRate` to make sure we end up with an integer

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -71,7 +80,14 @@ export async function claimEarnings(
   const { utils, config } = ctx.livepeer
   const { eth } = config
   const { endRound } = args
-  const txHash = await ctx.livepeer.rpc.claimEarnings(endRound)
+  const gas = await ctx.livepeer.rpc.estimategas(
```
yondonfu: "estimategas" -> "estimateGas"

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -143,5 +159,11 @@ export async function unbond(
   args,
   ctx: GQLContext,
 ): Promise<TxReceipt> {
-  return await ctx.livepeer.rpc.unbond()
+  const gas = await ctx.livepeer.rpc.estimategas('BondingManager', 'unbond', [
```
yondonfu: "estimategas" -> "estimateGas"

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -143,5 +159,11 @@ export async function unbond(
...
+    10,
```
yondonfu: I think this should be the actual amount that will be unbonded.

Since the explorer only allows full unbonding right now, the value passed in should be something like this: https://github.com/livepeer/livepeerjs/blob/1bbb10313a42882db84f3da6250087089aa22ff8/packages/sdk/src/index.js#L1881

Once the explorer supports partial unbonding it would be an amount specified by the user.

packages/sdk/src/index.js
Randy1Burrell: @yondonfu Making the changes required here throws this error: 
```
Uncaught (in promise) Error: [ethjs-rpc] rpc error with payload {"id":7207023777173,"jsonrpc":"2.0","params":[{"to":"0x511bc4556d823ae99630ae8de28b9b80df90ea2e","value":"0x0","data":"0x27de9e32000000000000000000000000000000000000000000000000000000000000000a"}],"method":"eth_estimateGas"} {"code":-32000,"message":"gas required exceeds allowance or always failing transaction"}
```

This happens even if I set a high gas value. Any thoughts about this error?

Randy1Burrell:
@yondonfu fixed!

packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -71,7 +81,14 @@ export async function claimEarnings(
   const { utils, config } = ctx.livepeer
   const { eth } = config
   const { endRound } = args
-  const txHash = await ctx.livepeer.rpc.claimEarnings(endRound)
+  const gas = await ctx.livepeer.rpc.estimateGas(
+    'BondingManager',
+    'claimEarnings',
+    [endround],
```
yondonfu: "endround" -> "endRound"

Randy1Burrell: Updated

yondonfu (approved):
LGTM 🚢

PR #240, opened 2018-10-11
fix(packages/explorer/src/utils.js): Fixed precision in formatting

### LINKED ISSUE #211: Lack of precision in formatted LPT/ETH amounts leads to user confusion

The [formatBalance](https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/utils.js#L80) helper is used in various places in the explorer to create a formatted string representing an amount of LPT/ETH. The function accepts a `decimals` parameter to specify the level of precision to use when representing the input amount which defaults to 8. At the moment, the default level of precision is used which leads to cases where the value displayed to the user is 0 even when the actual input amount > 0.

See:

- https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/views/AccountDelegating/index.js#L137
- https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/views/AccountDelegating/index.js#L129

If the user has an amount of LPT/ETH > 0, she does not see 0 displayed in the UI.

**To Reproduce (required)**
1. Navigate to the delegating tab in the explorer for an account that has bonded stake that cannot be represented with the default level of precision used by `formatBalance`
2. Observe that the displayed stake in the UI is 0 even though the actual bonded stake > 0

**Device (required):**
 - OS: OSX
 - Browser: Brave

Randy1Burrell:

Fixed precision in formatting to make it default to the length of number passed to the function.

Fixes #211 

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/AccountOverview/index.js
+++ b/packages/explorer/src/views/AccountOverview/index.js
@@ -90,7 +90,7 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
           title="Livepeer Token Balance"
           suffix="LPT"
           value={formatBalance(tokenBalance)}
-          subvalue={formatBalance(tokenBalance, 18)}
+          subvalue={formatBalance(tokenBalance)}
         >
           {isMe && (
             <React.Fragment>
--- a/packages/explorer/src/views/AccountOverview/index.js
+++ b/packages/explorer/src/views/AccountOverview/index.js
@@ -90,7 +90,7 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
           title="Livepeer Token Balance"
           suffix="LPT"
           value={formatBalance(tokenBalance)}
-          subvalue={formatBalance(tokenBalance)}
+          subvalue={formatBalance(tokenBalance, 18)}
         >
           {isMe && (
             <React.Fragment>
--- a/packages/explorer/src/utils.js
+++ b/packages/explorer/src/utils.js
@@ -79,9 +79,10 @@ export function formatPercentage(x: string, decimals = 2): string {
 
 export function formatBalance(
   x: string,
-  decimals: number = 6,
+  decimals: number = 0,
   unit: string = 'ether',
 ): string {
+  decimals = decimals ? decimals : x.length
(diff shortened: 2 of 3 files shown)


yondonfu (commented):
Perhaps we should consider applying this level of precision for all instances of amount formatting in the explorer?

Some examples:
- https://github.com/livepeer/livepeerjs/blob/7459bd4e841a69ea575eba4a13b2f9c7edb01a55/packages/explorer/src/components/ApproveForm/index.js#L69
- https://github.com/livepeer/livepeerjs/blob/7459bd4e841a69ea575eba4a13b2f9c7edb01a55/packages/explorer/src/components/BondForm/index.js#L142
- https://github.com/livepeer/livepeerjs/blob/7459bd4e841a69ea575eba4a13b2f9c7edb01a55/packages/explorer/src/components/BondForm/index.js#L142

Randy1Burrell:
It would probably be easier to apply this by changing the default value. @yondonfu what do you think?

yondonfu:
Seems reasonable (unless there is some UI reason for limiting the # of decimals)

Randy1Burrell:
There are some UI effects I saw when I changed the default value for decimals, but I have a fix for that.

Randy1Burrell:
format_precision

Hey @yondonfu, it works!

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -90,7 +90,7 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
           title="Livepeer Token Balance"
           suffix="LPT"
           value={formatBalance(tokenBalance)}
-          subvalue={formatBalance(tokenBalance, 18)}
+          subvalue={formatBalance(tokenBalance)}
```
yondonfu: Why are all the decimal parameters in the rest of this PR specified except here?

Randy1Burrell: Because it's no longer necessary with the default being the max length of unit to put the decimals in, but for good measure if you want me to put it in then I can.

yondonfu (approved):
LGTM

PR #243, opened 2018-10-12
[Ready] player /channel/local mode

ya7ya:
this checks for special `/local` instead of a broadcaster address to grab the local broadcasting. this works along with `current.m3u8` and allows devs to test their broadcast without needing a blockchain lookup

- checks for route `/channel/local`
- override blockchain lookup if it's running in local mode.

It works just like the production player, but will check the local broadcast if instructed. so this shouldn't break anything.

No, but it was a feature request.

**Checklist:**

### DIFF
--- a/packages/player/src/views/Channel/index.js
+++ b/packages/player/src/views/Channel/index.js
@@ -154,63 +154,109 @@ class Channel extends Component {
     this.setState({ tipAmount: e.target.value })
   }
 
+  getRootUrl(location, address) {
+    if (location.search && location.search.length > 0) {
+      let queryObject = parseQs(location.search)
+      if (queryObject && queryObject.source) {
+        console.log('using source qs', location)
+        return queryObject.source
+      }
+    } else {
+      let rootUrl = null
+      if (address === process.env.REACT_APP_LIVEPEER_TV_ADDRESS.toLowerCase()) {
+        rootUrl = process.env.REACT_APP_LIVEPEER_TV_STREAM_ROOT_URL
+      } else if (
+        address ===
+        process.env.REACT_APP_CRYPTO_LIVEPEER_TV_ADDRESS.toLowerCase()
+      ) {
+        rootUrl = process.env.REACT_APP_CRYPTO_LIVEPEER_TV_STREAM_ROOT_URL
+      } else if (
+        address === process.env.REACT_APP_INGEST2_ADDRESS.toLowerCase()
+      ) {
+        rootUrl = process.env.REACT_APP_INGEST2_STREAM_ROOT_URL
+      } else if (address === 'local') {
+        console.info('LOCAL ', address)
+        rootUrl = 'http://localhost:8935/stream'
+      } else {
+        rootUrl = process.env.REACT_APP_STREAM_ROOT_URL
+      }
+
+      return rootUrl
+    }
+  }
+
   async componentWillReceiveProps(nextProps) {
     const { data } = nextProps.account
     const location = nextProps.location
     const address = data.id.toLowerCase()
-    const [latestJob] = data.broadcaster.jobs
-    const manifestId = latestJob.streamId.substr(0, 68 + 64)
+    const match = nextProps.match
+    let rootUrl = this.getRootUrl(location, address)
 
-    if (!latestJob) {
-      if (nextProps.loading === false) this.setState({ live: false })
-      return
-    }
-
-    if (location.search) {
-      let queryObject = parseQs(location.search)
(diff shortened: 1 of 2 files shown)


packages/player/src/views/Channel/index.js
```diff
@@ -154,63 +154,63 @@ class Channel extends Component {
...
+          rootUrl = process.env.REACT_APP_LIVEPEER_TV_STREAM_ROOT_URL
+          break
+        case process.env.REACT_APP_CRYPTO_LIVEPEER_TV_ADDRESS.toLowerCase():
+          rootUrl = process.env.REACT_APP_CRYPTO_LIVEPEER_TV_STREAM_ROOT_URL
+          break
+        case process.env.REACT_APP_INGEST2_ADDRESS.toLowerCase():
+          rootUrl = process.env.REACT_APP_INGEST2_STREAM_ROOT_URL
+          break
+        case 'local':
+          rootUrl = 'http://localhost:8935/stream'
```
Randy1Burrell, line 175: No break statement here? Won't `case 'local'` fall through to `default`?

packages/player/src/views/Channel/index.js
```diff
@@ -154,63 +154,63 @@ class Channel extends Component {
...
+        default:
+          rootUrl = process.env.REACT_APP_STREAM_ROOT_URL
+          return rootUrl
```
Randy1Burrell: Why not put the `return` statement outside of the switch statement? This seems like it will only run on two conditions `local` and `default`.

packages/player/src/views/Channel/index.js
```diff
@@ -154,63 +154,63 @@ class Channel extends Component {
...
-      return
-    }
+    if (match && match.params && match.params.channel === 'local') {
+      return this.setState({
+        live: true,
+        url: `${rootUrl}/current.m3u8`,
+      })
+    } else {
+      const [latestJob] = data.broadcaster.jobs
+      const manifestId = latestJob.streamId.substr(0, 68 + 64)
```
Randy1Burrell: Can we try/catch this? I think a try catch might fix #248 or maybe another solution for checking if stream exists. Please let me know what you think.

Randy1Burrell (commented):
Hey @ya7ya, great work. I added some comment about possible changes. Let me know what you think.

packages/player/src/views/Channel/index.js
ya7ya: I have changed this part to run the check for `latestJob` first.

```suggestion
      const [latestJob] = data.broadcaster.jobs

      if (!latestJob) {
        if (nextProps.loading === false) this.setState({ live: false })
        return
      }

      const manifestId = latestJob.streamId.substr(0, 68 + 64)
```

ya7ya:
hey @Randy1Burrell , Thanks for the review, I've made the changes, take a look and let me know if this works for ya :)

packages/player/src/views/Channel/index.js
```diff
@@ -154,63 +154,66 @@ class Channel extends Component {
     this.setState({ tipAmount: e.target.value })
   }
 
+  getRootUrl(location, address) {
```
Randy1Burrell: @ya7ya Great job and thanks for fixing so quickly. Could we define this as a function just for good measure. Maybe `getRootUrl = (location, address) => {` or anything else you feel is necessary. Sorry for not mentioning this earlier.

Randy1Burrell (changes requested):
@ya7ya I found another issue I'd like to change before merging.

packages/player/src/views/Channel/index.js
ya7ya: @Randy1Burrell I'm not sure I follow, it is a function within `Channel` class, or do you mean `getRootUrl` should be defined outside the scope of `Channel` ??

Randy1Burrell: @ya7ya no I'm not saying you should define the method out of scope of `channel`. What I'm trying to say is that when I use a class method that begins with the word "get" then I'm usually returning a class variable. Otherwise, I think it's safer to use a function rather than an instance method.
Ex. 
```
getRootUrl() {
   return this.rootUrl
}
```
or

```
getRootUrl = (location, address) => {
    let rootUrl = location + address
    return rootUrl
}
```
Let me know what you think.

ya7ya: cool. I made the required change to the second option. 👍

Randy1Burrell (approved):
@ya7ya thanks man. Changes looks great!!

PR #245, opened 2018-10-19
Fetch maxEarningsClaimsRounds from contract

### LINKED ISSUE #233: Fetch maxEarningsClaimsRounds value from BondingManager

The explorer hardcodes 20 as the max number of rounds that a delegator can claim earnings for in a single transaction. Instead, the explorer should fetch the `maxEarningsClaimsRounds` parameter from the BondingManager contract so that even if the parameter value changes, the explorer will still be able to craft transactions that use the updated parameter value as the max number of rounds that a delegator can claim earnings for in a single transaction.

The explorer hardcodes 20 as the max number of rounds that a delegator can claim earnings for in a single transaction - see: https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/views/ClaimEarnings/index.js#L30

The SDK already contains a getter function [getMaxEarningsClaimsRounds()](https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L746).

The [Protocol](https://github.com/livepeer/livepeerjs/blob/master/packages/graphql-sdk/src/fragments/Protocol.js) GraphQL query could be updated to contain the output of this getter function in the result. The query would then be used in the explorer to retrieve `maxEarningsClaimsRounds`.

yondonfu:

Fetches the contract parameter `maxEarningsClaimsRounds` and uses it to determine what the maximum # of rounds to allow a user to claims earnings for in the explorer. Currently, the explorer hardcodes 20 as the maximum # of rounds to allow a user to claims for.

- *sdk*: Add `maxEarningsClaimsRounds` to `getProtocol()`
- *graphql-sdk*: Add `maxEarningsClaimsRounds` to the protocol query
- *explorer*: Connect ClaimsEarnings view with the protocol query to access `maxEarningsClaimsRounds`. Also updated the ClaimsEarningsForm to display the correct new last claim round and the remaining unclaimed rounds.

I tested claiming earnings on Rinkeby.

Fixes #233 

**Checklist:**

- [X] I have added tests to cover my changes.

Note: This PR does not update any of the hardcoded default gas limits attached to claimEarnings transactions. As a result, when a user claims earnings for the new maximum which is 100 on both mainnet and Rinkeby, the default gas limit will be insufficient. I did not updated any of these hardcoded values, because there is another open PR that implements gas estimation which should eliminate the requirement of updating hardcoded default gas limits. Thus, this PR should not be merged until after #239 is merged.


Randy1Burrell (approved):
I see no problem with this change.

PR #246, opened 2018-10-19
Use bondedAmount if already claimed earnings for past rounds when fully unbonding

yondonfu:

If a delegator has claimed earnings for all past rounds, the pendingStake returned from the BondingManager contract will be 0. In this case, bondedAmount should be up to date so it can be used as the parameter for fully unbonding (as opposed to using pendingStake which would be 0).

- *sdk*: Update `unbond()` to use bondedAmount if it is up to date and otherwise use pendingStake

I tested on Rinkeby.

**Checklist:**


Randy1Burrell (approved):
Change seems good to me.

Issue #178, opened 2018-08-08
When player is embedded using iframe, it doesn't go fullscreen


chrishobcroft:
**What's the problem? (required)**

When I load a webpage which has livepeer's player embedded in an iframe, and then click the "full screen" button (![image](https://user-images.githubusercontent.com/2212651/43825191-65fc28ec-9af4-11e8-9cf9-05d8572f4469.png) bottom right), the player does not fill the screen, instead it fills the iframe (if indeed it isn't already filling the iframe)

**What's the current behavior? (required)**

When I **have not** clicked on the "full screen" button ![image](https://user-images.githubusercontent.com/2212651/43825197-6aa23094-9af4-11e8-8550-d59d77f772af.png), it looks like this:

![image](https://user-images.githubusercontent.com/2212651/43825010-db9ea968-9af3-11e8-80d9-756acee32174.png)

When I **have** clicked on the "full screen" button ![image](https://user-images.githubusercontent.com/2212651/43825197-6aa23094-9af4-11e8-8550-d59d77f772af.png), it looks like this:

![image](https://user-images.githubusercontent.com/2212651/43825044-ef4f4508-9af3-11e8-9147-d33291141a53.png)

**Do you have any personal experiences dealing with this problem? (optional)**

Yes. I have embedded livepeer.tv into the website for [ETHIndia hackathon](http://ethindia.co), and when I click "full scree" ![image](https://user-images.githubusercontent.com/2212651/43825197-6aa23094-9af4-11e8-8550-d59d77f772af.png), it just fills up the frame reserved for the iframe :{

**Describe a possible solution you've considered (optional)**

When a user clicks the "full screen" button ![image](https://user-images.githubusercontent.com/2212651/43825197-6aa23094-9af4-11e8-8550-d59d77f772af.png), it should use the whole screen to show the content.

**Additional context (optional)**

The ability to embed livestreams is one of Livepeer's primary differentiators compared with YouTube Live / Facebook Live. Having it work very well is important as we seek to lure content publishers away from the Google mothership.

saskakol:
Hi @chrishobcroft! Just add the `allowfullscreen` attribute to the iframe element 😄 
So your iframe could look like this, for example: `<iframe width="240" height="160" src="https://media.livepeer.org/embed/0x0ddb225031ccb58ff42866f82d907f7766899014/?maxWidth=100%&aspectRatio=16:9" allowfullscreen></iframe>
`

[...]

PR #250, opened 2018-10-24
feat: Toggle fullscreen when embedding video

saskakol:
This pull request adds a switch in the player's embed dialog to add or remove the "allowfullscreen" flag from the iframe code.

- Imported switch from rmwc and added it to the layout
- Added logic to handle the flipping of the switch

No automated testing.

This does not fix any open issue, but addresses a closed one: #178 

**Checklist:**

### DIFF
--- a/packages/player/src/views/Channel/index.js
+++ b/packages/player/src/views/Channel/index.js
@@ -22,6 +22,7 @@ import Footer from '../../components/Footer'
 import { actions as routingActions } from '../../services/routing'
 import Modal from 'react-responsive-modal'
 import * as qs from 'query-string'
+import { Switch } from 'rmwc/Switch'
 
 const parseQs = str => {
   return qs.parse(str)
@@ -117,6 +118,7 @@ class Channel extends Component {
     tipAmount: 0,
     didCopy: false,
     bannerOpen: true,
+    allowFullscreen: true,
   }
 
   closeBanner = () => {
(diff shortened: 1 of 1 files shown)


packages/player/src/views/Channel/index.js
```diff
@@ -220,11 +222,21 @@ class Channel extends Component {
...
+      didCopy,
+      tipAmount,
+      bannerOpen,
+      allowFullscreen,
+    } = this.state
     const [latestJob] = jobs
     const { streamId, broadcaster = id } = latestJob || {}
     const web3IsEnabled = window.web3 && window.web3.eth.coinbase
-    const embedLink = `<iframe width="640" height="360" src="${
+    const embedLink = `<iframe width="640" height="360"${
```
Randy1Burrell: Let's add some space between the condition for readability purpose.

Maybe 
```JavaScript
const embedLink = `<iframe width="640" height="360" ${
￼            allowFullscreen ? 'allowfullscreen':''
￼     } src="${
```

Let me know what you think.

Randy1Burrell (changes requested):
@iSasFTW Awesome work!! Thanks for helping out with these issues. Could you please make this small change?

packages/player/src/views/Channel/index.js
saskakol: I did it in that way because I wanted there to be no extra space when the switch is turned off. So if the switch is off, there is one space to separate the other attributes, but when the switch is on there are two spaces around the `allowfullscreen` flag. If I did it the way you are proposing, wouldn't there be two spaces between `height="360"` and `src="...`?

Randy1Burrell: @iSasFTW how about we put the `${allowFullscreen ? 'allowfullscreen':''}` at the end of the opening iframe? That way the extra space will be at the end and not between height and src. 

I think a space at the end should be acceptable. Looking forward to hearing what you think.

saskakol: You're right, that would be better. I'll make the change

saskakol:
@Randy1Burrell Fixed the readability :) Still has the trailing space when switch is disabled, but it's not too bad.

Randy1Burrell (approved):
@iSasFTW Great job. Thanks for being so awesome!!

Randy1Burrell:
Thanks @iSasFTW. Great job!!

Issue #192, opened 2018-08-22
Add Reward Call Via Supermax Endpoint


ericxtang:
**What's the problem? (required)**
When a token holder is making the delegation decision, # of missed reward calls is not exposed.  As a result, people just look at "Total Stake".  We don't want this to happen.

**What's the current behavior? (required)**
No information about reward calls is in the "transcoders" tab of the explorer.

**Do you have any personal experiences dealing with this problem? (optional)**
The best way currently is to use Supermax, which is a very high-barrier action.

**Describe a possible solution you've considered (optional)**
We can use the Supermax API endpoint to load the reward information.  Try it with:
```
curl -X POST \
  https://supermax.cool/supermax/api/v1/activity/livepeer/aggregateCallCountByAccountOverRounds \
  -H 'Authorization: hello@world.com:eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImhlbGxvQHdvcmxkLmNvbSIsIl9pZCI6IjVhMTBjMzllNTQ2ZWZiMDc1MWEwYWUzYyIsImlhdCI6MTUxMTA1MDE4NH0.oIe738oMRmez_NI5U4-w9L0LiCjkssJ6f9KtbX9_N_k' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -d '{
  "network":"mainnet",
  "startRound":1000,
  "endRound":1101,
  "ethcallName":"reward",
  "accountAddresses":["0xbd91c9df3c30f0e43b19b1dd05888cf9b647b781", "0x4f4758F7167B18e1F5B3c1a7575E3eb584894dbc"]
}'
```

I propose we start with a simple metrics of "# of missed reward calls in last 30 rounds".  This can be queried with a single query to the Supermax API.

Randy1Burrell:
This has been worked on and pushed to production. Closing this, @ericxtang you can reopen if this has not been implemented.

PR #251, opened 2018-10-27
feat: 🎸Integrate Livepeer Subgraph

### LINKED ISSUE #199: Introduce an indexer for faster lookups and aggregate data
### What's the problem?
In addition to providing a friendly user interface for token holders to delegate, the explorer plays an important role in helping token holders make informed decisions on which transcoders to delegate towards by exposing relevant transcoder data. Currently, the explorer gets this data directly and exclusively off the blockchain resulting in a less than desirable user experience for a couple reasons:

1. Aggregate data that ought to play a crucial role in a delegator’s decision-making process such as “[total missed reward calls over n rounds](https://github.com/livepeer/livepeerjs/issues/192)” or “total segments claimed over n rounds” is precluded.
2. Querying data directly from the blockchain and formatting it client side to fit the requirements of the application is not performant. The transcoders view, for example, takes up to 20 seconds to render on initial page load.

### Describe a possible solution you've considered
An indexer would provide the explorer with both faster lookups and valuable aggregate data. I’ve been experimenting with [The Graph](https://thegraph.com/), a query protocol for getting data off blockchains and storage networks through a peer-to-peer network of indexing nodes using GraphQL. 

**How The Graph works:**

thegraph_dataflowdiagram

At a high level The Graph data flow is as follows:
1. A decentralized application creates/modifies data on Ethereum through a transaction to a smart contract.
2. The smart contract emits one or more events (logs) while processing the transaction.
3. Graph Node listens for specific events and fires handlers in a user-defined mapping.
4. The mapping is a WASM module which runs in a WASM runtime. It creates one or more store transactions in response to Ethereum events.
5. The store is updated along with indexes.
6. A decentralized application queries--via a GraphQL endpoint--the Graph Node, which in turn queries the store, for data which was ingested from the blockchain. This may include complex queries which take advantage of the store's indexes.
7. The decentralized application displays this data in a rich UI, which an end-user leverages in making new transactions against the Ethereum blockchain.
8. The cycle repeats.

**Creating a Livepeer Subgraph**

In The Graph, a project's GraphQL endpoint is referred as a subgraph, because once deployed to the decentralized network, it becomes just one subset of a global GraphQL endpoint. The subgraph is defined as a YAML file called a subgraph manifest. It comprises a schema, source data and mappings which are used to deploy an endpoint. 

I propose creating a Livepeer subgraph for the transcoder view to speed up the initial time to render and using it to add the following information to the transcoder card: 
- total missed reward calls (over last 30 rounds)
- total segments claimed (over last 30 rounds)

I started working on a Livepeer subgraph which you can check out [here](https://github.com/adamsoffer/livepeer-subgraph/). It's currently deployed to a Google Cloud instance if you’d like to play around with the [GraphiQL interface](http://35.237.114.197:8000/). Right now it simply indexes transcoder data (stored in a PostgreSQL database).

### Additional context
- The Graph Protocol is still heavily WIP. Its standalone indexer has been open sourced, but the decentralized network is yet to be launched, so the subgraph would be self-hosted to start.
- I’ve been in touch with The Graph team and they’re keen on making the software work well for Livepeer.
- GraphQL [schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html) can be used to combine the existing Livepeer GraphQL schema with the one generated by The Graph so everything can still be exposed over a single API endpoint.
- The transcoder card design would have to be updated to make room for the new information. Here’s a mockup of what that could look like.

Transcoder Card

adamsoffer:
This pull request reduces the Explorer's load time and incorporates aggregate data (missed reward calls over last 30 rounds to start) using [The Graph](https://thegraph.com/)'s self hosted data indexer ([Graph Node](https://github.com/graphprotocol/graph-node/)).

Note: The Explorer currently points to a Graph Node running on a staging server set up on a Google Cloud instance. I'm working (with the help of The Graph team) on setting up a production grade hosting environment and will update this pull request to point to the new server when ready.

- Adds a brand new package to the mono repo called `subgraph`. This package contains the source code for the Livepeer Subgraph, a project that defines what Livepeer events we want our Graph Node to source and what transformations we'd like to make on the data in those events before storing it in a PostgreSQL database and serving it over a GraphQL endpoint. Instructions for running a Graph Node locally and deploying our subgraph are included in the `README`.
- Combines multiple GraphQL APIs (the remote subgraph api and the local graph-sdk api) into a single GraphQL API endpoint using [schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html) inside the `apollo` package.
  - Before stitching the two schemas together, we first check the network conditions of our remote indexer. If it's down, we don't stitch, otherwise we do. This check ensures a resilient UI in the event that the indexer goes down.
  - The `transcoders` query in the `subgraph` api purposefully overrides the `transcoders` query in the `graphql-sdk` when stitched. This allows us to keep changes to the Explorer codebase to a minimum. For example, we don't have to conditionally check which GraphQL query to use based on network conditions.  
-  Adds a column to the Explorer transcoder card for missed reward calls over last 30 rounds. (We're able to display this data because the indexer creates a unique "reward" entry in our PostgreSQL database for each eligible/active transcoder on every round. If a transcoder calls `reward` for a given round, we store its reward tokens inside this entry in a field called `rewardTokens`. In our GraphQL query, if `rewardTokens` returns `null` for a given transcoder and round then we know the transcoder failed to call `reward`.)
- Makes a slight modification to the `graph-sdk` transcoders query. I had to rename one of the argument to match the query signature in our subgraph schema. This also required me to run the script that auto generates our graphql-sdk docs hence all the generated doc changes.
- Adds an environment variable, `REACT_APP_LIVEPEER_SUBGRAPH`, for setting the subgraph's graphql endpoint. This currently points to the staging server I set up.       

Ran all tests to see how updates affected other areas of code. All tests pass.

Fixes #199, #192 

**Checklist:**


packages/explorer/.env.development
```diff
@@ -1,6 +1,7 @@
 REACT_APP_HTTP_PROVIDER=
 REACT_APP_MAINNET_CONTROLLER_ADDRESS=0xf96d54e490317c557a967abfa5d6e33006be69b3
 REACT_APP_RINKEBY_CONTROLLER_ADDRESS=0x37dC71366Ec655093b9930bc816E16e6b587F968
+REACT_APP_LIVEPEER_SUBGRAPH=http://35.237.114.197:8000/by-name/livepeer/graphql
```
Randy1Burrell: @adamsoffer I think we should remove the dependency on http://35.237.114.197:8000/by-name/livepeer/graphql and just leave a comment in this file saying that anyone can use this link for development purposes if they don't have their own dev box.
That way dev won't depend directly on you keeping this server running.
Example 
```
## Use http://35.237.114.197:8000/by-name/livepeer/graphql or set up your own development environment with a subgraph
REACT_APP_LIVEPEER_SUBGRAPH=
```

packages/explorer/src/components/TranscoderCard/index.js
```diff
@@ -26,101 +26,120 @@ const TranscoderCard: React.ComponentType<TranscoderCardProps> = styled(
...
-            >
-              {ensName || `${id.substr(0, 10)}...`}
-            </Link>
+    rewards,
+  }) => {
+    let missedCalls: number = 0
+    if (rewards) {
+      missedCalls = rewards
+        .filter(reward => reward.rewardTokens === null)
+        .slice(-30).length
```
Randy1Burrell, line 35: Why do we limit missedCalls to the last 30?

packages/explorer/src/components/TranscoderCard/index.js
```diff
@@ -26,101 +26,120 @@ const TranscoderCard: React.ComponentType<TranscoderCardProps> = styled(
...
-                <span style={{ fontSize: 10 }}>{`(${MathBN.toBig(
-                  MathBN.div(bondedAmount + '00', totalStake),
-                ).toFixed(2)}%)`}</span>
-              }
+              label="Missed Calls"
+              type="number"
+              value={missedCalls}
               width="128px"
-              help="Amount of LPT you have staked towards this node."
+              help="Missed calls over last 30 eligible rounds."
```
Randy1Burrell, line 109: Why do we limit missed calls to the last 30 rounds? What if more rounds are missed?

packages/explorer/src/views/Transcoders/index.js
```diff
@@ -88,22 +85,27 @@ const TranscodersView: React.ComponentType<TranscodersViewProps> = ({
               towards a Transcoder and earn additional fees and LPT rewards.
               Read our Delegator Guide for a comprehensive overview, tutorials,
               and a FAQ. You can learn more about each transcoder on the forum
-              by checking out their social campaign on the Livepeer forum.<br />
+              by checking out their social campaign on the Livepeer forum.
```
Randy1Burrell, line 88: Can we put a link to the forum here?

packages/graphql-sdk/README.md
```diff
@@ -71,21 +72,21 @@ Contains all protocol data-fetching queries
...
 
-An Account by ETH address
+An Account by ETH address or ENS name
 
 </td>
 </tr>
 <tr>
 <td colspan="2" align="right" valign="top">id</td>
-<td valign="top"><a href="#string">String</a>!</td>
+<td valign="top"><a href="#string">String</a></td>
```
Randy1Burrell, line 84: This is not required anymore?

packages/explorer/.env.development
Randy1Burrell: @adamsoffer Let's remove this value and make a comment that they can use your subgraph for testing.

```
## COMMENTS ABOUT USING YOUR SUBGRAPH HERE
￼ REACT_APP_LIVEPEER_SUBGRAPH=
```

That way default dev doesn't depend on your subgraph

packages/explorer/src/components/TranscoderCard/index.js
Randy1Burrell, line 35: Why limit missedCalls to 30? What is missedCalls is more than 30?

packages/explorer/src/components/TranscoderCard/index.js
Randy1Burrell, line 109: Why limit missed calls to 30? What if missedCalls is more than 30? Can we make this more dynamic? 
Maybe `help="Missed calls over the last ${missedCalls} eligible rounds."`

packages/graphql-sdk/README.md
Randy1Burrell, line 84: Is this no longer required in graphql queries?

packages/subgraph/mappings/bondingManager.ts
```diff
@@ -0,0 +1,165 @@
+import "allocator/arena";
```
Randy1Burrell, line 2: @adamsoffer Can we include some more comments in the .ts files? Going forward I want to make it standard to have well documented/commented PRs.

Randy1Burrell (changes requested):
@adamsoffer great work. I have some minor requests. Let me know what you think.

packages/graphql-sdk/README.md
adamsoffer: @Randy1Burrell looks like this require was removed in a [commit](https://github.com/livepeer/livepeerjs/pull/113/files#diff-d945db54f3909cf2d4ff2b54cdc06e6b) back in June and the docs were just never regenerated.

Seems like it should be required though, so perhaps it wasn't intentional. If that's the case I can make that field required again and regenerate. cc: @yondonfu

packages/explorer/.env.development
adamsoffer: Good call 👍

packages/explorer/src/components/TranscoderCard/index.js
adamsoffer: Good question. I went along with the number suggested in #192.  It's like that saying "you're only as good as your last [fill in the blank]". If a transcoder missed a few calls over the course of one month (30 rounds) but then didn't miss a single round the following month, knowing the transcoder recently got its act together is valuable information that's otherwise lost if we display all. I'm open to suggestions though if we want to increase that time frame. cc: @ericxtang

packages/explorer/src/components/TranscoderCard/index.js
adamsoffer: Yeah ideally the user is able to select the number of past rounds. The transcoder detail view is probably a good candidate for this feature.

packages/subgraph/mappings/bondingManager.ts
adamsoffer: Absolutely 👍

packages/explorer/src/views/Transcoders/index.js
adamsoffer: Sure 👍

Randy1Burrell:
@adamsoffer any updates?

adamsoffer:
@Randy1Burrell Plan on pushing up update with your feedback EOD tomorrow. Been working on provisioning production server past week.

adamsoffer:
@Randy1Burrell I pushed up some updates based on your feedback. You'll notice lots more comments! I'll have this production server provisioned along with instructions very soon. Collaborating w/ Eric on that.

Randy1Burrell:
@adamsoffer thanks man. I will review this soon. I'm excited to deploy this!!

Randy1Burrell (approved):
@adamsoffer great work. Thanks!!

adamsoffer:
Woohoo! Thanks @Randy1Burrell

PR #252, opened 2018-10-30
[READY] implementing metmask privacy mode handler

### LINKED ISSUE #197: Update web3 instance access strategy

Metamask and other dapp browsers will stop injecting an Ethereum provider/web3 instance in a few months and instead users will have to actively approve the injection of a web3 instance.

For more details see the following links:
- Metamask announcement: https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8
- Relevant EIP: https://eips.ethereum.org/EIPS/eip-1102

We should adopt the new pattern for interacting with a web3 instance in the browser.

The explorer (and other livepeerjs based apps) rely on on an injected web3 instance from dapp browsers.

Since the change with dapp browsers will not be for another few months, this issue isn't immediately pressing, but I am creating it now to make sure we are aware of it such that we can incorporate the necessary changes such that the explorer is ready when the dapp browsers make the breaking change.

ya7ya:
this implements the required changes by [metamask new privacy mode](https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8)

- check for `window.ethereum` provider
- ask permission for accessing accounts.
- handle rejection gracefully.

I got the beta metamask and ran the new code against that to make sure it behaves like expected, and also going to test it with older versions to make sure it has backward compatibility 

Fixes #197 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/explorer/src/index.js
+++ b/packages/explorer/src/index.js
@@ -145,47 +145,51 @@ const trackingId = process.env.REACT_APP_GA_TRACKING_ID
         await window.ethereum.enable()
       } catch (e) {
         console.log('METAMASK | Access to accounts denied')
-        render(
-          <BasicModal title={'Metamask is Locked'}>
-            <p>
-              Not connected to web3. Please enable Metamask or another web3
-              enabled browser to be able to use full features in Livepeer
-              Explorer
-            </p>
-            <Button className={'primary'} onClick={enableAccounts}>
-              Enable
-            </Button>
-            <Button className={'primary'} onClick={limitedMode}>
-              Limited mode
-            </Button>
-          </BasicModal>,
-          document.getElementById('main-root'),
-        )
+        renderLockedWarning()
+        return
       }
     } else if (window.web3 && window.web3.version) {
       // this is the old way, accounts are always exposed.
       window.web3 = new window.Web3(window.web3.currentProvider)
     } else {
       // non dapp browser
-      render(
-        <BasicModal title={'Metamask is Locked'}>
-          <p>
-            Non-Ethereum browser detected. Livepeer requires Metamask or a dapp
-            browser like Mist
-          </p>
-        </BasicModal>,
-        document.getElementById('main-root'),
-      )
+      renderNonDappBrowserWarning()
     }
-
     return
   }
 
   const limitedMode = () => {
-    console.log('limited mode')
     window.web3 = new window.Web3(window.ethereum)
+  }
 
-    // window.web3 = new window.Web3(window.web3.currentProvider)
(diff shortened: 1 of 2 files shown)


packages/explorer/src/index.js
```diff
@@ -158,19 +188,19 @@ const trackingId = process.env.REACT_APP_GA_TRACKING_ID
...
-        4: process.env.REACT_APP_RINKEBY_CONTROLLER_ADDRESS,
-      }
-      while (version.network === null) {
-        console.log('waiting on web3 network version...')
-        await sleep(100)
-      }
-      opts.controllerAddress =
-        controllers[version.network] || process.env.REACT_APP_CONTROLLER_ADDRESS
-    }
+    await enableAccounts()
```
Randy1Burrell, line 219: @ya7ya I think we should allow users the ability to be able to use the explorer without having metamask unlocked because there are other pages such as the about page that they can view, which should be able to function properly without metamask. 

Let me know what you think.

packages/explorer/src/index.js
```diff
@@ -134,6 +136,34 @@ const trackingId = process.env.REACT_APP_GA_TRACKING_ID
...
+      // this is the new metamask way. details: https://bit.ly/2QQHXvF
+      window.web3 = new window.Web3(window.ethereum)
+      try {
+        await window.ethereum.enable()
+        return
+      } catch (e) {
+        console.log('METAMASK | Access to accounts denied')
+        render(
+          <BasicModal title={'Metamask is Locked'}>
+            <p>Please unlock metamask to be able to use Livepeer Explorer</p>
```
Randy1Burrell: Can we make this message more general? Maybe "Not connected to web3. Please unlock Metamask or another web3 enabled wallet to be able to use full features in Livepeer Explorer"

ya7ya: Yeah, I can make it a bit more general.

packages/explorer/src/index.js
ya7ya: @Randy1Burrell , The problem is the `web3` wouldn't be injected if privacy mode is enabled and the user doesn't unlock it. 
details here: https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

ya7ya: more details here : https://github.com/MetaMask/metamask-extension/pull/4703

Randy1Burrell (approved):
@ya7ya great work, although MetaMask just messed up every dapp's UX I think the difficult decisions they made is for the user's best interest at heart.

Issue #285, opened 2018-12-13
Rinkeby transcoder info not loaded based on graph node


ericxtang:
**Describe the bug (required)**
The explorer shows transcoder information from mainnet, even when the user switches Metamask to Rinkeby

**Expected behavior (required)**
Explorer should show Rinkeby transcoder information in this case.

**To Reproduce (required)**
1. Go to the explorer
2. Select "Rinkeby" in Metamask
3. Transcoder info is from mainnet

**Additional context (optional)**
This is caused by enabling of the graph node.  @adamsoffer - what do you think is the best strategy here?

adamsoffer:
Nice catch. I'd say we have two options:
1. We can spin up a separate graph node that indexes Rinkeby events. ([The network is configurable in a graph node](https://github.com/livepeer/livepeerjs/blob/master/packages/subgraph/docker-compose.yml#L19)).
2. We can fallback from the graph-node to infura when the network detected is not mainnet.

Thoughts?

adamsoffer:
For now, I'm leaning towards option 2 and falling back to infura whenever the network detected is not mainnet. We can always spin up another graph node for rinkeby down the line.

Randy1Burrell:
Sounds good to me.

Randy1Burrell:
PR #286 fixes this.

PR #286, opened 2018-12-14
Fall back from graph node to Infura when not on mainnet

adamsoffer:
This PR fixes an issue on staging where the explorer shows transcoder information from mainnet, even when the user switches Metamask to Rinkeby

- Falls back from the graph node to Infura whenever the network detected is _not_ mainnet.

#285 

**Screenshots (optional):**
screen shot 2018-12-13 at 9 04 52 pm

**Checklist:**

### DIFF
--- a/packages/apollo/src/index.js
+++ b/packages/apollo/src/index.js
@@ -133,7 +133,7 @@ export default async function createApolloClient(
 
     // If no subgraph endpoint is provided OR if web3 is available and not
     // on mainnnet, do not use subgraph data
-    if (!url || (IS_WEB3_AVAILABLE && !IS_MAINNET)) {
+    if ((IS_WEB3_AVAILABLE && !IS_MAINNET) || !url) {
       return false
     }
     try {
--- a/packages/apollo/src/index.js
+++ b/packages/apollo/src/index.js
@@ -127,7 +127,13 @@ export default async function createApolloClient(
    * @return {boolean}
    */
   async function isSubgraphAvailable(url: string): boolean {
-    if (!url) {
+    const IS_WEB3_AVAILABLE = window.web3
+    const IS_MAINNET =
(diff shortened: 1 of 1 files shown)


packages/apollo/src/index.js
```diff
@@ -127,7 +127,13 @@ export default async function createApolloClient(
...
    */
   async function isSubgraphAvailable(url: string): boolean {
-    if (!url) {
+    const IS_WEB3_AVAILABLE = window.web3
+    const IS_MAINNET =
+      IS_WEB3_AVAILABLE && `${window.web3.version.network}` === '1'
+
+    // If no subgraph endpoint is provided OR if web3 is available and not
+    // on mainnnet, do not use subgraph data
+    if (!url || (IS_WEB3_AVAILABLE && !IS_MAINNET)) {
```
Randy1Burrell: @adamsoffer can we change the order of the if statement to ensure that no check is done for url once we're not on mainnet? Maybe:

```JavaScript
if (!IS_MAINNET || !url) {
   return false
}
```
I think we can remove the `IS_WEB3_AVAILABLE` from the if statement, like the example above, being that you already checked for web3 when initializing `IS_MAINNET` . What do you think?

Randy1Burrell (commented):
@adamsoffer code is looking good but, just requesting a minor change.

packages/apollo/src/index.js
adamsoffer: Good call - I'll change the order to ensure that no check if done for url when not on mainnet. 

The reason why I check to see if web3 is available is because if it's not then the explorer won't use the graph node. We still want to use the graph node if a user is in incognito mode or doesn't have metamask installed.

Randy1Burrell (approved):
LGTM!! Thanks @adamsoffer.

PR #287, opened 2018-12-15
Fix bonding notification error

adamsoffer:
This PR fixes an issue on staging where a bonding notification error appears even in the case where a transaction is successfully confirmed. The issue was caused by [a breaking change in graphql-js v14.0.0](https://github.com/graphql/graphql-js/pull/1382) where servers that used to accept incorrect variable values now return errors to clients. 

- Updated the bond mutation so that it returns the correct custom scalar type (`JSON` as opposed to `Boolean`).
- Regenerated the graphql-sdk README and docs

- Received a success notification after successfully bonding in the explorer
- Received a success notification after successfully unbonding in the explorer
- Received a success notifcation after successfully claiming rewards in the explorer
- Reran `graph-sdk` tests

**Checklist:**


Randy1Burrell (approved):
This change looks good.

PR #293, opened 2018-12-18
Feature/dynamic max claim rounds

### LINKED ISSUE #233: Fetch maxEarningsClaimsRounds value from BondingManager

Randy1Burrell:

Removes the hard coded value 20 that isdicated maxEarningsClaimsRounds and replaced them with values attained from the protocol smart contract. 

- Replaces hard coded 20 with maxEarningsClaimsRounds 

I tested manually on Rinkeby by running the explorer and tried to bond from an address that has a high claim.

Fixes #233 

**Screenshots (optional):**

image

**Checklist:**


.vimsess
```diff
@@ -0,0 +1,1404 @@
+let SessionLoad = 1
```
yondonfu: Doesn't seem like this file should be committed?

Randy1Burrell: Whoops, good catch. Sorry about that man, I thought there was an ignore rule for this. Removing it now.

Randy1Burrell: @yondonfu I just removed it.

Randy1Burrell:
@yondonfu can you review this PR for me, please?

ya7ya (approved):

PR #310, opened 2019-01-11
fix(packages/explorer/graphql-sdk/src/resolvers/Mutations.js): Fix un…

### LINKED ISSUE #315: Can't view explorer after declining Metamask login
My Metamask is active but not logged in. When I try to visit the explorer, I cancel the login prompt. All I see on the explorer page is a pink screen.

The explorer should be shown.

**To Reproduce (required)**
1. Activate Metamask, but don't log in
2. Visit https://explorer.livepeer.org/
3. Dismiss the login prompt
4. See a pink screen

**Device (required):**
Chrome OSX

Randy1Burrell:
…defined defaultTX err

Fixes a a bug that causes undefined defaultTx err when unbonding.

FIxes #315 and #311 

**Screenshots (optional):**

### Before Giving Permission to MetaMask
image

### After refusing MetaMask Permission
image

**Checklist:**

### DIFF
--- a/packages/explorer/src/components/BondForm/index.js
+++ b/packages/explorer/src/components/BondForm/index.js
@@ -2,12 +2,13 @@ import * as React from 'react'
 import { Form } from 'react-final-form'
 import { Field } from 'react-final-form-html5-validation'
 import { Link } from 'react-router-dom'
-import Confetti from 'react-dom-confetti'
+import Confetti from '../Confetti'
 import { withProp } from '../../enhancers'
 import { formatBalance, toBaseUnit, MathBN } from '../../utils'
 import InlineAccount from '../InlineAccount'
 import InlineHint from '../InlineHint'
 import Button from '../Button'
+import { H1 } from '../HTags'
 import type { BondFormProps } from './props'
 
 /**
@@ -34,18 +35,6 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
   values,
   ...props
 }) => {
-  const confetti = (
-    <Confetti
-      active={submitSucceeded}
-      config={{
-        angle: 90,
-        spread: 197,
-        startVelocity: 45,
-        elementCount: 50,
-        decay: 0.9,
-      }}
-    />
-  )
   const noAllowance = allowance === '0'
   const noBondedAmount = bondedAmount === '0'
   const cannotBond =
@@ -54,7 +43,7 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
   if (submitFailed && submitError && !/User denied/.test(submitError)) {
     return (
       <React.Fragment>
-        {confetti}
+        <H1>Failed</H1>
         <p>
           There was an error submitting your transaction. See error message
           below for more details:
@@ -84,7 +73,8 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
     // console.log('rendering bond success')
     return (
       <React.Fragment>
-        {confetti}
+        <Confetti active={submitSucceeded} />
+        <H1>Success!</H1>
         <p>
           Congratulations! You successfully bonded your tokens to{' '}
           {delegateAddress}. You may{' '}
@@ -103,7 +93,7 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
   }
   return (
     <React.Fragment>
-      {confetti}
+      <H1>Bond Your Token</H1>
       <div>
         <p>
           <strong>Bonding Tips:</strong>
--- a/packages/explorer/src/components/BondForm/index.js
+++ b/packages/explorer/src/components/BondForm/index.js
@@ -20,9 +20,11 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
   delegateAddress,
   errors,
   handleSubmit,
+  handleChange,
   loading,
   max,
   maxEarningsClaimsRounds,
+  me,
   onCancel,
   onUpdateAllowance,
   reset,
@@ -35,11 +37,13 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
   values,
   ...props
 }) => {
+  const { delegator } = me.data
   const noAllowance = allowance === '0'
   const noBondedAmount = bondedAmount === '0'
   const cannotBond =
     (noAllowance && noBondedAmount) ||
     (values.amount && MathBN.gt(toBaseUnit(values.amount), allowance))
+  const noZero = delegator.delegateAddress === delegateAddress
   if (submitFailed && submitError && !/User denied/.test(submitError)) {
     return (
       <React.Fragment>
@@ -146,21 +150,29 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
           <p style={{ marginTop: 0 }}>Amount to Bond</p>
           {noBondedAmount ? null : (
             <p style={{ fontSize: 14, lineHeight: 1.5 }}>
-              <strong>Note:</strong> By entering 0 or leaving this field blank,
-              you will transfer your existing bonded amount of{' '}
-              {formatBalance(bondedAmount, 18)} LPT the selected delegate.
+              <strong>Note:</strong>
+              {noZero
+                ? ` You cannot bond 0 LPT to a delegate you are already bonded to.`
+                : ` By entering 0 or leaving this field blank,
+                  you will transfer your existing bonded amount of
+                  ${formatBalance(
+                    bondedAmount,
+                    18,
+                  )} LPT the selected delegate.`}
             </p>
           )}
           <div>
             <Field
-              name="amount"
               component="input"
-              type="number"
-              min="0"
-              max={max}
               disabled={loading}
+              max={max}
+              maxLength={20}
+              min={noZero ? Math.pow(10, -18).toString() : '0'}
+              minLength={1}
+              name="amount"
               rangeOverflow={`This amount is too large. Either your token balance is too low or you need to approve a higher transfer allowance.`}
               step="any"
+              type="number"
               style={{
                 width: '90%',
                 height: 48,
--- a/packages/explorer/src/components/ClaimEarningsForm/index.js
+++ b/packages/explorer/src/components/ClaimEarningsForm/index.js
@@ -2,7 +2,7 @@
 import * as React from 'react'
 import { Form, Field } from 'react-final-form'
 import { Link } from 'react-router-dom'
-import Confetti from 'react-dom-confetti'
+import Confetti from '../Confetti'
 import { withProp } from '../../enhancers'
 import { MathBN } from '../../utils'
 import InlineHint from '../InlineHint'
@@ -19,6 +19,7 @@ const ClaimEarningsForm: React.ComponentType<ClaimEarningsFormProps> = ({
   loading,
   min,
   max,
+  maxEarningsClaimsRounds,
   onCancel,
   pristine,
   reset,
@@ -35,24 +36,9 @@ const ClaimEarningsForm: React.ComponentType<ClaimEarningsFormProps> = ({
   const currentRound = MathBN.add(lastClaimRound, max)
   const endRound = MathBN.add(values.numRounds, lastClaimRound)
   const remaining = MathBN.sub(diff, values.numRounds)
-  const confetti = (
-    <Confetti
-      active={submitSucceeded}
-      config={{
-        angle: 90,
-        spread: 197,
-        startVelocity: 45,
-        elementCount: 50,
-        decay: 0.9,
-      }}
-    />
-  )
-  // console.log(props)
   if (submitFailed && submitError && !/User denied/.test(submitError)) {
-    // console.log('rendering claimEarnings fail')
     return (
       <React.Fragment>
-        {confetti}
         <p>
           There was an error submitting your transaction. See error message
           below for more details:
@@ -79,14 +65,13 @@ const ClaimEarningsForm: React.ComponentType<ClaimEarningsFormProps> = ({
     )
   }
   if (submitSucceeded) {
-    // console.log('rendering claimEarnings success')
     return (
       <React.Fragment>
-        {confetti}
+        <Confetti active={submitSucceeded} />
         <p>
           Congratulations! You successfully claimed your earnings through round
-          #{endRound}. You now have {remaining} rounds of earnings left to
-          claim.
+          #{endRound}. You now have {Math.abs(remaining)} rounds of earnings
+          left to claim.
         </p>
         <div style={{ textAlign: 'right', paddingTop: 24 }}>
           {onCancel && (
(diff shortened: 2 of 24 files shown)


packages/graphql-sdk/src/resolvers/Mutation.js
```diff
@@ -160,19 +160,18 @@ export async function unbond(
...
-      ? bondedAmount
-      : pendingStake
+  const { amount } = args
+  let gas = 0
+  try {
+    gas = await ctx.livepeer.rpc.estimateGas('BondingManager', 'unbond', [
+      amount,
+    ])
+  } catch (err) {
+    console.log(err)
```
yondonfu: I think that if `estimateGas()` throws an error we will continue and try submitting a transaction with 0 gas which would always fail - perhaps we should prevent transaction submission in this case.

packages/explorer/src/components/ClaimEarningsForm/index.js
```diff
@@ -79,14 +65,13 @@ const ClaimEarningsForm: React.ComponentType<ClaimEarningsFormProps> = ({
...
-    // console.log('rendering claimEarnings success')
     return (
       <React.Fragment>
-        {confetti}
+        <Confetti active={submitSucceeded} />
         <p>
           Congratulations! You successfully claimed your earnings through round
-          #{endRound}. You now have {remaining} rounds of earnings left to
-          claim.
+          #{endRound}. You now have {Math.abs(remaining)} rounds of earnings
```
yondonfu: Has it ever been the case where `remaining` is negative? If so, I think something somewhere else is going wrong i.e. allowing someone to try to claim earnings for more rounds than should be possible and that the problem should be resolved there. Otherwise, if `remaining` is indeed negative here, then the user is still shown inaccurate information i.e. taking the absolute value of a negative # of rounds claimed which wouldn't make sense.

packages/explorer/src/components/ClaimEarningsForm/index.js
```diff
@@ -100,11 +85,11 @@ const ClaimEarningsForm: React.ComponentType<ClaimEarningsFormProps> = ({
   }
   return (
     <React.Fragment>
-      {confetti}
       {max === '20' && (
         <InlineHint flag="claim-earnings-max" disableHide>
           <p>
-            <strong>Note:</strong> You may claim up to 20 rounds at a time.
+            <strong>Note:</strong> You may claim up to {maxEarningsClaimsRounds}{' '}
```
yondonfu, line 91: Out of curiosity what does `{' ' }` do? Is it not sufficient to just have `{maxEarningsClaimsRounds} rounds`?

packages/explorer/src/components/UnbondTx.js
```diff
@@ -32,55 +34,60 @@ const UnbondTx = ({
...
+            style={{ fontSize: 12, marginTop: canWithdraw ? '0px' : '-10px' }}
+          >
+            {canWithdraw
+              ? `The unbonding period has completed${
+                  isMe
+                    ? ' and you can withdraw your LPT'
+                    : ' for this amount of LPT'
+                }.`
+              : isMe
+              ? `You will be able to unbond approximately on
```
yondonfu: Should this be "You will be able to withdraw approximately on..."?

packages/explorer/src/components/UnbondTx.js
```diff
@@ -32,55 +34,60 @@ const UnbondTx = ({
...
+                 ${formatRoundsToDate(withdrawRound - currentRound)}.`
+              : `This amount of LPT can be unbonded on
```
yondonfu: Should this be "This amount of LPT can be withdraw on..."?

packages/explorer/src/components/UnbondingForm/index.js
```diff
@@ -76,14 +65,15 @@ const UnbondForm: React.StatelessFunctionalComponent<UnbondFormProps> = ({
   if (submitSucceeded) {
     return (
       <React.Fragment>
-        {confetti}
-        <h3>Success: The Unbonding Process Has Begun.</h3>
+        <Confetti active={submitSucceeded} />
+        <H1>Success!</H1>
+        <h3>The Unbonding Process Has Begun.</h3>
         <p>
           It will take 7 days (the unbonding period) in order to be able to
```
yondonfu, line 72: Just noting that ideally the unbonding period would be fetched from the contract instead of using a hardcoded number of days since technically the unbonding period could be changed in the future. This can probably be resolved outside of this PR though.

packages/explorer/src/enhancers/index.js
```diff
@@ -538,3 +538,63 @@ export const connectAccountDelegatorUnbondLockQuery = graphql(
...
+    status
+    delegateAddress
+    bondedAmount
+    fees
+    delegatedAmount
+    lastClaimRound
+    pendingFees
+    pendingStake
+    startRound
+    withdrawAmount
```
yondonfu: I don't think we need `withdrawAmount` and `withdrawRound` in this query anymore since we can also query for unbonding locks now? These fields were originally included to hide unbonding locks from the user (since a UI for unbonding locks was not available yet)

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -50,6 +58,15 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
   const transactionData = transactions.data
     // only livepeer transactions
     .filter(x => x.method)
+
+  if (unbondlocks) {
+    const reducer = (accumulator, itemNext) =>
+      (accumulator = MathBN.add(accumulator, itemNext.amount))
+    const filter = item => item['withdrawRound'] !== '0'
+    unbondlocks = unbondlocks.filter(filter)
+    unbondlocks = unbondlocks.reduce(reducer, 0)
```
yondonfu: This is a little confusing to me as `unbondlocks` initially is an array and then after this reduction it stores a single value - perhaps track the accumulated value in a separate variable for clarity

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -106,7 +128,32 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
...
-                <span>bond to a transcoder</span>
+                <span style={{ fontSize: '10px' }}>bond to a transcoder</span>
+                <span style={{ marginLeft: 8 }}>&rarr;</span>
+              </Button>
+            </React.Fragment>
+          )}
+        </MetricBox>
+        <MetricBox
+          className="eth-balance"
+          help={`The amount of Livepeer Token (LPT) owned  owned by this
```
yondonfu: "(LPT) owned  owned" -> "(LPT) owned by"?

packages/explorer/src/views/ClaimEarnings/index.js
```diff
@@ -37,7 +37,8 @@ const ClaimEarningsView: React.ComponentType<ClaimEarningsViewProps> = ({
         from={from}
         initialValues={{ numRounds: '1' }}
         loading={loading}
-        max={MathBN.min(diff, maxClaims)}
+        max={MathBN.min(diff, maxEarningsClaimsRounds)}
+        maxEarningsClaimsRounds={maxEarningsClaimsRounds}
```
yondonfu, line 41: Are the brackets around `maxEarningsClaimsRounds` necessary?

yondonfu (changes requested):

packages/explorer/src/views/AccountDelegating/index.js
```diff
@@ -134,10 +134,13 @@ const AccountDelegating: React.ComponentType<AccountDelegatingProps> = ({
...
                 withdrawAmount,
                 18,
-              )} LPT may be withdrawn at round #${withdrawRound}`
+              )} LPT may be withdraw approximately on ${formatRoundsToDate(
+                withdrawRound - lastInitializedRound,
+              )}`
             : `${formatBalance(withdrawAmount, 18)} LPT may be withdrawn`
         }
       >
+        {console.log(withdrawRound)}
```
iameli: This is a bit noisy

packages/explorer/src/components/ClaimEarningsForm/index.js
iameli: Usually these are introduced by Prettier being overzealous in maintaining the input formatting

packages/explorer/src/utils.js
```diff
@@ -182,14 +189,17 @@ export async function enableAccounts() {
...
   return
 }
 
 export async function limitedMode() {
-  window.limitedWeb3Conn = true
-  window.web3 = new window.Web3(window.ethereum)
+  // Enable limited mode if and only if window.Web3 exists
+  if (window.Web3) {
+    window.limitedWeb3Conn = true
+    window.web3 = new window.Web3(window.ethereum)
```
iameli, line 200: I'm not yet familiar with these, but might it ever be the case that `window.Web3` is defined but `window.ethereum` is not? `if (window.Web3 && window.ethereum)`, perhaps?

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -61,28 +78,33 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
           className="eth-address"
           help="The Ethereum address representing this account"
           title="ETH Address"
-          width="100%"
-          subvalue={<code style={{ fontSize: 16 }}>{id}</code>}
+          value={
```
iameli, line 82: I also fixed this in my branch; I'll take that commit out of my PR.

packages/graphql-sdk/src/resolvers/Mutation.js
iameli: Could just take out the try/catch entirely and bubble up the error.

packages/explorer/src/components/BondForm/index.js
```diff
@@ -156,21 +150,29 @@ const BondForm: React.StatelessFunctionalComponent<BondFormProps> = ({
...
             <Field
-              name="amount"
               component="input"
-              type="number"
-              min="0"
-              max={max}
               disabled={loading}
+              max={max}
+              maxLength={20}
+              min={noZero ? Math.pow(10, -18).toString() : '0'}
```
iameli, line 170: Does this field take a string? It looks like every other field takes a number.

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -9,38 +10,45 @@ import { Card } from 'rmwc/Card'
...
   coinbase: GraphQLProps<Coinbase>,
+  currentRound: GraphQLProps<Round>,
   history: History,
   match: Match,
   onDepositETH: (e: Event) => void,
   onRequestETH: (e: Event) => void,
   onRequestLPT: (e: Event) => void,
   onTransferLPT: (e: Event) => void,
   transactions: GraphQLProps<Transaction>,
+  unbondlocks: GraphQLProps<UnbondLock>,
```
iameli, line 36: Perhaps camelcase'd `unbondLocks` for consistency?

packages/explorer/src/views/AccountOverview/index.js
```diff
@@ -50,6 +58,15 @@ const AccountOverview: React.ComponentType<AccountOverviewProps> = ({
   const transactionData = transactions.data
     // only livepeer transactions
     .filter(x => x.method)
+
+  if (unbondlocks) {
+    const reducer = (accumulator, itemNext) =>
+      (accumulator = MathBN.add(accumulator, itemNext.amount))
```
iameli: I don't believe the `accumulator =` assignment is necessary here, `MathBN.add` is already returning the value you want

packages/explorer/src/components/MetricBox.js
```diff
@@ -58,6 +59,7 @@ const MetricBox = ({
               textOverflow: 'ellipsis',
               overflow: 'hidden',
               whiteSpace: 'nowrap',
+              fontSize: valueSize || '1.17em',
```
iameli: Another way to handle this would be to pass a `textStyle` prop to `MetricBox` and deconstruct it at the end of this `style`:

```
<strong
  style={{
    fontWeight: 400,
    textOverflow: 'ellipsis',
    overflow: 'hidden',
    whiteSpace: 'nowrap',
    fontSize: '1.17em',
    ...(textStyle || {})
  }}
>
```

And invoke with:
```
<MetricBox textStyle={{fontSize: "100px"}} />
```

That'll allow for more customization in the future without adding more props

packages/explorer/src/components/UnbondTx.js
```diff
@@ -32,55 +34,60 @@ const UnbondTx = ({
...
-          justifyContent: 'flex-start',
-        }}
-      >
-        {withdrawRound <= currentRound ? (
-          <Tooltip text="You can withdraw now">
+      {isMe && (
+        <div
+          style={{
+            minWidth: '40%',
+            justifyContent: 'flex-start',
```
iameli, line 59: I don't believe this will have an effect unless there's a `display: flex` on the box (I know this wasn't your change though)

packages/explorer/src/views/ClaimEarnings/index.js
iameli: Yeah, it'd put that name in there as a string otherwise

iameli (changes requested):

packages/explorer/src/components/ClaimEarningsForm/index.js
Randy1Burrell: Yes it has been a case where remaining is negative, and sure there is something wrong somewhere. I have found it to say I have -30 unclaimed rounds when I actually have 30 unclaimed rounds. This Math.abs I'll remove until someone is able to find the actual bug.

packages/explorer/src/components/UnbondingForm/index.js
Randy1Burrell: Sure let's resolve this in another PR.

packages/explorer/src/views/AccountDelegating/index.js
Randy1Burrell: My mistake. I should have removed this when I was cleaning up the code. I guess I missed it, removing it now.

packages/explorer/src/utils.js
Randy1Burrell: @iameli thanks. I've add this in for good measure in case someone screws up and forget to inject ethereum along with web3, but to answer your question MetaMask usually injects both Web3 and ethereum into the window object.

packages/explorer/src/views/AccountOverview/index.js
Randy1Burrell: @iameli UnbondLock is Pascal Cased for consistency with the rest of the src files in the explorer. Notice how line 35 `Transaction` is Pascal case.

Randy1Burrell: @iameli My bad, I thought you meant camel casing the `UnbondLock` prop for GraphQL. Changing it in this area means I will have to change other areas that have already been merged. It's probably best not to touch any other area yet, being that I'm on a tight deadline.

packages/explorer/src/components/MetricBox.js
Randy1Burrell: Great suggestion! I have implemented this.

packages/explorer/src/components/UnbondTx.js
Randy1Burrell: @iameli Thanks for the review. Implemented!

Randy1Burrell:
@yondonfu and @iameli changes have been implemented.

iameli:
LGTM! 🚢

yondonfu (approved):
LGTM

PR #325, opened 2019-02-07
Update subgraph

adamsoffer:
**What does this pull request do? Explain your changes.**
This PR updates the Livepeer Subgraph

**Specific updates**
- Introduces two new entities, `Delegator` and `Share`
- Introduces a new data source, `JobsManager` for indexing earned fee shares using the `DistributeFees` event
- Accounts for deprecated event signatures `Bond(address,address)`, `Unbond(event,event)` for properly indexing delegators
- Accounts for deprecated return types in `getDelegator()` for properly indexing delegator reward and fee shares
- Moves mappings and generated types inside `src` directory (convention)
- Updates `graph-cli` `graph-ts` dependencies
- Updates package.json scripts for creating and deploying the subgraph locally as well as to The Graph's hosted service.

Note: because delegator reward and fee shares are indexed at the time they are distributed and before any are claimed, there are some discrepancies in delegator rewards and fee shares earned _before_ [the protocol accounting bug was patched on 8.31.18](https://forum.livepeer.org/t/protocol-accounting-issue-identified-8-28-18-9-42pm-et-fixed-8-31-18-5-30pm-et/369/2). We don't emit an event inside the protocol's `claimEarnings` function so we can't account for this, however, watching smart contract function calls that don't emit events is on The Graph's near term roadmap so we will be able to in the near future. In the meantime, if we know the round number the patch was deployed on, we can choose to ignore delegator shares earned before that round.

I tested these changes by deploying to The Graph's hosted service [here](https://thegraph.com/explorer/subgraph/adamsoffer/livepeer-canary) and cross-referencing the indexed data with the underlying contract state to make sure it's correct. For example, I added up the shares (fees and rewards per round) for a delegator and compared it to its pending fees/rewards in the contract.

Doesn't close but opens the door for closing livepeer/explorer#74 since it's now possible to view a transcoder's bonded addresses

**Checklist:**


iameli:
This looks great! I'll be looking through this today but I'm still bootstrapping a lot of my Livepeer ecosystem knowledge so it will likely take me a big chunk of today to have feedback to you.

adamsoffer:
Cool sounds good! And no rush. Thanks @iameli.

Randy1Burrell:
@adamsoffer great work but seems we need to update node in the CI config for the tests to pass.

adamsoffer:
@Randy1Burrell Yeah I noticed that. Looks like `graph-cli` dropped support for Node v8 and v9. https://github.com/graphprotocol/graph-cli/commit/2a47ec39f657727352a6d6d5201c4522dffad297

iameli:
@adamsoffer FYI some minor changes to automate the codegen, fix capitalization, bump Node version in CI https://github.com/livepeer/livepeerjs/pull/325/commits/bb6f3b95d4649a261a07318e7162251264690ad7

iameli:
One thought I've had so far — I think we should perhaps be more specific than "deprecated" for the logic that handles the old Bond/Unbond events. We'll inevitably need to accommodate more changes in the future, so it seems like maybe we should name stuff (when pertinent) with the pertinent LIP, in this case [LIP-11](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-11.md). Not totally sure but maybe something like `bondingManger_LIP0.ts` and `bondingManger_LIP11.ts`?

(edit: whoops, accidentally hit "close" 😬)

adamsoffer:
> One thought I've had so far — I think we should perhaps be more specific than "deprecated" for the logic that handles the old Bond/Unbond events. We'll inevitably need to accommodate more changes in the future, so it seems like maybe we should name stuff (when pertinent) with the pertinent LIP, in this case [LIP-11](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-11.md). Not totally sure but maybe something like `bondingManger_LIP0.ts` and `bondingManger_LIP11.ts`?
> 
> (edit: whoops, accidentally hit "close" 😬)

Love that naming convention. Let's go with it!

Update:
I renamed `bondingManager.ts` to `bondingManager_LIP11.ts` and `bondingManger_deprecated.ts` to `bondingManager.ts`. I also renamed the `Reward` entity to `Pool` and added descriptions to the graphql schema.

iameli:
@adamsoffer Nice! I've been caught up in travel plans but am hoping to review this today or tomorrow

iameli:
Still looking at this/figuring out what tweaks need to be made to the explorer to merge it. Currently that's led me to this as I tried to figure out how I could track down the errors https://github.com/livepeer/livepeerjs/issues/330

adamsoffer:
Hey @iameli 👋 - I pushed up some new subgraph updates.

The Graph recently released support for call handers which allowed me to index the following missing delegator fields: `bondedAmount`, `fees`, `pendingStake`, and `pendingFees`. This data was previously inaccessible due to the fact that the `claimEarnings` function in the `bondingManager` contract doesn't emit an event. I also added `unbondingLocks`, `delegatedAmount`, and `allowance` fields to the delegator entity.

I know you've been working incrementally improving the explorer's underlying infrastructure so hopefully, this helps. I think we now have enough data to begin leveraging the graph for the remaining views :) 

A few things to note:
1. Call handler support requires a Parity archive node with tracing enabled — Infura sadly doesn't provide that but I believe the archive node you set up has this enabled. 
2. I refrained from indexing the delegator `status` field since that would require updating each delegator upon every `newRound` event. This would take an unreasonable amount of time (it already takes over two days to sync as is with the introduction of delegator data) so we'll have to make sure to resolve that specific field directly from the blockchain for now. Side note: I believe The Graph team is working on identifying the bottleneck causing these handlers to take forever to run. 
3. To test these updates on the Graph Node you have running, you'll have to update to version 0.12.0. I tested on the hosted service, but it has a timeout in place for handlers that run too long I so wasn't able to test it all the way through to the end.

cristianmalfesi:
Hi guys! These changes look very interesting. Do you have a plan to get progress with this PR? We would like to use the power of these changes from the Livepeer subgraph!

cristianmalfesi:
Hi @adamsoffer! I have talked with Jannis from The Graph. They don't think an exception is feasible. 
Jannis checked a little the code and he said: 
"It's this part that triggers the timeout: https://github.com/livepeer/livepeerjs/blob/85ebb36856bdb6dfa4afc35d557a2a61c9ec8756/packages/subgraph/mappings/roundsManager.ts#L32-L63
It only makes a couple of contract calls for each transcoder, but together with the looping it becomes too much.
One thing that could be avoided there is the store.get Instead, you could just do
`let transcoder = new Transcoder(currentTranscoder.toHex())
transacoder.active = active
transcoder.save()
`
At the beginning of each loop after the isActiveTranscoder call.
That will save some time, but the contract calls are still the problem."

Can you check this change? or can I try an improvement here?

adamsoffer:
@chm1975 Thanks for looking into this! 

I can try using `new Transcoder(currentTranscoder.toHex())` instead of `Transcoder.load(transcoderAddress.toHex());` as suggested, although as Jannis pointed out it's the contract calls that are the problem. (FYI it's actually the `reward` mapping that's [causing the timeout in the hosted solution](https://thegraph.com/explorer/subgraph/adamsoffer/livepeer-canary?selected=logs).)

I greatly appreciate the offer to look into this. Please feel free to make any improvements that you think could improve performance! Would be great if we can ultimately get this working on The Graph's hosted solution. BTW these updates are on my fork located [here](https://github.com/adamsoffer/livepeerjs/) if you want to pull down and run some test deployments.

cristianmalfesi:
@adamsoffer thanks for the quick feedback. I will take a look using your fork.

adamsoffer:
Closing in favor of livepeer/livepeerjs#506

PR #328, opened 2019-02-12
update development ports for explorer and player

iameli:
I wanna be able to run them both at once without them bouncing around
between different ports. I went with 3001 and 3002 'cause 3000 is also used for
the stats interface in the test harness. Any objections?

### DIFF
--- a/packages/explorer/package.json
+++ b/packages/explorer/package.json
@@ -55,7 +55,7 @@
     "eject": "npx react-scripts eject",
     "postbuild": "cp build/index.html build/404.html",
     "serve": "npx serve -s build",
-    "start": "npx react-scripts start",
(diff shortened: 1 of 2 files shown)


Randy1Burrell:
@iameli this might cause problems when running other application on port 3001 or 3002. I found that npx will ask to find the best available port whenever the default port is not available, which is why I think the best solution would be to not bind to any specific ports.

iameli:
This change doesn't eliminate that behavior — `create-react-app` will still prompt for a new port if there's a conflict with 3001 or 3002. This just helps ensure there isn't a conflict by default when you boot up both `player` and `explorer` at the same time.

Randy1Burrell (commented):
LGTM!

Randy1Burrell (approved):

PR #333, opened 2019-02-15
Implement error handling/display for transcoder loading

### LINKED ISSUE #330: Subgraph GraphQL Errors are being silently ignored
Errors returned from the subgraph server aren't bubbling up to the components for some reason. When GraphQL sends us an `errors` field back it gets lost somewhere in the Apollo internals. We can see them [in the Apollo client here](https://github.com/livepeer/livepeerjs/blob/2374db65bbb2e2c326e7f5a6f9aed13df3f382ba/packages/apollo/src/index.js#L335) if we log out `value.errors`:

image

image

but the client seems to completely lose track of them after that point. I can call `observer.error()` with them but that just makes it log out a "network error" and still nothing makes it to a component.

Per [the Apollo docs](https://www.apollographql.com/docs/react/features/error-handling.html), I'd expect to see a `props.error` (or maybe `props.data.error`?) available to, for example, [this enhancer function in the Explorer](https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/enhancers/index.js#L448).

**To Reproduce (required)**
1. Start up `explorer` with the environment variable `REACT_APP_LIVEPEER_SUBGRAPH` set to `https://api.thegraph.com/subgraphs/name/iameli/livepeerdev`
2. Navigate to the transcoders page
3. It won't load, nor will it provide an error to the component

iameli:
Fixes #330

All the `notifyOnNetworkStatusChange: true` changes are necessary to get
the "error" field of a graphQL query propagating all the way back to our
component, unfortunately. See discussion here:
https://github.com/livepeer/livepeerjs/issues/330

Also bumped all the Apollo versions. Didn't fix the problem but it also
didn't introduce any so I kept the upgrade.

**Screenshots (optional):**

image

**Checklist:**

- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/packages/apollo/package.json
+++ b/packages/apollo/package.json
@@ -38,12 +38,12 @@
   "dependencies": {
     "@livepeer/graphql-sdk": "^1.0.0-alpha.5",
     "@livepeer/sdk": "^1.0.0-alpha.5",
-    "apollo-cache-inmemory": "^1.3.11",
+    "apollo-cache-inmemory": "^1.4.3",
     "apollo-cache-persist": "^0.1.1",
-    "apollo-client": "^2.4.7",
-    "apollo-link-http": "^1.5.5",
-    "apollo-link-state": "^0.4.1",
-    "apollo-link": "^1.2.4",
+    "apollo-client": "^2.4.13",
+    "apollo-link": "^1.2.8",
+    "apollo-link-http": "^1.5.11",
+    "apollo-link-state": "^0.4.2",
     "graphql": "^14.0.2",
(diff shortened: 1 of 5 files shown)


iameli:
(To be clear, those screenshots are from a different branch — I didn't, like, introduce errors for the purpose of showing error handling 😃)

yondonfu (approved):
LGTM

iameli:
I'm going to revert this for now — it's caused inconsistency (#340) and "flickering" in a few areas of the UI, and there are higher-priority things to deploy.

PR #343, opened 2019-02-22
feat: use GraphQL for round initialization and estimate gas

### LINKED ISSUE #339: Update initializeRound as GraphQL mutation and use gas estimation

The round initialization transaction is currently submitted using a [SDK function call](https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/views/ProtocolStatus/enhance.js#L18) from the explorer with a hardcoded gas amount. This is inconsistent with other transactions submitted via the explorer which are modeled as GraphQL mutations. Furthermore, if the gas cost of round initialization increases such that it is greater than the hardcoded gas amount, then round initialization attempts from the explorer will fail and waste user funds.

The explorer submits round initialization transactions using a SDK function call with a hardcoded gas amount.

Model the round initialization transaction as a GraphQL mutation so the explorer can use the GraphQL mutation without touching lower level SDK function calls. Within the GraphQL mutation, gas estimation can be used (see this [example](https://github.com/livepeer/livepeerjs/blob/master/packages/graphql-sdk/src/resolvers/Mutation.js#L34)).

iameli:
Moves the round initialization logic into a GraphQL mutation

- Introduce a new GraphQL mutation, initializeRound
- Use it in the explorer
- I had to remove one line in `estimateGas`, which was capping the amount of gas that was used to estimate the amount of gas necessary. Defeated the purpose. [It appears to be an optimal property for the estimateGas call](https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethestimategas).

Initialized rounds in Rinkeby.

Fixes #339 

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/ProtocolStatus/enhance.js
+++ b/packages/explorer/src/views/ProtocolStatus/enhance.js
@@ -4,10 +4,11 @@ import {
   connectCurrentBlockQuery,
   connectCurrentRoundQuery,
   connectToasts,
+  connectInitializeRoundMutation,
 } from '../../enhancers'
 
 const mapMutationHandlers = withHandlers({
-  initializeRound: ({ toasts }) => async () => {
+  initializeRound: ({ toasts, initializeRound }) => async () => {
     try {
       toasts.push({
         id: 'initialize-round',
@@ -15,7 +16,7 @@ const mapMutationHandlers = withHandlers({
         body: 'The current round is being initialized.',
       })
       // TODO: move into graphql schema as mutation
-      await window.livepeer.rpc.initializeRound({ gas: 3200000 })
+      await initializeRound()
       toasts.push({
         id: 'initialize-round',
         type: 'success',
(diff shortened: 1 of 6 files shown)


packages/explorer/src/views/ProtocolStatus/enhance.js
```diff
@@ -4,18 +4,19 @@ import {
...
 const mapMutationHandlers = withHandlers({
-  initializeRound: ({ toasts }) => async () => {
+  initializeRound: ({ toasts, initializeRound }) => async () => {
     try {
       toasts.push({
         id: 'initialize-round',
         title: 'Initializing round...',
         body: 'The current round is being initialized.',
       })
       // TODO: move into graphql schema as mutation
```
yondonfu: We can remove this TODO now?

iameli: :+1:

yondonfu (approved):
LGTM! 🚢

PR #345, opened 2019-02-23
chore: update react, react-dom to 16.8

iameli:
Updates React and ReactDOM to 16.8 so we can use [hooks](https://reactjs.org/docs/hooks-intro.html)!

* That.
* Also uses a state hook on the front of the player just to prove we can.

Booted up the explorer and player and clicked around both. Minor-version React releases oughta be pretty stable at this point.


adamsoffer:
Woohoo! 🎣

packages/player/src/views/Landing/index.js
```diff
@@ -25,7 +25,7 @@ const enhance = connect(
 )
 
 const Landing = ({ query, changeURL }) => {
-  let searchBarRef
+  const [search, setSearch] = useState('')
```
ericxtang, line 28: Is there a difference in using `false` instead of '' here?  I remember reading some doc that used `false`...

iameli: That sets the default value for that piece of state. In this case, we want the content of the search box to start out as an empty string, then get populated as the user types in the field.

j0sh (approved):

PR #356, opened 2019-03-02
allow user to bail from transfer prompt loop

jakeburden:
In the explorer, the user is unable to bail from the transfer prompts when they click cancel.  This pull request enables the cancel button to allow the user to bail from the prompt loop.

Steps to reproduce the bug:

Click transfer:
image

In the prompt, click Cancel:
image

You'll then have to click cancel on the next prompt until you can click OK in the final prompt:
image

This pull request fixes that issue by breaking the user from the prompt loop upon clicking Cancel.

- Checks if the user has clicked `Cancel` in their transfer prompt (this returns null).
-  If the value of the user's response is null (they clicked Cancel) break from the prompt loop. 

After making my change to the promptForArgs function, I clicked the Transfer button in the explorer and checked if I could bail from the prompt loop from clicking cancel.  I then checked if I could enter values and continue to each prompt in the loop.

Sidenote:  There seems to be a separate issue on the live site where the transfer prompt adds zeros to my value.

I wrote 1 LPT as the value to transfer (I have zero LPT unbonded) and this was what the final prompt said:

image

**Checklist:**

### DIFF
--- a/packages/explorer/src/utils.js
+++ b/packages/explorer/src/utils.js
@@ -126,6 +126,7 @@ export function promptForArgs(
   const args = []
   for (const prompt of prompts) {
     const val = window.prompt(prompt.ask)
+    if (!val) break
     args.push(prompt.format ? prompt.format(val) : val)
   }
   return args


iameli (approved):
LGTM!

iameli:
Thanks!

Issue #364, opened 2019-03-22
Explorer Showing "Missed Reward Call" for every transcoder


dob:
**Describe the bug (required)**
The explorer is currently showing 1 missed reward call for every transcoder that actually has zero missed reward calls. (It is likely showing +1 in general for every transcoder)

**Expected behavior (required)**
It should show accurate reward call count

**To Reproduce (required)**
1. Go to explorer.livepeer.org/transcoders
2. View the Missed Reward Calls column, and observe 1/30 listed for all of the top transcoders.

**Device (required):**
 - Device: Brave, OS X, Macbook Pro
 
**Screenshots (optional)**

https://cl.ly/f96100e86715

**Additional context (optional)**
This is widely reported and observed in Discord. It is possibly due to a graph indexing issue, though that's just an unconfirmed theory. It could also possibly due to an OBO error in the math of the round/reward calculations.

adamsoffer:
Strange. One thing I'm noticing is that after 30 seconds or so it switches back to 0/30 and then back to 1/30 ~30 seconds later.

Looking into this.

adamsoffer:
Hey @ericxtang - how's our graph node AWS instance holding up? Anything look out of the ordinary? One thing I'm noticing is that when querying the latest round inside [graphiql](https://thegraph.livepeer.org/subgraphs/name/livepeer?query=%7B%0A%20%20round(id%3A%20%221288%22)%20%7B%0A%20%20%20%20id%0A%20%20%7D%0A%7D%0A), sometimes it returns the latest round (1288) and sometimes it returns null, so something is not right. It also seems to take longer than usual to fetch the results.

iameli:
Hrmph, will look into this first thing today. @adamsoffer Checking the instance now.

adamsoffer:
@iameli Thanks! Definitely seems to be an issue with our self-hosted graph node. I just replaced our subgraph api endpoint with the The Graph's and the issue goes away.

See here: https://build-jdelbgxfni.now.sh

Replaced:
`REACT_APP_LIVEPEER_SUBGRAPH=https://thegraph.livepeer.org/subgraphs/name/livepeer/graphql`

With this:
`REACT_APP_LIVEPEER_SUBGRAPH=https://api.thegraph.com/subgraphs/name/graphprotocol/livepeer`

adamsoffer:
[...]

PR #365, opened 2019-03-23
fix(explorer): use thegraph hosted subgraph

iameli:
There doesn't seem to be much of a reason to set this in CircleCI — it's
not a secret or anything. Let's just put the default here instead.

I deleted the variable from CircleCI.

Works around #364


adamsoffer (approved):
Makes sense. LGTM

PR #366, opened 2019-03-23
Fix missed reward call totals

adamsoffer:
Replacing the subgraph API endpoint with The Graph's fixed the stability issue we were experiencing with our self-hosted graph node yesterday, but introduced some skewed missed reward call totals due to a slight discrepancy in the data returned by The Graph's hosted subgraph and our own (round numbers were no longer getting returned in ascending order). This fix accounts for that discrepancy and fixes those inaccurate missed reward call totals.

- Ensures that only the past 30 rounds are used to calculate a transcoder's total missed reward calls.

Cross-referenced the transcoders' missed reward calls with what I'm seeing in the Livepeer subgraph. 

#364 

**Checklist:**

### DIFF
--- a/packages/explorer/src/components/TranscoderCard/index.js
+++ b/packages/explorer/src/components/TranscoderCard/index.js
@@ -31,6 +31,7 @@ const TranscoderCard: React.ComponentType<TranscoderCardProps> = styled(
   }) => {
     let missedCalls: number = 0
     if (rewards) {
+      console.log('hiii', currentRound.data.id)
       missedCalls = rewards
         // If transcoder is active and has participated in more than 30 rounds,
         // slice the last 31 rounds since we're excluding the current round.
@@ -39,6 +40,7 @@ const TranscoderCard: React.ComponentType<TranscoderCardProps> = styled(
         .filter(
           reward =>
             reward.rewardTokens === null &&
+            reward.round.id >= currentRound.data.id - 30 &&
             reward.round.id !== currentRound.data.id,
         ).length
     }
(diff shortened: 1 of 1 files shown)


packages/explorer/src/components/TranscoderCard/index.js
```diff
@@ -39,6 +39,7 @@ const TranscoderCard: React.ComponentType<TranscoderCardProps> = styled(
         .filter(
```
iameli, line 36: Hmm. So if the problem is that the rounds are being returned out of order, the `slice()` on line 38 is going to be a problem, yeah? I get what the `filter` does but if anything is filtered out, won't the resulting data have fewer than 30 results?

Could we just sort the rewards? Something like `sort((a, b) =>  b.round.id - a.round.id)` before the slice

iameli (changes requested):

packages/explorer/src/components/TranscoderCard/index.js
adamsoffer: Ah yes, good spot. We should remove the slice and sort. That should do it.

iameli: Ah yeah, even just removing the slice should work now that I think about it as long as you keep your addition to the filter in place

adamsoffer: Nice. Yup that works 👍

adamsoffer: Just pushed that up

iameli (approved):

PR #398, opened 2019-04-30
fix(apollo): fix transcoders view when on rinkeby

adamsoffer:
This PR fixes an issue that prevents the transcoders view from loading when switching to the Rinkeby network (as reported by Rayj on Discord). This issue was introduced recently when we added `orderBy` & `orderDirection` arguments to the `rewards` query. Since these arguments only exist on the subgraph schema and the explorer doesn't use the subgraph while on rinkeby, we have to make sure to extend the local schema with these same arguments otherwise the query won't resolve. That's what this PR does.

**Checklist:**

### DIFF
--- a/packages/apollo/src/index.js
+++ b/packages/apollo/src/index.js
@@ -94,8 +94,27 @@ export default async function createApolloClient(
           rewardTokens: String,
           round: Round
         }
+        enum Transcoder_orderBy {
+          id
+          active
+          ensName
+          status
+          lastRewardRound
+          rewardCut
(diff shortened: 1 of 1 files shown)


iameli:
Ah, nice fix... hm. I should really just set up a Rinkeby subgraph for this sort of thing. On the other hand, it's sort of a shame to make the subgraph a hard dependency of the explorer... mixed feelings.

Deploying this to staging now, will merge when I verify it.

iameli (approved):

PR #402, opened 2019-05-06
fix(explorer): use UnbondingForm in Account view

### LINKED ISSUE #360: Can't unbond
When trying to Unbond from the delegates account page I get an error message "e is undefined", also "withdraw stake" option is grayed out in my acccount.

yondonfu:

Resolves the `Cannot read property "toString()" of undefined error` described in [this comment](https://github.com/livepeer/livepeerjs/issues/360#issuecomment-489392439). Currently, we [directly call the unbond SDK method](https://github.com/livepeer/livepeerjs/blob/master/packages/explorer/src/views/Account/enhance.js#L161) when unbonding from the Account view, but there is no value provided for the `amount` parameter. As a result, an error is thrown when we pass [undefined into the contract call](https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L2052). This PR also changes the Account view so the UnbondingForm is popped up when the unbond button is clicked (which is what happens when a user clicks the unbond button in the Transcoders view).

Resolves issue where if the user leaves the amount field as 0 when unbonding, gas estimation will throw an error despite the UI telling the user that leaving the amount field as 0 will unbond the user's entire bonded amount. Currently, the pre-processing for using the user's entire bonded amount when the user provides an amount of 0 occurs [in the unbond SDK method](https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L2046). However, gas estimation is executed in the [GraphQL package prior to this pre-processing step](https://github.com/livepeer/livepeerjs/blob/master/packages/graphql-sdk/src/resolvers/Mutation.js#L164). As a result, if the user provides an amount of 0, gas estimation will be executed with an amount of 0 which would result in a transaction that always fails because the contract will revert if the user tries to unbond with an amount of 0.

Longer term we might want to move gas estimation for other transactions into their corresponding SDK methods as well so that gas estimation can always happen after pre-processing that might be necessary for the particular transaction.

- Pop up UnbondingForm when unbonding from Account view by pushing to history
- Move gas estimation to the unbond SDK method after the pre-processing step for the amount field

Manual testing

Fixes #360

**Checklist:**

### DIFF
--- a/packages/explorer/src/views/Account/enhance.js
+++ b/packages/explorer/src/views/Account/enhance.js
@@ -133,7 +133,9 @@ const mapMutationHandlers = withHandlers({
     }
     history.push(`#/bond/${id}`)
   },
-  unbond: ({ currentRound, me, protocol, toasts }) => async ({ id }) => {
+  unbond: ({ currentRound, history, me, protocol, toasts }) => async ({
+    id,
+  }) => {
     try {
       const { id: lastInitializedRound } = currentRound.data
       const { status, lastClaimRound } = me.data.delegator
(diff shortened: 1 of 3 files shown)


iameli (approved):
Tested, looks solid. Thanks!

PR #425, opened 2019-06-25
added new package to monorepo for 3box profiles integration

S1nus:


iameli:
Looks like there are a lot of `.swp` files included in this - vim artifacts of some sort. You could add `*.swp` them to your `~/.gitignore_global` file to stop that kind of thing from showing up in the future.

packages/profiles/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@livepeer/profiles",
+  "description": "A component library containing shared code for decentralized profile management.",
+  "version": "1.0.0",
```
iameli: This should match the version parameter in all the other `package.json` files

packages/profiles/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "main": "main/index.js",
+  "module": "lib/index.js",
+  "license": "MIT",
+  "scripts": {
+    "build": "run-p --npm-path npm build-module build-main",
+    "build-main": "cross-env NODE_ENV=production babel src --copy-files --out-dir main --ignore spec.js,test.js --plugins=@babel/plugin-transform-modules-commonjs --source-maps inline",
+    "build-module": "cross-env NODE_ENV=production babel src --copy-files --out-dir lib --ignore spec.js,test.js --source-maps inline",
+    "parcel": "parcel serve src/index.html",
+    "copy-docs": "echo 'No copy-docs script specified'",
```
iameli, line 13: don't need this

packages/profiles/src/TestComponent.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState, useEffect } from "react";
+import styled from "styled-components";
+const Box = require("3box");
```
iameli: should be `import Box from "3box"`

packages/profiles/src/TestComponent.js
```diff
@@ -0,0 +1,43 @@
...
+
+const TestComponent = styled.div`
+  color: red;
+`;
+
+export default () => {
+  const [jsontext, setJsonText] = useState("Loading...");
```
iameli: Hm, looks like this doesn't follow the formatting used in the rest of the files... you should copy the `.prettierrc` file used in the other packages into this one and then run this, should take care of it: `git diff --name-only master | grep 'js$' | xargs npx prettier --write`

packages/profiles/src/TestComponent.js
```diff
@@ -0,0 +1,43 @@
...
+  const [boxStatus, setBoxStatus] = useState("Loading status...");
+  const [boxName, setBoxName] = useState("Loading name...");
+
+  useEffect(() => {
+    setJsonText(window.web3.eth.defaultAccount);
+    Box.getProfile(window.web3.eth.defaultAccount).then(p => {
```
iameli: This code is duplicated within the `update()` function — you could just call `update()` immediately after declaring it to be a bit more DRY

iameli (changes requested):

iameli (approved):

Issue #434, opened 2019-07-25
Explorer not showing all registered transcoder nodes


dob:
**Describe the bug (required)**
It looks like our explorer isn't showing all the transcoders for some reason. It lists only 23 active. After doing a comparison with Scout, I noticed these particular nodes are missing (in my view). Note also that Scout is missing some nodes, but it's not on us to debug their index:

Explorer is missing:
0xE9E284277648fcdb09B8EfC1832c73c09b5Ecf59
0xFb9849b0B53F66B747Bfa47396964a3FA22400A0

Scout is missing:
0x9c10672cee058fd658103d90872fe431bb6c0afa
0x9e14de5cc84e437518c2c303c88570fe13f0ad15
0xd84781e1a9b74d71ea76cda8bb9f30893bfd00d1

**Expected behavior (required)**
All registered nodes should be shown.

**To Reproduce (required)**
1. Go to https://explorer.livepeer.org
2. View the list of transcoders. 
3. Observe that the above mentioned nodes are missing.

**Device (required):**
 - Device: Macbook Pro, OS X, Chrome (and Brave)
 
**Additional Context:**

Adam had mentioned that the Graph node was updated recently. This may have introduced an issue.

adamsoffer:
Eli and I upgraded the Graph Node to the latest version yesterday, which seems to now paginate graphql query results (100 results per page). Currently, the explorer queries for all transcoders and then filters out those that are registered and active client-side. Since the latest Graph Node now paginates results, that means we're excluding some from the list. I've confirmed the two missing transcoders are on the second page of results. 

I think I simple fix would be to include a filter on our transcoders query. ie:
```
{
  transcoders(where:{status: "Registered"}) {
     ...
  }
}
```

iameli:
Fixed by #435

PR #435, opened 2019-07-25
fix issue causing missing transcoders

adamsoffer:
This PR fixes an issue that caused a couple of transcoders to disappear from the transcoders view. 

- Applies a filter to the `transcoders` graphql query so that it only fetches all registered transcoders, so that none of the registered transcoders get left off the first page of results.

Tested locally and confirmed I can now see the missing transcoders.

#434 

**Screenshots (optional):**
Missing transcoders are appearing again while testing locally:
Screen Shot 2019-07-25 at 10 52 54 AM

### DIFF
--- a/packages/explorer/src/enhancers/index.js
+++ b/packages/explorer/src/enhancers/index.js
@@ -449,7 +449,7 @@ export const TranscodersQuery = gql`
   }
 
   query TranscodersQuery {
-    transcoders {
+    transcoders(where: { status: "Registered" }) {
       ...TranscoderFragment
     }
   }


dob (commented):
I'll let Eli give the go ahead before merge and deploy. But one thing to note is that I'm guessing this filter solves the issue **today** as there are less Registered transcoders than would fit in one "page" of Graph query results, but if the transcoder set size grew beyond what was returned by default within a page, then the frontend handling would still have to support pagination?

adamsoffer:
Yeah, that's right. As soon as the transcoder set increases to a number greater than 100, we'll want to implement pagination on the frontend.

iameli:
Looks good!

PR #441, opened 2019-08-08
chore(subgraph): subgraph tweaks for graph-node 0.14.0

iameli:
Bumped the graphprotocol libraries and made a few fixes for the latest version.

The cause of the current trouble was `Failed to start subgraph, code: SubgraphStartFailure, error: subgraph cannot index data from different Ethereum networks, subgraph_id: QmeBPrDCK3nc8LY9qfJZWhK9Caaq5fRFNShFkoPK1dcLAR, component: SubgraphInstanceManager`, which I think was because of the missing `mainnet` reference in the second contract definition I guess?

This is resyncing from scratch now but it seems to be working with our Parity archive node finally maybe so it's only going to take about an hour.

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -9,11 +9,11 @@
     "build": "graph build subgraph.yaml",
     "build-ipfs": "graph build --ipfs /ip4/127.0.0.1/tcp/5001 subgraph.yaml",
     "build-wast": "graph build -t wast subgraph.yaml",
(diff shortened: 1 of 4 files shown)


adamsoffer (approved):
Perfect 🚢

FYI I noticed a separate issue (that doesn't impact the explorer at all). `lastClaimRound`, `rewardCut` and `feeShare` are being indexed incorrectly. It doesn't impact the explorer because we don't actually use those fields in the transcoders view (we use `pendingRewardCut` and `pendingFeeShare`). Regardless, that should be addressed. I can submit a separate PR for that.

PR #450, opened 2019-09-04
[Ready] feat: http bearer token authentation middleware

### LINKED ISSUE #449: Api Token Authentication
we need a http bearer token authentication middleware to protect the API endpoint(s)

there is no auth checks, this is needed to ship to production

ya7ya:
allows the API to create and verify http tokens

- 
- 
- 

closes #449 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -39,6 +39,9 @@ export default async function makeApp({
   })
 
   app.use('/authtoken', authMiddleware)
+  app.get('/authtoken/test', (req, res) => {
+    res.send(req.token)
+  })
 
   if (kubeNamespace && kubeBroadcasterService) {
     const kc = new k8s.KubeConfig()
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -3,7 +3,7 @@ import 'express-async-errors' // it monkeypatches, i guess
 import morgan from 'morgan'
 import { json as jsonParser } from 'body-parser'
 import { LevelStore, PostgresStore } from './store'
-import { healthCheck } from './middleware'
+import { healthCheck, authMiddleware } from './middleware'
 import logger from './logger'
 import schema from './schema'
 import * as controllers from './controllers'
@@ -37,6 +37,8 @@ export default async function makeApp({
     req.store = store
     next()
   })
+
+  app.use('/authtoken', authMiddleware)
 
   if (kubeNamespace && kubeBroadcasterService) {
     const kc = new k8s.KubeConfig()
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -18,6 +18,7 @@ export default async function makeApp({
   listen = true,
   kubeNamespace,
   kubeBroadcasterService,
+  authEnabled,
 }) {
   // Storage init
   let store
@@ -38,8 +39,10 @@ export default async function makeApp({
     next()
   })
 
-  // HTTP bearer token middleware
-  app.use(authMiddleware({ adminOnly: 1 }))
+  if (authEnabled) {
+    // HTTP bearer token middleware
+    app.use(authMiddleware({ adminOnly: 1 }))
+  }
 
   if (kubeNamespace && kubeBroadcasterService) {
     const kc = new k8s.KubeConfig()
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -38,10 +38,8 @@ export default async function makeApp({
     next()
   })
 
-  app.use('/authtoken', authMiddleware)
-  app.get('/authtoken/test', (req, res) => {
-    res.send(req.token)
-  })
+  // HTTP bearer token middleware
+  app.use(authMiddleware)
 
   if (kubeNamespace && kubeBroadcasterService) {
     const kc = new k8s.KubeConfig()
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -19,7 +19,6 @@ export default async function makeApp({
   listen = true,
   kubeNamespace,
   kubeBroadcasterService,
-  authEnabled,
 }) {
   // Storage init
   let store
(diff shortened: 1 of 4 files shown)


packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,54 @@
+import bearerToken from 'express-bearer-token'
+import { Router } from 'express'
+import uuid from 'uuid/v4'
+import logger from '../logger'
+
+const router = Router()
```
iameli: Somehow it never occurred to me to use a router for middleware like this. Nice!

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,54 @@
...
+
+router.use(bearerToken())
```
iameli: Somehow it never occurred to me to use a router for a middleware function like this. Nice! Seems to work well.

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+/**
+ * generate token middleware
+ * @param {object} req req object
+ * @param {object} res res object
+ * @param {fn} next callback
+ */
+async function generateToken(req, res, next) {
```
iameli, line 9: I don't think at this point we ever want to generate a new token - we want all existing users to have a token already. Two scenarios:

1. The user doesn't provide a token — `req.token` is undefined. In this case, we should return `401`, and throw out the request (by not calling `next()`)
2. The user provides a token. In this case, we should check to make sure it's in our database — and if it's not, we should do a `req.store.create` with the `id` as the id they provided.

Once we've added all the existing API tokens to the database, we can then push a new update that sends `403`s if they don't send a valid token.

packages/api/src/index.js
```diff
@@ -38,6 +38,9 @@ export default async function makeApp({
     next()
   })
 
+  // HTTP bearer token middleware
+  app.use(authMiddleware)
```
iameli: I think using the same `authMiddleware` for every single request will quickly become limiting. We'll want some endpoints to be public, some to require valid API tokens (Wowza client), and some to require some other form of administrator access (Livepeer admin client). So let's leave it off here.

The other thing we can do — and I should have done this with all of our other middlewares — is put the middleware in a factory function so that we can pass various parameters to the middleware. That way we can eventually have:

```
router.get("/extremely-secret-secrets", authMiddleware({adminsOnly: true}), (req, res) => {})
```

For now, all that'll be necessary is wrapping `authMiddleware` in a big factory function so we can call it like 

```
app.use(authMiddleware())
```

All that'll be necessary for now is wrapping all of `authMiddleware` in a `function

iameli (changes requested):
Solid start, `express-bearer-token` seems very useful. Proposed one change to the structure and one change to the token validation logic.

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,74 @@
...
+  return router.use(async (req, res, next) => {
+    if (!req || !req.token) {
+      return res.sendStatus(401)
+    }
+    logger.info('authFactory params ', params)
+    let resp
+    try {
+      // check token against token DB
+      resp = await req.store.get(`apitoken/${req.token}`)
+    } catch (e) {
```
iameli, line 44: Let's make sure this still throws an error if it's anything other than a `NotFoundError` so we don't accidentally create new tokens if the database is down or something.

```
if (e.type !== "NotFoundError") {
  throw e;
}
```

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {object} req req object
+ * @param {object} res res object
+ * @param {fn} next callback
+ */
+async function generateToken(req, res, next) {
+  const id = uuid()
+  let resp = await req.store.create({
+    id,
+    kind: 'apitoken',
+    otherdata: 1,
```
iameli: This can be removed

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * generate token middleware
+ * @param {object} req req object
+ * @param {object} res res object
+ * @param {fn} next callback
+ */
+async function generateToken(req, res, next) {
+  const id = uuid()
```
iameli: For the autogen case, we want the token in the database to match up with what they provided in the `token` field, so `id` should be `req.token`.

packages/api/src/middleware/auth.js
iameli: I think all of this has been addressed except the very end of 2: "with the id as the id they provided."

packages/api/src/index.js
```diff
@@ -38,6 +39,11 @@ export default async function makeApp({
     next()
   })
 
+  if (authEnabled) {
```
iameli: Good example, but maybe comment this stuff for now? Don't think we'll implement this as a CLI parameter.

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,74 @@
...
+
+  req.token = id
+  logger.debug(`token = ${id}`)
+  return next()
+}
+
+// router.use(verifyToken)
+
+function authFactory(params) {
+  return router.use(async (req, res, next) => {
```
iameli: This shouldn't reference the `router`, as it's middleware intended to be used elsewhere. Here, we can just have

```
function authFactory(params) {
``` 

And then we'll utilize it on a per-endpoint basis - something like this should be added to `POST /stream` over in that controller:

```
router.post("/", authMiddleware(), (req, res, next) => {
```

That way we can use the same middleware with slightly different parameters for each function.

packages/api/src/middleware/auth.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * verify the http brearer token
+ * NOTE: this generates a token if no token is provided for now
+ * FIXME: stop generating tokens on the fly
+ * @param {object} req req object, this should already have req.store
+ * @param {object} res res object
+ * @param {fn} next callback
+ */
+async function verifyToken(req, res, next) {
```
iameli: Looks like this is duplicated logic from what's in the `authFactory`

iameli (changes requested):
Good improvements, added a few more comments regarding how we'll utilize this middleware on a per-endpoint basis

packages/api/src/index.js
ya7ya: how would we toggle this on and off ?  or is the auth middleware is always on ?

iameli: Yep, it'll always be on.

ya7ya: got it , done.

iameli (approved):
LGTM!

PR #451, opened 2019-09-06
Feat: API pagination for the streams endpoint

ya7ya:
adds `limit` and `offset` capabilities to the API so it won't list the entire list of streams every time. and it can be used for pagination

- updates to `leveldb` to use `limit` 
- updates to the `postgresql` to use `limit` and `offset`
- 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/store/level-store.js
+++ b/packages/api/src/store/level-store.js
@@ -39,7 +39,7 @@ export default class LevelStore {
       filter = {
         gte: prefix,
         lt: endKey,
-        limit,
+        // limit,
       }
     }
     await this.ready
@@ -53,6 +53,11 @@ export default class LevelStore {
     for await (const val of this.listStream(prefix, limit, offset)) {
       ret.push(val)
     }
+
+    if (limit > 0) {
+      ret = ret.slice(offset, offset + limit)
+    }
+
     return ret
   }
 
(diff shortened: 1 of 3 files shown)


packages/api/src/store/level-store.js
```diff
@@ -43,11 +48,16 @@ export default class LevelStore {
     }
   }
 
-  async list(prefix = '') {
+  async list(prefix = '', limit = DEFAULT_LIMIT, offset = 0) {
```
iameli, line 51: This is fine for now; it's hard for me to imagine a LevelDB API user with thousands of thousands of streams. We will eventually need a more sophisticated way to handle query continuations but I can file that as a separate enhancement.

packages/api/package.json
```diff
@@ -36,6 +36,7 @@
     "cross-env": "^5.2.0",
     "express": "^4.16.4",
     "express-async-errors": "^3.1.1",
+    "express-bearer-token": "^2.4.0",
```
iameli: Mind doing a `git rebase -i eli/api-node-prototype` and taking out the auth commits so this can merge separately?

iameli (changes requested):
Looks solid if you can rebase out the auth commits. I'll file a separate issue for a "continuation token" kind of design.

packages/api/src/store/level-store.js
ya7ya: yeah, I'll figure out a more efficient way then 👍

iameli (approved):

iameli:
LGTM! I'll merge to bypass the failing CI, hopefully fixing that today.

PR #463, opened 2019-09-18
test(api): unbreak tests

### LINKED ISSUE #455: api: unbreak tests
Two things that're broken:

* We never implemented a proper multi-store test pattern for `/stream`, only `/ingress`. So ingress's tests are really limited.
* I changed the signature for `.create(obj)` and `.update(obj)` at the last moment to make use of and require the `kind` argument. None of the tests do this.

If we fix those two things, we should be back up to pretty high test coverage!

iameli:
fixes #455


iameli:
@ya7ya @angyangie This is mostly me cleaning up some crufty old code, but could I get y'all to take a quick look at the new `stream.test.js` and see if it seems like a reasonable way to do tests for this server? We're probably going to have to write a _lot_ of these so I wanna make sure everyone's sold on the format 😃 https://github.com/livepeer/livepeerjs/blob/00a2c786ab0af55d4af732fa385ba6391555d430/packages/api/src/controllers/stream.test.js

Issue #459, opened 2019-09-16
api frontend: figure out codesharing and compilation process


iameli:
Currently the preliminary API frontend is being built off of @adamsoffer's `as/explorer-streamflow` branch. We should figure out how these projects can collaborate without breaking the monorepo.

adamsoffer:
@iameli I can move the `theme-ui` config file, compiled components, and storybook into a separate shared package and then we can just both submit PRs to that package. I was originally thinking we could use the existing Chroma package for this but to avoid confusion with the "[Chroma Design System](https://www.chromadesignsystem.com/)" perhaps wise to call it something else. `[company-name]-ui` seems to have become the defacto naming convention for brand-specific ui libraries. `livepeer-ui` sound good?

iameli:
Yeah, that would be great! The only thing that might cause problems is that (AFAICT) you currently have a separate Yarn workspace set up in `explorer-streamflow`; I had trouble getting the Yarn repo to properly symlink packages in that context. If you know an easy way around that, awesome, otherwise it oughta be possible to integrate explorer-streamflow with the rest of the monorepo without too much effort but such things can end up being tricky.

adamsoffer:
Great! Yeah, I was originally using a separate workspace for `explorer-streamflow` because I was running into some weird dependency issue on install with the `web3` package but I figured out the issue last week and now it's part of the same workspace as everything else so should be alright.

adamsoffer:
Merged in `livepeer-ui` yesterday so you no longer have to use the `as/explorer-streamflow` branch. There's a limited set of components to choose from but we can add more over time.

PR #467, opened 2019-09-24
Introduce Livepeer UI

adamsoffer:
This PR introduces the `livepeer-ui` package, a set of primitive react components built with [theme-ui](https://theme-ui.com).

The goal is for this package to be used by all Livepeer products in an effort to unify our brand and product design around a shared visual language, promote constraint-based design principles, promote consistency in how we use patterns and UI elements, and to ship products faster and more efficiently.

This package includes the following:
- An initial set of components (`Button`, `Chip`, `Card`, `List`, `ListItem`, `Spinner`, `Logo`). More to come.
- A theme object adhering to the [system-ui spec](https://system-ui.com/).
- A [storybook](https://storybook.js.org/) for viewing and interacting with components in isolation.

Addresses #459 

**Screenshots (optional):**

**Checklist:**


iameli:
Looking great! Remind me the difference between `<div sx={...}>` and `<Styled.div sx={...}>`? I think that's the one part I don't really follow so far

adamsoffer:
:)

Sure. `<Styled.div>`, `<Styled.h1>`, etc inherit styles defined inside our theme object. So if we have, for example, the following styles assigned to `h1` in our theme object then `<Styled.h1>` will inherit those styles:

```javascript
{
  fonts: {
    body: 'system-ui, sans-serif',
    heading: 'Georgia, serif',
  },
  styles: {
    h1: {
      fontSize: 32,
      fontFamily: 'heading',
      fontWeight: 'heading',
      color: 'primary',
      mt: 4,
      mb: 2,
    },
  }
}
```

I don't actually define any global style to a `div` in our theme object so I should actually remove any instances of `Styled.div`.

packages/livepeer-ui/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@livepeer/ui",
+  "version": "2.0.0-alpha.1",
+  "scripts": {
+    "dev": "run-p dev:babel",
+    "dev:babel": "babel --ignore node_modules,src/**/*.stories.tsx --out-dir dist -x '.tsx' --watch .",
```
iameli, line 7: Can we add a `dev:storybook` that runs `start-storybook`? Would be nice to run that app alongside everything else when hacking on it.

(I'm thinkin' let's just rename all the existing `start` scripts to `dev` to line up with Now's conventions but still have a nice `yarn run dev` command that starts everything. But that doesn't need to happen in this PR.)

packages/livepeer-ui/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "@types/styled-jsx": "^2.2.8",
+    "@types/theme-ui": "^0.2.2",
+    "awesome-typescript-loader": "^5.2.1",
+    "babel-plugin-inline-react-svg": "^1.1.0",
+    "babel-plugin-styled-system": "^0.0.5",
+    "npm-run-all": "^4.1.5",
+    "react-docgen-typescript-loader": "^3.1.1",
+    "typescript": "^3.5.3"
+  },
+  "license": "ISC",
```
iameli: We're MIT-licensed everywhere else, probably no reason to switch that up

iameli (changes requested):
Bumping `explorer` and `player`'s `react-scripts` to `^3.1.2` seems to unbreak the CI build - looks great other than that! Feel free to merge when that's in :shipit:

packages/livepeer-ui/package.json
adamsoffer: Yeah good call.

packages/livepeer-ui/package.json
adamsoffer: oops! Was unintended, I'll switch to MIT.

PR #473, opened 2019-09-27
[READY] cursor based pagination

### LINKED ISSUE #456: api: implement cursor tokens for pagination
Follow-up work from #451. Right now we have a limit and offset, which works great with Postgres but is currently faked with LevelDB, and won't work at all with Cloudflare KV. What we should do instead is implement a "cursor" token basically in line with what we see in Cloudflare docs here: https://developers.cloudflare.com/workers/reference/storage/listing-keys/

For LevelDB, we can also have a cursor token, but instead of being an opaque string like it is from Cloudflare we can have it be the `id` of the last document encountered in the previous set. We can then use that token to ask Level for all documents "greater than" that token to get the next "page".

ya7ya:
adds cursor based pagination to the api , [details](https://github.com/livepeer/livepeerjs/issues/456)

- 
- 
- 

Fixes #456 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -17,7 +17,12 @@ app.get('/', authMiddleware({}), async (req, res) => {
 
   const output = await req.store.list(`stream/`, cursor, limit, offset)
   res.status(200)
-  res.json(output)
+  let context = {}
+  if (output.length > 0) {
+    context.cursor = output[output.length - 1].id
+  }
+
+  res.json({ payload: output, context })
 })
 
 app.get('/:id', authMiddleware({}), async (req, res) => {
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -14,7 +14,9 @@ app.get('/', authMiddleware({}), async (req, res) => {
   let cursor = req.query.cursor
   logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
 
-  const output = await req.store.list(`stream/`, cursor, limit)
+  const resp = await req.store.list(`stream/`, cursor, limit)
+  const output = resp.data
+  const nextCursor = resp.cursor
   res.status(200)
 
   let baseUrl = new URL(
@@ -23,7 +25,7 @@ app.get('/', authMiddleware({}), async (req, res) => {
   if (output.length > 0) {
     try {
       let next = baseUrl
-      next.searchParams.append('cursor', output[output.length - 1].id)
+      next.searchParams.set('cursor', nextCursor)
       res.links({
         next: next.href,
       })
@@ -61,7 +63,7 @@ app.post('/hook', async (req, res) => {
       errors: ['missing url'],
     })
   }
-  logger.info(`got webhook: ${JSON.stringify(req.body)}`)
+  // logger.info(`got webhook: ${JSON.stringify(req.body)}`)
   // These are of the form /live/:manifestId/:segmentNum.ts
   let { pathname, protocol } = parseUrl(req.body.url)
   // Protocol is sometimes undefined, due to https://github.com/livepeer/go-livepeer/issues/1006
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -12,8 +12,10 @@ const app = Router()
 app.get('/', authMiddleware({}), async (req, res) => {
   let limit = req.query.limit
   let offset = req.query.offset
+  let cursor = req.query.cursor
+  logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
 
-  const output = await req.store.list(`stream/`, limit, offset)
+  const output = await req.store.list(`stream/`, cursor, limit, offset)
   res.status(200)
   res.json(output)
 })
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -11,12 +11,12 @@ const app = Router()
 
 app.get('/', authMiddleware({}), async (req, res) => {
   let limit = req.query.limit
-  let offset = req.query.offset
   let cursor = req.query.cursor
   logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
 
-  const output = await req.store.list(`stream/`, cursor, limit, offset)
+  const output = await req.store.list(`stream/`, cursor, limit)
   res.status(200)
+
   let context = {}
   if (output.length > 0) {
     context.cursor = output[output.length - 1].id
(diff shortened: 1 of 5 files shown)


iameli:
Looks like a good start! A few things we'll want in here...

1. The cursor will need to be returned from any `list` queries. Probably the easiest way to implement this would be to change `store.list()` to return a `[docs, cursor]` tuple, something like:

```
const [output, cursor] = await req.store.list(`stream/`, cursor, limit, offset)
```
2. That cursor will have to be communicated back to the user somehow. In the past I've seen this handled with an HTTP header: [I like the way Mixer handles it with the "link" header.](https://dev.mixer.com/guides/core/continuationtokens)

3. I kind of think this mechanism should replace "offset", which is unsupported by both LevelDB and Cloudflare Worker KV.

packages/api/src/controllers/stream.js
```diff
@@ -11,11 +11,18 @@ const app = Router()
 
 app.get('/', authMiddleware({}), async (req, res) => {
   let limit = req.query.limit
-  let offset = req.query.offset
+  let cursor = req.query.cursor
+  logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
 
-  const output = await req.store.list(`stream/`, limit, offset)
+  const output = await req.store.list(`stream/`, cursor, limit)
```
iameli: Hmm, I think that it's probably both possible that a user could provide a cursor and no limit (to just get page 2 with the default limit) and also a limit and no cursor (page 1, just give me limit number of results please). So either one of these could be undefined, so let's probably just put them in an options object? ` const output = await req.store.list(`stream/`, {cursor, limit})`

packages/api/src/controllers/stream.js
```diff
@@ -11,11 +11,18 @@ const app = Router()
...
   res.status(200)
-  res.json(output)
+
+  let context = {}
+  if (output.length > 0) {
+    context.cursor = output[output.length - 1].id
```
iameli: This might work in both of our existing cases but it won't work with other stores, including Cloudflare Worker KV. So this logic should live in the stores themselves, and we return the cursor along with the content in `req.store.list` - that's that tuple technique that I mentioned before.

packages/api/src/controllers/stream.js
```diff
@@ -11,11 +11,18 @@ const app = Router()
...
+  }
+
+  res.json({ payload: output, context })
```
iameli: Hmm, I see what you meant by returning an object now. I still kind of like the idea of returning the next page in the `link` header [like Mixer does](https://dev.mixer.com/guides/core/continuationtokens), allowing end users not to have to look at the `.payload` property all over the place, unless you can think of some other stuff we might want to return in the context? If it's just going to be this, I think I kind of prefer the header technique to keep things cleaner.

packages/api/src/store/postgres-store.js
```diff
@@ -31,11 +31,20 @@ export default class PostgresStore {
...
-    const res = await this.pool.query(
-      `SELECT data FROM ${TABLE_NAME} WHERE id LIKE $1 LIMIT $2 OFFSET $3`,
-      [`${prefix}%`, `${limit}`, `${offset}`],
-    )
+  async list(prefix = '', cursor = null, limit = DEFAULT_LIMIT) {
+    let res = null
+    if (cursor) {
+      res = await this.pool.query(
+        `SELECT data FROM ${TABLE_NAME} WHERE id LIKE $1 AND id > $2 LIMIT $3`,
+        [`${prefix}%`, `${prefix}/${cursor}`, `${limit}`],
```
iameli: I think the slash between prefix and cursor isn't necessary here, prefix should already include the slash

iameli (changes requested):
I'm still inclined to return the cursor in the `Link` header unless we can think of other stuff that we'd want to return in the context, I think

packages/api/src/controllers/stream.js
ya7ya: ah , this is much cleaner 👏 , done 👍

packages/api/src/controllers/stream.js
ya7ya: so I moved the logic to the store level and followed the cloudflare way of of returning a data array and a cursor, the only difference is the they call the data array `keys` which is specific to KV , but I thought it might be confusing so I call it `data`
```js
    return { data: [], cursor: <cursor_id> } // how it's currently implemented
    return { keys: [], cursor: <cursor_id> } // cloudflare KV
```

packages/api/src/controllers/stream.js
ya7ya: well. it's not that it'd be cleaner, but for this case the code handles the `undefined` vars, and has store specific defaults for `postgres` and `level` stores. 

putting these options in an object means we'd have to add some default option merging statement right in the next step. currently it's done at function definition 

```js
async *listStream(prefix = '', cursor, limit = DEFAULT_LIMIT) {
  // no if (!opts.limit) opts.limit = DEFAULT_LIMIT required here
  // .....
}
```

packages/api/src/store/postgres-store.js
```diff
@@ -31,12 +31,21 @@ export default class PostgresStore {
...
+        [`${prefix}%`, `${prefix}${cursor}`, `${limit}`],
+      )
+    } else {
+      res = await this.pool.query(
+        `SELECT data FROM ${TABLE_NAME} WHERE id LIKE $1 LIMIT $2`,
+        [`${prefix}%`, `${limit}`],
+      )
+    }
+    let data = res.rows.map(({ data }) => data)
+    return { data: data, cursor: data[data.length - 1].id }
```
iameli, line 52: I suspect this will break for empty returns (`cannot read property id of undefined`) — let's add a test that checks that case, too!

iameli (changes requested):
Looking good! Let's just add a test that looks at the `Link` header in the response and uses that information to request the next page of results; that'll be what we want to do in the production case. Good to go after that!

packages/api/src/store/postgres-store.js
ya7ya: done 👍

packages/api/src/controllers/stream.js
```diff
@@ -11,10 +11,28 @@ const app = Router()
...
+    `${req.protocol}://${req.get('host')}${req.originalUrl}`,
+  )
+  if (output.length > 0) {
+    try {
+      let next = baseUrl
+      next.searchParams.set('cursor', nextCursor)
+      res.links({
+        next: next.href,
+      })
+    } catch (e) {
```
iameli: Need to handle this error somehow... or just remove the try/catch and let the global error handler handle it.

iameli (changes requested):
tiny change

packages/api/src/controllers/stream.js
ya7ya: i forgot we handle uncaught errors with a global handler. I'll remove the try/catch then 👍

iameli (approved):
sweet! merging when CI passes

Issue #458, opened 2019-09-16
api: implement admin OAuth access


iameli:
We'd like the API to be able to make use of OAuth providers; this can be accomplished with a variety of libraries. This feature should make endpoints like `/api/broadcaster/status` only available to folks with a livepeer.org G Suite account.

PR #477, opened 2019-09-30
Adding GoogleAuth to `/api/stream` route + `/login` page

angyangie:

Adding GoogleAuth to `/api/stream` route + `/login` page. More details here: https://github.com/livepeer/livepeerjs/issues/458#event-2655489324

- Adding google login and logout buttons to new login page
- Adding `LP_TRUSTED_DOMAIN` and `LP_CLIENT_ID` to api `.env` file
- Adding `client-id` and `trusted-domain` flags to cli
- Adding ability to create users using GoogleAuth information & check user domains
- Requiring `livepeer.org` email admin privileges to `/api/broadcaster/status` and `/api/stream` routes.

- Locally: `yarn run dev` + `yarn run go-livepeer` +  streaming from OBS to `rtmp://localhost:3035/live `
- Running api tests: `yarn run test:local`

https://github.com/livepeer/livepeerjs/issues/458#event-2655489324

**Checklist:**


packages/api/src/index.js
```diff
@@ -34,7 +34,7 @@ export default async function makeApp(params) {
   await store.ready
 
   // Logging, JSON parsing, store injection
-  const app = express()
+  const app = express() // top level app! middleware for entire application
```
angyangie: will remove comment in later version

packages/api/src/middleware/health-check.js
```diff
@@ -12,6 +13,13 @@ const healthcheck = (req, res) => {
   res.json({ ok: true })
 }
 router.get('/healthz', healthcheck)
-router.get('/', healthcheck)
+// router.get('/', healthcheck)
+
+router.get('/', async (req, res, next) => {
+  res.sendFile(path.join(__dirname + '/index.html'))
```
angyangie: test route here until we place it somewhere

packages/api/src/middleware/index.html
```diff
@@ -0,0 +1,51 @@
...
+        token = id_token
+        console.log('ID: ' + profile.getId()) // Do not send to your backend! Use an ID token instead.
+        console.log('Name: ' + profile.getName())
+        console.log('Image URL: ' + profile.getImageUrl())
+        console.log('Email: ' + profile.getEmail()) // This is null if the 'email' scope is not present.
+      }
+
+      const CLIENT_ID = 'fill_in_client_id'
+
+      const { OAuth2Client } = require('google-auth-library')
```
angyangie: This doesn't work ... need to figure out way of adding this JS somewhere. Didn't seem to work when I added it in `health-check.js`, inside the `router.get('/',` file. Would love pointers on how to think about where to place this code so it executes correctly!

Good news is ... we do get correct info logged out on lines 26 - 29 in this file

iameli: I think everything up to this line will run on the client, and the stuff after will run on the server in some manner of auth middleware - we can take a look in a sec!

packages/api/src/controllers/stream.js
```diff
@@ -18,6 +19,18 @@ app.get('/', authMiddleware({}), async (req, res) => {
   res.json(output)
 })
 
+app.get('/login', async (req, res, next) => {
+  res.sendFile(path.join(__dirname + '/index.html'))
+})
+
+app.get(
```
iameli: Can probably take this out now that we can test against `/api/streams` above

packages/api/src/index.js
iameli: 👍

packages/api/src/index.js
```diff
@@ -34,12 +36,14 @@ export default async function makeApp(params) {
...
   app.use(healthCheck)
   app.use(morgan('combined'))
   app.use(jsonParser())
   app.use((req, res, next) => {
     req.store = store
+    req.trustedDomain = trustedDomain
```
iameli: Hmmm, we're going to have to keep doing stuff like this... maybe let's put the whole config in here `req.config = params` and then we'll have access to whatever CLI parameters we'd need in a controller, including this one.

packages/api/src/middleware/auth.js
```diff
@@ -33,6 +33,25 @@ async function generateToken(req, res, next) {
...
+    if (params.admin === true) {
+      try {
+        const user = await getUser(req, res, next)
+        if (user.domain === req.trustedDomain) {
+          req.user = user
+          return next()
+        } else {
+          res.status(403)
+          return res.json({
+            errors: ['not livepeer.org email address'],
```
iameli: Should fill this in from `req.trustedDomain`

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +73,44 @@ function authFactory(params) {
   }
 }
 
+async function getUser(req, res, next) {
+  var lpToken = req.headers.lptoken
```
iameli: Not sure I follow this logic. I think we want to keep everything in `authorization`, yeah?

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +73,44 @@ function authFactory(params) {
...
+  if (lpToken != null) {
+    req.lpToken = lpToken
+  } else {
+    lpToken = req.lpToken
+  }
+  let clientId = req.clientId
```
iameli: I think most of these can be `const`, `let` is only necessary if you're going to change what `clientId` points to later

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +73,44 @@ function authFactory(params) {
...
+  let { OAuth2Client } = require('google-auth-library')
```
iameli: This should just be `import { OAuth2Client } from 'google-auth-library` at the top of the file

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +73,44 @@ function authFactory(params) {
...
+  let client = new OAuth2Client(clientId)
+  let ticket = await client.verifyIdToken({
```
iameli: So presumably this is the line that'll fail if a hacker is trying to slip in with an invalid token. Probably we should `catch` that error and send a `403` back, yeah? This'll be a good unit test case.

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +73,44 @@ function authFactory(params) {
...
+      throw error
+    }
+    await req.store.create({
+      id: payload.sub,
+      name: payload.name,
+      email: payload.email,
+      domain: payload['hd'], // this doesn't always exist!!!
+      kind: 'user',
+    })
+    user = await req.store.get(`user/${payload.sub}`)
```
iameli: We can skip a step here - `req.store.create` returns the created document, so you can just do `user = await req.store.create({` up on line 99.

iameli (changes requested):
Solid start! Let's talk about tests for this tomorrow

packages/api/src/index.js
angyangie: good idea! This? I'll do that
```javascript
req.store = store
req.config = {
      trustedDomain: trustedDomain,
      clientId: clientId,
    }
```

packages/api/src/middleware/auth.js
angyangie: Good call. Should we specify somewhere that `trustedDomain` is required for `admin` routes? Otherwise, if `req.config.trustedDomain` is empty, this will always result in an error `user.domain === req.config.trustedDomain`. Alternatively, we can add a check for `req.config.trustedDomain != null` prior to the if else statement.

packages/api/src/middleware/auth.js
angyangie: I went a little "let" crazy. Yes, just educated myself on the distinction. Makes sense!

packages/api/src/middleware/auth.js
angyangie: Yes, that makes sense. I'll wrap this in a try/catch statement and return a 403 + `return res.json({ errors: ['invalid oauth token'] })`

packages/api/src/index.js
iameli: I was thinking that we could put the whole `params` variable in there, which would have everything that we defined in the `makeApp` function there

packages/api/src/middleware/auth.js
iameli: I'm cool with it being required for now, we're the only real users of this thing. In the future we could add an option to disable auth entirely or some such.

packages/api/src/index.js
angyangie: including all the kubeApi variables?

packages/api/src/middleware/auth.js
angyangie: it doesn't seem like this is the case for some reason 🤔

packages/api/src/middleware/auth.js
angyangie: hmm we have authorization set to have a Bearer token in `test-helpers.js`, `authorization: `Bearer ${this.apiKey}`. Do we want to change that to only hold our `lptoken`?

packages/api/package.json
```diff
@@ -24,11 +24,11 @@
     "deploy:prerelease": "gsutil rsync ./docs gs://docs.livepeer.live",
     "dev:live-server": "live-server --port=3005 docs",
     "dev:redoc": "nodemon -w src/schema/schema.yaml -x npm run prepare:redoc",
-    "dev:server": "nodemon -w src -x node -r esm src/cli.js",
+    "dev:server": "nodemon -w src -x node -r esm src/cli.js --trusted-domain=livepeer.org --client-id=211814034878-r7iugl9pa9fo2shd2pvmmadjatd5o4af.apps.googleusercontent.com",
```
iameli: Now that we have the `.env` file, all of this can be added there in the form `LIVEPEER_TRUSTED_DOMAIN` et al.

packages/api/src/index.js
iameli: 👋

packages/api/src/middleware/auth.js
```diff
@@ -33,6 +34,30 @@ async function generateToken(req, res, next) {
...
+          return res.json({
+            errors: [`not ${req.config.trustedDomain} email address`],
+          })
+        }
+      } catch (error) {
+        res.status(403)
+        return res.json({ errors: ['not logged in', error.toString()] })
+      }
+    } else if (!req.headers.authorization.startsWith('Bearer')) {
+      return next()
```
iameli: Hmm, I don't think this accomplishes the "admins can look at everything" case. This seems to grant full access to all `authMiddleware({})` endpoints when the user provides just nonsense in the `Authorization` header:

```javascript
fetch("/stream/secret-stream-id", {
  headers: {
    authorization: "nonsense"
  }
})
```

^^ Pretty sure I just hacked my way into that stream.

packages/api/src/middleware/auth.js
```diff
@@ -33,6 +34,30 @@ async function generateToken(req, res, next) {
...
+          req.user = user
+          return next()
+        } else {
+          res.status(403)
+          return res.json({
+            errors: [`not ${req.config.trustedDomain} email address`],
+          })
+        }
+      } catch (error) {
+        res.status(403)
```
iameli, line 54: Code coverage says these two lines aren't hit, and this seems like a pretty important test case - this would be when a user makes a request to an admin-secured endpoint with a totally nonsense JWT.

packages/api/src/middleware/auth.js
```diff
@@ -54,5 +79,46 @@ function authFactory(params) {
   }
 }
 
+async function getUser(req, res, next) {
+  var googleAuthToken = req.headers.authorization
+  if (googleAuthToken != 'null') {
```
iameli: This might work but isn't a very idiomatic check - I think you could just do`if (googleAuthToken) {` to check for truthiness here.

But I still don't follow this logic - if `req.headers.authorization` is null, surely we don't need to move forward with validating a token or creating a user? They haven't provided us with any credentials.

iameli (approved):
Looking solid, just a few things I don't yet follow

packages/api/package.json
angyangie: Cool, definitely can do that. Didn't understand the `.env` file before. Is this the best way of calling/setting the variables?
```javascript
app.use((req, res, next) => {
    req.store = store
    params.clientId = process.env.CLIENT_ID
    params.trustedDomain = process.env.TRUSTED_DOMAIN
    req.config = params
    next()
  })
```

iameli: Nah, you don't even have to do that — [we're using a yargs feature to automatically populate command-line parameters from the environment.](https://yargs.js.org/docs/#api-envprefix). So adding

```
LP_TRUSTED_DOMAIN=livepeer.org
```

has exactly the same effect as running the server with
```
--trusted-domain=livepeer.org
```

iameli: Which is to say you can just move that bit from `package.json` to `.env` - nothing else has to change

angyangie: Thank you for the link and explanation! I see how it works now. Fun JS/Yargs magic 🎉

packages/api/src/middleware/auth.js
angyangie: Not good. I will fix this!

packages/api/src/middleware/auth.js
angyangie: I see where the confusion here on my end came from. I had my prior HTML page not sending `req.headers.authorization` every time, only if on the login page. In that scenario, we would not always have the authorization token - it would only have it in the `/login` page when the user logged in.

We're providing the authorization header on every request we're sending on the front-end now. You're right that we don't need to do much other than return "null" if header isn't there!

packages/api-frontend/pages/login.tsx
```diff
@@ -0,0 +1,82 @@
...
+        />
+        <GoogleLogout
+          clientId={clientId}
+          buttonText="Logout"
+          onLogoutSuccess={logout}
+        />
+      </Flex>
+    </Layout>
+  )
+  // return (
```
iameli: This can go away

packages/api/src/middleware/auth.js
```diff
@@ -33,6 +34,37 @@ async function generateToken(req, res, next) {
...
+          res.status(403)
+          return res.json({
+            errors: [`not ${req.config.trustedDomain} email address`],
+          })
+        }
+      } catch (error) {
+        res.status(403)
+        return res.json({ errors: ['not logged in', error.toString()] })
+      }
+    } else if (!req.headers.authorization.startsWith('Bearer')) {
```
iameli: This will throw an error if there's no `authorization` header at all, we could do something like

```
} else if (req.headers.authorization && !req.headers.authorization.startsWith('Bearer')) {
```

packages/api/src/middleware/auth.js
```diff
@@ -33,6 +34,37 @@ async function generateToken(req, res, next) {
...
+      const user = await getUser(req, res, next)
```
iameli: This should probably get a `try{}` similarly to the previous block or we'll error out of this function and send a `500` when we hit this error case: https://github.com/livepeer/livepeerjs/pull/477/files#diff-4c61da62170e37facb95e23c25f8c602R105

iameli (changes requested):
Just minor changes to the auth logic.

I'm getting this error trying to log in using Brave, but it works with Firefox:

image

Happy to take a look at tomorrow!

packages/api/src/middleware/auth.js
angyangie: That's righttt

packages/api/src/middleware/auth.js
angyangie: That's right. We want to return a 403 here. I'll add it. I hadn't thought of the hacker scenario you mentioned previously ... that's the only scenario in which someone would have a token not starting with "Bearer" who wasn't logged into Google.

iameli (approved):

iameli:
LGTM!

Issue #472, opened 2019-09-27
api: handle `width: 0` case from Wowza 4.7.5


iameli:
@iameli commented on [Fri Sep 27 2019](https://github.com/livepeer/livepeer-wowza/issues/37)

PR #486, opened 2019-10-11
Adding proper handling to incoming width/height 0 from wowza templates

angyangie:

Adding a 16/9 or 9/16 multiplier to any height or width coming in as 0 from incoming enabled wowza rendition templates. 

- Added tests to output 4 correct rendition templates for various encodings coming in with heights / widths = 0

https://github.com/livepeer/livepeerjs/issues/472

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/wowza-hydrate.test-data.json
+++ b/packages/api/src/controllers/wowza-hydrate.test-data.json
@@ -32,7 +32,7 @@
       "profileDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/profiles",
       "templateDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/templates",
       "licensesInUse": 0,
-      "templatesInUse": "${SourceStreamName}.xml, transrate.xml",
+      "templatesInUse": "${SourceStreamName}.xml,transrate.xml",
       "createTemplateDir": false,
       "liveStreamTranscoder": "transcoder"
     },
--- a/packages/api/src/controllers/wowza-hydrate.test-data.json
+++ b/packages/api/src/controllers/wowza-hydrate.test-data.json
@@ -32,7 +32,7 @@
       "profileDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/profiles",
       "templateDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/templates",
       "licensesInUse": 0,
-      "templatesInUse": "${SourceStreamName}.xml,transrate.xml",
+      "templatesInUse": "${SourceStreamName}.xml, transrate.xml",
       "createTemplateDir": false,
       "liveStreamTranscoder": "transcoder"
     },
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/wowza-hydrate.test-data.json
```diff
@@ -32,7 +32,7 @@
       "profileDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/profiles",
       "templateDir": "${com.wowza.wms.context.VHostConfigHome}/transcoder/templates",
       "licensesInUse": 0,
-      "templatesInUse": "${SourceStreamName}.xml,transrate.xml",
```
angyangie: @iameli I removed this so we could test the `'no template found from templatesInUse'` error scenario, but am not sure if this is the default coming in with Wowza. If it is, I can add it back and remove the extra test! All in an effort to get closer to 100% test coverage for this file.

packages/api/src/controllers/wowza-hydrate.test.js
```diff
@@ -29,6 +29,38 @@ describe('wowzaHydrate', () => {
...
+    expect(newStream.presets).toEqual([
+      'P720p60fps16x9',
+      'P360p30fps16x9',
+      'P240p30fps16x9',
+      'P144p30fps16x9',
+    ])
+    expect(newStream.renditions).toEqual({
+      width_height_test_source:
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/source.m3u8',
+      width_height_test_720p:
```
angyangie, line 47: Interesting to note here @iameli ... if a 720p rendition template comes in, the `P720p60fps16x9` rendition will always get chosen. The `P720p30fps16x9` will never be chosen, given that we're selecting based on `if (thisDiff < diff) {` .... diff will always be 0 once we get to this line after selecting `P720p60fps16x9`. Do we see this as a problem?

packages/api/src/controllers/wowza-hydrate.test-data.json
iameli: Easy enough to generate a version with `transrate` and a version without — 
```javascript
const stream = require('./wowza-hydrate.test-data.json');
const streamWithTransrate = JSON.parse(JSON.stringify(stream)); // "deep copy" but I'm lazy
streamWithTransrate.wowza.transcoderAppConfig.templatesInUse = "${SourceStreamName}.xml,transrate.xml"
```

Then we can have test cases both `${SourceStreamName}.xml` and the fallback!

packages/api/src/controllers/wowza-hydrate.test.js
iameli: Oh hmmmmmmmm good point. FPS is tricky, 'cause I think Wowza just does the same FPS as the input stream but our presets require us to pick ahead of time. Wondering if for now we should just do 30fps for maximum compatibility...

iameli (changes requested):
Looks solid, one test to tweak.

packages/api/src/controllers/wowza-hydrate.test-data.json
angyangie: Good idea! Actually, just figured out I don't need to do that. I'm already changing the stream name for that test `stream.name = 'fake_name'`, so no reason it should find a template name even if `transrate.xml` is included. Thank you Eli

packages/api/src/controllers/wowza-hydrate.test.js
angyangie: That sounds fine to me. Would that mean removing `P720p60fps16x9` from the `VIDEO_PROFILES`?

packages/api/src/controllers/wowza-hydrate.test-data.json
```diff
@@ -32,7 +32,7 @@
...
+      "templatesInUse": "${SourceStreamName}.xml, transrate.xml",
```
iameli: I'm nervous about this space right here, I don't think Wowza includes that

packages/api/src/controllers/wowza-hydrate.test.js
iameli: Something like that... I'll open a ticket, doesn't need to block this PR

packages/api/src/controllers/wowza-hydrate.test.js
```diff
@@ -29,6 +29,38 @@ describe('wowzaHydrate', () => {
...
+      width_height_test_240p:
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/P240p30fps16x9.m3u8',
+      width_height_test_160p:
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/P144p30fps16x9.m3u8',
+    })
+  })
+
+  it('should fail if stream name does not equal any template name', () => {
+    stream.name = 'fake_name'
+    try {
```
iameli, line 61: This will return a false positive in the event that `wowzaHydrate` _doesn't_ throw an error - the `catch` block will never be hit and the test will complete without incident. Easy enough fix:

```javascript
    let message;
    try {
      const newStream = wowzaHydrate({ ...stream })
    } catch (e) {
      message = e.message;
    }
    expect(message).toEqual('no template found from templatesInUse')
```

iameli (changes requested):
looks good, minor fixes

packages/api/src/controllers/wowza-hydrate.test.js
angyangie: Good catch!

iameli (approved):

iameli:
LGTM!

PR #490, opened 2019-10-17
feat(player): add mux data integration to track video playback analytics

ericxtang:

Media player integration to mux data.  This helps us track video playback metrics.

- Added mux initialization
- Added mux keys (development env)

Local testing


packages/player/src/index.js
```diff
@@ -76,6 +77,13 @@ const trackingId = process.env.REACT_APP_GA_TRACKING_ID
       window.gtag('config', trackingId)
     }
     script.src = `https://www.googletagmanager.com/gtag/js?id=${trackingId}`
+
+    // Mux
+    script = document.createElement('script')
```
iameli, line 82: `script` is defined as a `const` up on line 67 there, so assigning a value to it is what's causing your CI failure. Can either change it to `let` or make a new variable

packages/player/src/views/Channel/index.js
```diff
@@ -50,6 +50,24 @@ export default class Channel extends Component {
     this.setState({ didCopy: true })
   }
 
+  componentDidUpdate = () => {
```
iameli: I'm nervous about this function getting called more frequently than we'd want when other properties of the `Channel` get updated, with `window.mux.monitor` then getting called multiple times for the same video. Perhaps try triggering it in the `ref` callback function of the `<VideoPlayer>` element? Theoretically that should get called every time a new `<video>` element is booted up.

```javascript
          <VideoPlayer
            
            // ...

            ref={ref => {
              if (!ref) {
                return
              }
              // do stuff here
            }}
          />
```

iameli (changes requested):
Looks solid, two minor changes assuming I'm right about `ref`

packages/player/src/index.js
ericxtang: Updated to `let` from `const`

packages/player/src/views/Channel/index.js
```diff
@@ -63,6 +63,9 @@ export default class Channel extends Component {
     if (redirect) {
       return <Redirect to={`/play?url=${encodeURIComponent(redirect)}`} />
     }
+
+    window.muxPlayerInitTime = Date.now()
```
iameli: I'm trying to think if there's an advantage to this over just having `player_init_time: Date.now()` in the Mux init code? `render` functions don't make any guarantees as to how often they'll be called, so this could potentially be updated many times before `window.mux.monitor` gets called. In any event, the delta between the original page render and the `window.mux.monitor` call shouldn't be more than a few seconds anyway - can't imagine it'd make a big difference.

ericxtang: Lol you are right - I was copying from the Mux doc and didn't think too deeply about this.  Yeah I'll fix it.

packages/player/src/views/Channel/index.js
```diff
@@ -164,6 +165,26 @@ export default class Channel extends Component {
...
+              // Mux Data
+              if (typeof window.mux !== 'undefined') {
+                window.mux.monitor('.video-react-video', {
+                  debug: true,
+                  data: {
+                    env_key: process.env.REACT_APP_MUX_ENV_KEY,
+
+                    // Metadata
+                    player_name: 'Media Player Main',
+                    player_init_time: window.muxPlayerInitTime,
```
iameli: i imagine this should be `Date.now()`

iameli (changes requested):

iameli (approved):
LGTM!

Issue #452, opened 2019-09-16
api: implement json-schema validation & swagger documentation for /streams


iameli:
We can model this off of the deprecated `/ingress` endpoint, which was fully set up in this way but isn't useful anymore.

This ticket includes:
* Writing a json-schema for the "Stream" object. For the "Wowza" portion, we can consult `http://wowza-docs.livepeer.live/index.html` for application transcoder settings. (If you don't yet have access to those docs, Eli can get you access)
* Enforcing that json-schema through validation middleware, again see the `/ingress` example.
* Adding relevant endpoint documentation in `schema.yaml`.

PR #494, opened 2019-10-28
adding swagger schema to api

angyangie:
Adding swagger schema to LivepeerJS `/stream` endpoint.

**Specific updates**
https://github.com/livepeer/livepeerjs/issues/452

- Added tests
- Ran livepeerjs locally

https://github.com/livepeer/livepeerjs/issues/452

**Checklist:**

### DIFF
--- a/packages/api/src/middleware/validators.js
+++ b/packages/api/src/middleware/validators.js
@@ -82,7 +82,7 @@ export const validatePut = name => {
     if (!validate(data)) {
       res.status(422)
       return res.json({
-        errors: validate.errors.map(({ message }) => message),
+        errors: validate.errors.map(({ message }) => JSON.stringify(message)),
       })
     }
 
--- a/packages/api/src/middleware/validators.js
+++ b/packages/api/src/middleware/validators.js
@@ -12,8 +12,11 @@ export const validatePost = name => {
       return
     }
     const oldRequired = node.required
+    if (!oldRequired) {
+      return
+    }
     node.required = oldRequired.filter(key => {
-      if (node.properties[key].readOnly === true) {
+      if (node.properties && node.properties[key].readOnly === true) {
         delete node.properties[key]
         return false
       }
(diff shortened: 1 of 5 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -85,6 +135,53 @@ paths:
...
+                items:
+                  $ref: '#/components/schemas/stream'
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/error'
+
   /ingress/{id}:
```
ya7ya: nice work :+1: , `/ingress/{id}` endpoint doesn't have a title in the generated docs, this can be solved by adding `description` or `summary` field

ya7ya (commented):
great start :+1:

packages/api/src/schema/schema.yaml
angyangie: Thank you @ya7ya ! Good feedback. Since that's an unused endpoint though, I think we can actually delete that one altogether. @iameli ?

iameli: Yeah, all of the ingress resources can be removed from documentation, their controller doesn't even exist anymore

packages/api/src/middleware/validators.js
```diff
@@ -26,7 +29,7 @@ export const validatePost = name => {
     if (!validate(body)) {
       res.status(422)
       return res.json({
-        errors: validate.errors.map(({ message }) => message),
+        errors: validate.errors.map(message => message),
```
iameli, line 32: This is fine, I like getting more error detail back from the validator, but let's do `message => JSON.stringify(message)` so we're at least confident that the field is a string.

iameli (changes requested):
Sweet! Just minor changes.

Let's make sure one or both of us runs through the Wowza workflow after this rolls out to staging, that's the only real critical path for this work at the moment

iameli (approved):
LGTM!

We can test this in staging during our meeting tomorrow.

PR #496, opened 2019-10-29
Ar/add user id to stream object

angyangie:
Adds a `userId` field to the stream object, and populates it when user is logged through gmail. 

Added tests + local testing

https://github.com/livepeer/livepeerjs/issues/488

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.test.js
+++ b/packages/api/src/controllers/stream.test.js
@@ -91,6 +91,27 @@ describe('controllers/stream', () => {
       expect(streams.length).toBeLessThan(11)
     })
 
+    it('should create a stream', async () => {
+      await server.store.create(user)
+      const res = await client.post('/stream', { name: 'test-stream' })
+      expect(res.status).toBe(201)
+      const stream = await res.json()
+      expect(stream.id).toBeDefined()
+      expect(stream.kind).toBe('stream')
+      expect(stream.name).toBe('test-stream')
+      expect(stream.userId).toBe('mock_sub')
+      const document = await server.store.get(`stream/${stream.id}`)
+      expect(document).toEqual(stream)
+    })
+
+    it('should accept empty body for creating a stream', async () => {
+      await server.store.create(user)
+      const res = await client.post('/stream')
+      expect(res.status).toBe(201)
+      const stream = await res.json()
+      expect(stream.id).toBeDefined()
+    })
+
     it('should not get all streams', async () => {
       user = {
         id: 'mock_sub',
@@ -170,6 +191,7 @@ describe('controllers/stream', () => {
       expect(stream.id).toBeDefined()
       expect(stream.kind).toBe('stream')
       expect(stream.name).toBe('test_stream')
+      expect(stream.userId).toBe('')
       const document = await server.store.get(`stream/${stream.id}`)
       expect(document).toEqual(stream)
     })
--- a/packages/api/src/controllers/stream.test.js
+++ b/packages/api/src/controllers/stream.test.js
@@ -184,14 +184,14 @@ describe('controllers/stream', () => {
       })
     })
 
-    it('should create a stream', async () => {
-      const res = await client.post('/stream', { ...mockStream })
+    it('should create a stream, no `user` registered', async () => {
+      const res = await client.post('/stream', { name: 'test-stream' })
       expect(res.status).toBe(201)
       const stream = await res.json()
       expect(stream.id).toBeDefined()
       expect(stream.kind).toBe('stream')
-      expect(stream.name).toBe('test_stream')
-      expect(stream.userId).toBe('')
+      expect(stream.name).toBe('test-stream')
+      expect(stream.userId).toBe(client.apiKey)
       const document = await server.store.get(`stream/${stream.id}`)
       expect(document).toEqual(stream)
     })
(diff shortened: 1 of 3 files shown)


iameli (changes requested):
This looks solid, but I think we should go one step further here - API token requests should also create a user the same way that administrative users create a user. That way we can start to set up a record of which API tokens are associated with which users, which users are currently streaming, that sort of thing.

Basically the process will be 

1. After you've retrieved/created an API token, check to see if that API token has a userId field.
  * If so, retrieve it.
  * If not, create an empty user and modify the API token record to include the userId.
2. Populate `req.user` with the appropriate user record.

angyangie:
@iameli sounds like a great addition. We have two options:
1. Make the `id` field on `user` be the same as the `id` on the token object
2. Autogenerate an `id` on `user` object, and copy to additional `userId` field on token object

Which did you have in mind?

iameli:
@angyangie 2 - we need the ability to revoke access tokens while leaving the user object intact.

packages/api/src/middleware/auth.js
```diff
@@ -35,8 +43,10 @@ async function generateToken(req, res, next) {
 function authFactory(params) {
   return async (req, res, next) => {
     if (params.admin === true) {
+      // if admin credentials required, use google auth to validate admin access
       try {
-        const user = await getUser(req, res, next)
+        const user = await getUserWithGoogleAUth(req, res, next)
```
iameli: `AUth` typo

packages/api/src/middleware/auth.js
```diff
@@ -18,6 +18,14 @@ async function generateToken(req, res, next) {
     return res.sendStatus(406)
   }
 
+  await req.store.create({
```
iameli: Gots to assign a new UUID here and save that onto the API token, per discussion on the ticket

iameli (changes requested):

packages/api/src/controllers/stream.test.js
```diff
@@ -3,9 +3,31 @@ import { TestClient, clearDatabase } from '../test-helpers'
 import uuid from 'uuid/v4'
 
 let server
+let mockStream
 
 beforeAll(async () => {
   server = await serverPromise
+  mockStream = require('./wowza-hydrate.test-data.json').stream
+  delete mockStream.id
```
iameli: Same deal here as on the other PR - this will delete those fields for every other unit test that runs. (Possibly. Jest's actual execution patterns are kinda murky.) Good to clone the object before modifying it.

packages/api/src/controllers/stream.test.js
```diff
@@ -179,6 +229,45 @@ describe('controllers/stream', () => {
...
+      expect(stream.userId).toBe(tokenObject.userId)
+      const document = await server.store.get(`stream/${stream.id}`)
+      expect(document).toEqual(stream)
+    })
+
+    it('should create a stream, token without userId already registered', async () => {
+      const token = {
+        id: client.apiKey,
+        kind: 'apitoken',
+        userId: '',
```
iameli: I think this should just be omitted? I sure hope we don't end up with any empty-string userIds in the database.

packages/api/src/middleware/auth.js
```diff
@@ -18,14 +19,20 @@ async function generateToken(req, res, next) {
     return res.sendStatus(406)
   }
 
-  let resp = await req.store.create({
-    id,
-    kind: 'apitoken',
-  })
+  try {
+    await req.store.create({
```
iameli, line 32: I don't think this intermediate step is necessary - we can just create the `user` prior to creating the `apitoken` and save ourselves the trouble of having this in-between state with the `apitoken` has a empty-string userId. This'll be especially important as we move to Cloudflare Worker KV where we won't always have strong read-after-write consistency; this empty record could stick around for longer than we'd prefer.

packages/api/src/middleware/auth.js
```diff
@@ -74,27 +85,71 @@ function authFactory(params) {
...
-      return await generateToken(req, res, next)
+      tokenObject = await generateToken(req, res, next)
     }
 
+    if (tokenObject) {
+      // if apiToken provided, get or create user with apiToken as user `id`
+      try {
+        const user = await getUserWithApiToken(req, tokenObject)
+        req.user = user
+        if (tokenObject.userId == '') {
```
iameli: As above, can cut out this step by creating the user first. Though we will still need replace logic here for when it's an old `tokenObject` that doesn't have a `userId` quite yet.

iameli (changes requested):
Looks cool, just suggested a tweak to the workflow so avoid the empty-string-user case

packages/api/src/controllers/stream.test.js
angyangie: Sure thing! Since that's the stream we're using for all tests but one, I'll just clone it for the `should not accept additional properties for creating a stream` test. `mockStream` should be the standard for all tests. I'll change the name to `postMockStream` to perhaps convey the idea better

packages/api/src/controllers/stream.test.js
angyangie: Sure thing. I hope not!

packages/api/src/middleware/auth.js
angyangie: Hmm perhaps I interpreted this differently: 
```
We need the ability to revoke access tokens while leaving the user object intact.
```
I assumed "revoke" meant we would delete the access token altogether. So theoretically, we'd have a user in our DB with no API Token. If that's not the case, then the logic here can be simpler. I had it setup with user being created first in the previous local interation of this - I'll set it up that way again :)

iameli: Right — we don't yet have the ability to issue new API tokens for a given user but once we do, your description will line up a bit better.

packages/api/src/middleware/auth.js
```diff
@@ -74,27 +96,55 @@ function authFactory(params) {
...
       if (e.type !== 'NotFoundError') {
         throw e
       }
 
       logger.warn('api Token not found... generating one')
-      return await generateToken(req, res, next)
+      tokenObject = await generateUserAndToken(req, res, next)
     }
 
+    if (tokenObject && tokenObject.userId == '') {
```
iameli: I think this'll fail for most of the existing database — they don't have userIds at all, so those fields would be `undefined`. This check could just be `if (tokenObject && !tokenObject.userId) {` — that'll catch both empty strings and undefined/null values.

iameli (approved):
LGTM after one small change

packages/api/src/middleware/auth.js
angyangie: ohhh right. good catch!

Issue #497, opened 2019-11-01
Match incoming FPS and aspect ratio from the Wowza plugin


angyangie:
**What's the problem? (required)**

- We need to be able to match incoming stream fps to an existing profile in `VIDEO_PROFILES` in `packages/sdk/src/index.js`
- We should be able to derive aspect ratio from incoming stream height and width.

**What's the current behavior? (required)**
We select profiles for use taking into consideration only height and width - all profiles at the moment output video in 30fps. We'd like to, at least in the immediate future, output 25fps as well. 

**Do you have any personal experiences dealing with this problem? (optional)**
This should unblock Camcast requirement of 25fps.

**Describe a possible solution you've considered (optional)**
- As a temporary solution, add 25fps profiles to `VIDEO_PROFILES`, mirroring profiles that exist in `go-livepeer`: https://github.com/livepeer/go-livepeer/pull/1148 (see comments in link for more long-term solutions).
- Add `sourceInfo` object to the request body coming from livepeer-wowza, to be accessed as such:  `req.body.wowza.sourceInfo`, with the following fields:
```json
"sourceInfo": {
   "width": 1920,
   "height": 1920,
   "fps": 25
}
```
- In addition to using height / width, also use incoming `sourceInfo.fps` to select correct present.
- When `width` or `height` in `transcoderTemplateAppConfig` template comes back as zero, use aspect ratio to fill in correct missing number.

PR #498, opened 2019-11-01
Match incoming FPS and aspect ratio

angyangie:
- We need to be able to match incoming stream fps to an existing profile in `VIDEO_PROFILES` in `packages/sdk/src/index.js`
- We should be able to derive aspect ratio from incoming stream height and width.

- As a temporary solution, add 25fps profiles to `VIDEO_PROFILES`, mirroring profiles that exist in `go-livepeer`: https://github.com/livepeer/go-livepeer/pull/1148 (see comments in link for more long-term solutions).
- Add `sourceInfo` object to the request body coming from livepeer-wowza, to be accessed as such:  `req.body.wowza.sourceInfo`, with the following fields:
```json
"sourceInfo": {
   "width": 1920,
   "height": 1920,
   "fps": 25
}
```
- In addition to using height / width, also use incoming `sourceInfo.fps` to select correct present.
- When `width` or `height` in `transcoderTemplateAppConfig` template comes back as zero, use aspect ratio to fill in correct missing number.

Test file

https://github.com/livepeer/livepeerjs/issues/497

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/wowza-hydrate.js
+++ b/packages/api/src/controllers/wowza-hydrate.js
@@ -43,7 +43,6 @@ export default stream => {
   let presets = []
   const encodeNameToRenditionName = {}
   const aspectRatio = sourceInfo.width / sourceInfo.height
-  const fps = sourceInfo.fps
   for (const encode of enabledEncodes) {
     let { width, height, name, streamName, videoCodec } = encode
     let renditionName = replaceStreamName(streamName, stream.name)
@@ -62,22 +61,25 @@ export default stream => {
     // TODO: do we need to validate whether or not the incoming height/width fit the aspect ratio?
 
     const wowzaSize = width * height
+    const wowzaFps = sourceInfo.fps
     let diff = Infinity
+    let fpsDiff = Infinity
     let foundPreset = null
+
     for (const preset of PRESETS) {
       const [livepeerWidth, livepeerHeight] = preset.resolution
         .split('x')
         .map(x => parseInt(x))
       const livepeerSize = livepeerWidth * livepeerHeight
       const thisDiff = Math.abs(livepeerSize - wowzaSize)
-      if (thisDiff < diff && fps == preset.framerate) {
-        // TODO : need to account for when framerate isn't found in VIDEO_PRESETS.
-        // - However, if we'll implement long-term solution soon after this, like enabling full transcoder
-        //   configurability throughout the network, perhaps don't have to, since only Camcast is using this.
+      const thisFpsDiff = Math.abs(preset.framerate - wowzaFps)
+      if (thisDiff < diff || (thisDiff == diff && thisFpsDiff < fpsDiff)) {
         foundPreset = preset.name
         diff = thisDiff
+        fpsDiff = thisFpsDiff
       }
     }
+
     if (!foundPreset) {
       throw new Error(
         `couldn't find Livepeer preset for Wowza encode: ${JSON.stringify(
--- a/packages/api/src/controllers/wowza-hydrate.js
+++ b/packages/api/src/controllers/wowza-hydrate.js
@@ -1,6 +1,6 @@
 import { VIDEO_PROFILES } from '@livepeer/sdk'
 
-const PRESETS = Object.values(VIDEO_PROFILES).filter(p => p.framerate === 30)
+const PRESETS = Object.values(VIDEO_PROFILES)
 
 /**
  * Replace a string containing "${SourceStreamName}" with the appropriate stream name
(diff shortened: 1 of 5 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -54,10 +56,13 @@ components:
               type: object
             streamNameGroups:
               type: array
+            sourceInfo:
+              type: object
```
iameli, line 60: I think we should fully define the schema here; the other properties in here are Wowza-internal blobs, so we didn't do a strict schema for forwards-compatibility, but the contents of this field are entirely up to us so we may as well define the fields explicitly.

packages/api/src/controllers/wowza-hydrate.test-data.json
```diff
@@ -2732,6 +2737,11 @@
             }
           ]
         }
+      },
+      "sourceInfo": {
```
iameli, line 2741: Let's match the rest of the Camcast settings for this second test case - should be 1080p25.

packages/api/src/controllers/wowza-hydrate.test.js
```diff
@@ -32,7 +32,77 @@ describe('wowzaHydrate', () => {
...
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/source.m3u8',
+      transrate_360p:
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/P360p30fps16x9.m3u8',
+      transrate_160p:
+        '/stream/de7818e7-610a-4057-8f6f-b785dc1e6f88/P144p30fps16x9.m3u8',
+    })
+  })
+
+  it('should correctly determine renditions and presets 30fps', () => {
+    stream.name = 'width_height_test'
```
iameli: I get nervous about directly modifying the stream object like this, as that data sticks around between tests and could theoretically cause tests to pass or fail if they're reordered. `const newStream = JSON.parse(JSON.stringify(stream))` is a hacky-but-totally-fine-for-a-test way to do a deep clone prior to doing anything with it. Or if you wanna do it the "right" way...

```javascript
const newStream = {
  ...stream,
  name: "width_height_test",
  wowza: {
    ...stream.wowza,
    sourceInfo: {
      ...stream.wowza.sourceInfo
      fps: 50
    }
  }
}
```

But that's not so fun 😄

iameli (changes requested):
Solid, a few changes.

packages/api/src/schema/schema.yaml
angyangie: good point - I'll define it!

iameli:
@angyangie I just pushed a tiny change to master to ignore this field so that it doesn't break the version of LivepeerWowza I'm releasing just now. https://github.com/livepeer/livepeerjs/commit/dee9a37fda2e1a9bb522de2291bdbb635179deb5

Let me know when this is ready and I'll switch over to the actual logic :+1:

packages/api/src/controllers/wowza-hydrate.js
```diff
@@ -38,6 +42,7 @@ export default stream => {
   const renditions = {}
   let presets = []
   const encodeNameToRenditionName = {}
+  const aspectRatio = sourceInfo.width / sourceInfo.height
```
iameli, line 45: Let's add `minValue: 1` to the schema for this just to avoid the possibility of a division-by-zero error.

packages/api/src/controllers/wowza-hydrate.js
```diff
@@ -51,25 +56,30 @@ export default stream => {
       continue
     }
 
-    const SIXTEEN_BY_NINE = 16 / 9
-    const NINE_BY_SIXTEEN = 9 / 16
-    if (width === 0) width = height * SIXTEEN_BY_NINE
-    if (height === 0) height = width * NINE_BY_SIXTEEN
+    if (width === 0) width = height * aspectRatio
+    if (height === 0) height = (width * 1) / aspectRatio
+    // TODO: do we need to validate whether or not the incoming height/width fit the aspect ratio?
```
iameli, line 61: I don't think so — if Wowza lies to us about the dimensions of the video that's on them.

iameli (approved):
Looks great! Made one tiny change so I can get it merged in advance of demo tomorrow morn: https://github.com/livepeer/livepeerjs/pull/498/commits/994bb7b4fc02798ac611afafa83d7af64c0e2450

packages/api/src/controllers/wowza-hydrate.js
angyangie: Thank you!!!

Issue #453, opened 2019-09-16
api: implement Cloudflare Worker KV as a data store


iameli:
Alongside `LevelStore` and `PostgresStore` we should have a `CloudflareStore` that implements the same interface.

angyangie:
**Notes from Discord:** 
This unblocks a lot, like intelligent API broadcaster routing, conditional per-user routing based on datacenter location, etc.

Currently, data on European streams would have to be stored in our Iowa Postgres database, so that's a big ugly round-trip hit every time. Instead, we'd like to move the database onto CloudFlare Worker KV, so we can look up the address of the pertinent broadcaster from anywhere! Pretty nice.


**Reasons for doing this:** 
- Lower latency when hitting /broadcaster on our API (or any other routes)?
- Eliminates the single point of failure that's the current Postgres server in Iowa, decreases latency for all regions that aren't Iowa, and unblocks us compiling the entire API server to a Cloudflare Worker which would dramatically increase availability and latency
- Currently, we can't do anything stateful in our existing worker (livepeerjs/packages/api/src/worker.js) because access to our database requires a connection to Postgres which you can't do in pure JavaScript on a worker. So incoming requests for manifests currently hit every single broadcaster to check to see who has the data.
- What I'd like to do instead is add logic to the auth webhook handler (POST /api/stream/hook) that remembers which broadcasters have handled a given manifestID. Then when we get a https://livepeer.live/stream/abc123.m3u8 request, we can ask for the manifests from the exact broadcasters that might have data for us instead of checking against every single broadcaster in the world. But we can't query that data from the worker unless we move from Postgres to Cloudflare Worker KV.

**Info on infrastructure:**
- The livepeerci/api Docker container runs in all of our Kubernetes clusters; that's the thing we've worked with locally. They're accessible at https://esh.livepeer.live/api, https://mcw-staging.livepeer-stagig.live/api and whatnot. Additionally, we compile a Cloudflare Worker from src/worker.js and deploy that to CloudFlare, which is accessible at https://livepeer.live and https://livepeer-staging.live.
Jumping into a different meeting, will continue explanation after
- The worker does a few things for us, mostly related to geolocation.

[...]

PR #504, opened 2019-11-11
AR/Adding CloudFlare DB to API

angyangie:
Alongside LevelStore and PostgresStore we should have a CloudflareStore that implements the same interface.

https://github.com/livepeer/livepeerjs/issues/453

To run test suite locally, do the following:
1. Create a namespace here (or use the `unit-testing` namespace). Take note of namespace ID (this link won't work ... sign into the livepeer account and navigate to the route here): https://dash.cloudflare.com/{cloudflare_account_id}/workers/kv/namespaces
2. Take note of api-token here (`LP_CLOUDFLARE_AUTH` below): https://dash.cloudflare.com/{cloudflare_account_id}/profile/api-tokens
3. Take note of the {cloudflare_account_id} in the above URLs
4. Set the following local variables using values from steps 1 - 3 above
```
export LP_CLOUDFLARE_NAMESPACE=""
export LP_CLOUDFLARE_ACCOUNT=""
export LP_CLOUDFLARE_AUTH="" 
```
- Run `yarn run postgres:start` in one terminal window
- Run `yarn run test` in another window

https://github.com/livepeer/livepeerjs/issues/453

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.test.js
+++ b/packages/api/src/controllers/stream.test.js
@@ -4,6 +4,7 @@ import uuid from 'uuid/v4'
 
 let server
 let postMockStream
+jest.setTimeout(50000)
 
 beforeAll(async () => {
   server = await serverPromise
@@ -72,7 +73,7 @@ describe('controllers/stream', () => {
 
     it('should get all streams with prior user created', async () => {
       await server.store.create(user)
-      for (let i = 0; i < 10; i += 1) {
+      for (let i = 0; i < 4; i += 1) {
         const document = {
           id: uuid(),
           kind: 'stream',
@@ -86,66 +87,26 @@ describe('controllers/stream', () => {
       const res = await client.get('/stream')
       expect(res.status).toBe(200)
       const streams = await res.json()
-      expect(streams.length).toEqual(10)
+      expect(streams.length).toEqual(4)
     })
 
-    it('should get all streams without a prior user created', async () => {
-      for (let i = 0; i < 10; i += 1) {
-        const document = {
-          id: uuid(),
-          kind: 'stream',
-        }
-        await server.store.create(document)
-        const res = await client.get(`/stream/${document.id}`)
-        const stream = await res.json()
-        expect(stream).toEqual(document)
-      }
-
-      const res = await client.get('/stream')
-      expect(res.status).toBe(200)
-      const streams = await res.json()
-      expect(streams.length).toEqual(10)
-    })
-
-    it('should get some of the streams', async () => {
+    it('should get some of the streams & get a working next Link', async () => {
       await server.store.create(user)
-      const ids = []
-      for (let i = 0; i < 10; i += 1) {
+      for (let i = 0; i < 13; i += 1) {
         const document = {
           id: uuid(),
           kind: 'stream',
         }
         await server.store.create(document)
         const res = await client.get(`/stream/${document.id}`)
-        ids.push(document.id)
         const stream = await res.json()
         expect(stream).toEqual(document)
       }
-      const res = await client.get(`/stream?cursor=${ids[2]}&limit=5`)
-      const streams = await res.json()
-      expect(streams.length).toBeLessThan(6)
-    })
-
-    it('should get a working next Link', async () => {
-      await server.store.create(user)
-      const ids = []
-      for (let i = 0; i < 30; i += 1) {
-        const document = {
-          id: uuid(),
-          kind: 'stream',
-        }
-        await server.store.create(document)
-        const res = await client.get(`/stream/${document.id}`)
-        ids.push(document.id)
-        const stream = await res.json()
-        expect(stream).toEqual(document)
-      }
-      const res = await client.get(`/stream?limit=10`)
+      const res = await client.get(`/stream?limit=11`)
       const streams = await res.json()
       expect(res.headers._headers.link).toBeDefined()
       expect(res.headers._headers.link.length).toBe(1)
-      console.log('res headers', res.headers._headers.link[0])
-      expect(streams.length).toBeLessThan(11)
+      expect(streams.length).toEqual(11)
     })
 
     it('should create a stream', async () => {
@@ -171,7 +132,7 @@ describe('controllers/stream', () => {
       }
       await server.store.create(user)
 
-      for (let i = 0; i < 10; i += 1) {
+      for (let i = 0; i < 3; i += 1) {
         const document = {
           id: uuid(),
           kind: 'stream',
--- a/packages/api/src/middleware/hardcoded-nodes.js
+++ b/packages/api/src/middleware/hardcoded-nodes.js
@@ -8,7 +8,7 @@ export default function hardcodedNodes({ broadcasters, orchestrators }) {
     broadcasters = JSON.parse(broadcasters)
     orchestrators = JSON.parse(orchestrators)
   } catch (e) {
-    console.error('Error parsing LP_BROADCASTERS and LP_ORCHESTRATORS', e)
+    console.error('Error parsing LP_BROADCASTERS and LP_ORCHESTRATORS')
     throw e
   }
   return (req, res, next) => {
--- a/packages/api/src/store/cloudflare-store.js
+++ b/packages/api/src/store/cloudflare-store.js
@@ -1,0 +1,131 @@
+import logger from '../logger'
+import { NotFoundError } from './errors'
+import fetch from 'isomorphic-fetch'
+// import { timeout } from '../util'
+import { parse as parseUrl, format as stringifyUrl } from 'url'
+
+// const CONNECT_TIMEOUT = 5000
+const DEFAULT_LIMIT = 10
+const cloudflareUrl = 'https://api.cloudflare.com/client/v4/accounts'
+let namespace
+let accountId
+let auth
+
+export default class CloudflareStore {
+  constructor({ cloudflareNamespace, cloudflareAccount, cloudflareAuth }) {
+    if (!cloudflareNamespace || !cloudflareAccount || !cloudflareAuth) {
+      throw new Error(
+        'no cloudflare namespace, account id, or authorization key provided',
+      )
+    }
+    namespace = cloudflareNamespace
+    accountId = cloudflareAccount
+    auth = cloudflareAuth
+  }
+
+  async list(prefix = '', cursor = null, limit = DEFAULT_LIMIT) {
+    const requestUrl = `${cloudflareUrl}/${accountId}/storage/kv/namespaces/${namespace}/keys?limit=${limit}&prefix=${prefix}`
+    const cursorParam = `&cursor=${cursor}`
+    if (cursor) {
+      requestUrl.concat(cursorParam)
+    }
+
+    const res = await fetch(requestUrl, {
+      method: 'GET',
+      headers: {
+        authorization: `${auth}`,
+      },
+    })
+
+    const data = await res.json()
+    checkStatus(res, data)
+    return { data: data.result, cursor: data.result_info.cursor }
+  }
+
+  async get(value) {
+    const requestUrl = `${cloudflareUrl}/${accountId}/storage/kv/namespaces/${namespace}/values/${value}`
+    const res = await fetch(requestUrl, {
+      method: 'GET',
+      headers: {
+        Authorization: `${auth}`,
+      },
+    })
+
+    const data = await res.json()
+    checkStatus(res, data)
+    return data
+  }
+
+  async create(data) {
+    const { id, kind } = data
+
+    if (!id || !kind) {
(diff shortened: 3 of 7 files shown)


packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,131 @@
+import logger from '../logger'
+import { NotFoundError } from './errors'
+import fetch from 'isomorphic-fetch'
+// import { timeout } from '../util'
+import { parse as parseUrl, format as stringifyUrl } from 'url'
+
+// const CONNECT_TIMEOUT = 5000
```
angyangie: Not sure if we need to implement any kind of timeout here ... so left this as a reminder to ask

packages/api/src/controllers/stream.test.js
```diff
@@ -3,6 +3,7 @@ import { TestClient, clearDatabase } from '../test-helpers'
 import uuid from 'uuid/v4'
 
 let server
+jest.setTimeout(50000)
```
angyangie, line 7: We can remove this once the CloudFlare rate limit is increased. For now, tests will take a long time to run!

packages/api/src/controllers/stream.test.js
```diff
@@ -47,48 +48,26 @@ describe('controllers/stream', () => {
...
-        ids.push(document.id)
         const stream = await res.json()
         expect(stream).toEqual(document)
       }
-      const res = await client.get(`/stream?cursor=${ids[2]}&limit=5`)
-      const streams = await res.json()
-      expect(streams.length).toBeLessThan(6)
-    })
-
-    it('should get a working next Link', async () => {
```
angyangie, line 129: Combined this test with the last one ... esp with CloudFlare rate limits, didn't see a point to having two of them testing stream object lengths. Also, a limit of 5 doesn't work with cloudflare ... we need a limit of 10 or more.

packages/api/src/controllers/stream.test.js
```diff
@@ -101,7 +80,7 @@ describe('controllers/stream', () => {
       }
       await server.store.create(user)
 
-      for (let i = 0; i < 10; i += 1) {
+      for (let i = 0; i < 3; i += 1) {
```
angyangie, line 135: Only need to test 403s a couple of times. 10 seemed like a lot (again, keeping in mind these rate limits)

packages/api/src/middleware/hardcoded-nodes.js
```diff
@@ -8,7 +8,7 @@ export default function hardcodedNodes({ broadcasters, orchestrators }) {
     broadcasters = JSON.parse(broadcasters)
     orchestrators = JSON.parse(orchestrators)
   } catch (e) {
-    console.error('Error parsing LP_BROADCASTERS and LP_ORCHESTRATORS', e)
+    console.error('Error parsing LP_BROADCASTERS and LP_ORCHESTRATORS')
```
angyangie, line 11: Annoying log ... took it out

packages/api/src/store/cloudflare-store.js
```diff
@@ -86,46 +75,55 @@ export default class CloudflareStore {
...
+  }, error_messages: ${JSON.stringify(respData.errors)}`
+
+  if (res.status != 200) {
+    console.log(errorMessage)
+
+    if (res.status == 404) {
+      throw new NotFoundError()
+    } else if (res.status == 429) {
+      console.log('Sleeping for 3 seconds')
+      await sleep(3000)
```
angyangie, line 120: We can remove this when rate limit is increased, along with other sleeps

angyangie:
Still need to figure out how to give circleci default values for the `cloudflare namespace`, `account id`, and `authorization key` so the tests actually pass 🤔 🤔

packages/api/src/controllers/stream.test.js
iameli: I think this is fine for the tests... we want it to work with other folks' Cloudflare accounts out-of-the-box too. Still a good idea for me to open that ticket though, I'll do that now

packages/api/src/controllers/stream.test.js
iameli: Cool, looks fine to me

packages/api/src/controllers/stream.test.js
iameli: Sure.

packages/api/src/middleware/hardcoded-nodes.js
iameli: :+1:

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
+import logger from '../logger'
+import { NotFoundError } from './errors'
+import fetch from 'isomorphic-fetch'
+import { parse as parseUrl, format as stringifyUrl } from 'url'
+
+const querystring = require('querystring')
```
iameli: Can also be spelled `import querystring from 'querystring'` for consistency

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
...
+const cloudflareUrl = 'https://api.cloudflare.com/client/v4/accounts'
```
iameli: Kinda prefer `CAPITAL_CASE` for constants like this... but idc very much tbh

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
...
+      prefix: prefix,
+      cursor: cursor,
+    })
+
+    const reqUrl = `${cloudflareUrl}/${accountId}/storage/kv/namespaces/${namespace}/keys?${params}`
+    const respData = await cloudFlareFetch(reqUrl, null, 'GET', 0)
+
+    const values = []
+    var i
+    for (i = 0; i < respData.result.length; i++) {
```
iameli: `let` is preferable to `var` here and can be defined inline: `for (let i = 0; i < respData.result.length; i++) {`

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
...
+
+  async replace(data) {
+    const { id, kind } = data
+
+    if (!id || !kind) {
+      throw new Error("object missing 'id' and/or 'kind'")
+    }
+    const key = `${kind}/${id}`
+    const reqUrl = `${cloudflareUrl}/${accountId}/storage/kv/namespaces/${namespace}/values/${key}`
+    const respData = await cloudFlareFetch(reqUrl, data, 'PUT', 0) // can we not declare?
```
iameli: Sure! `await cloudFlareFetch(reqUrl, data, 'PUT', 0)`

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async delete(id) {
+    const reqUrl = `${cloudflareUrl}/${accountId}/storage/kv/namespaces/${namespace}/values/${id}`
+    const respData = await cloudFlareFetch(reqUrl, null, 'DELETE', 0)
+  }
+}
+
+async function cloudFlareFetch(reqUrl, data, method, retries) {
```
iameli: I like this function and its parameters but I think the `null`s for body make it a little awkward to call: 

```
await cloudFlareFetch(reqUrl, null, 'DELETE', 0)
```

Perhaps we could give it the same `fetch(url, opts)` signature that `fetch` itself uses? So that could look like uhhh

```javascript
async function cloudFlareFetch(reqUrl, {method = 'GET', retries = 0, data = null} = {}) {
```

And then it could be called like

```javascript
const respData = await cloudFlareFetch(reqUrl, {
  data: data,
  method: 'PUT',
  retries: 0
)
```

Or a simple GET could be:
```javascript
await cloudFlareFetch(reqUrl)
```

Oh! And Cloudflare doesn't capitalize the F, so probably `cloudflareFetch`.

packages/api/src/store/cloudflare-store.js
```diff
@@ -0,0 +1,129 @@
...
+    req.body = JSON.stringify(data)
+  }
+
+  const res = await fetch(reqUrl, req)
+  const respData = await res.json()
+  const errorMessage = `Cloudflare ${res.status} error: ${
+    res.statusText
+  }, error_messages: ${JSON.stringify(respData.errors)}`
+
+  if (res.status != 200) {
```
iameli, line 110: Confirming - it's a 200 even for `POST`s and `DELETE`s? That's unusual.

packages/api/src/store/cloudflare-store.js
iameli: Hmmm yeah, trying to figure out what to do here - these `sleep`s and retries aren't going to help much if we're under any kind of load and making thousands of requests — the requests will just loop and keep exceeding the limit. So that'd call for a rate-limiting system with a queue and rules about the number of concurrent requests. On the other hand, all of that would be just to use this thing outside of the Worker, and we really aren't designing this to work outside of the Worker. So I kinda wanna say let's ship this if the tests are consistent and worry about performance after we're actually deploying to the worker.

iameli (changes requested):
Cool! Just a few pieces of polish.

packages/api/src/controllers/stream.test.js
angyangie: Thank you for filing that ticket. Cool, we'll just leave the timeout regardless

packages/api/src/store/cloudflare-store.js
angyangie: hahaha love it. I don't either, so I'll just change to all capitals :)

packages/api/src/store/cloudflare-store.js
angyangie: Thank you! Oh javascript

packages/api/src/store/cloudflare-store.js
angyangie: 🙇‍♀🙇‍♀🙇‍♀

packages/api/src/store/cloudflare-store.js
angyangie: Good point, I'll confirm and adjust if needed

packages/api/src/store/cloudflare-store.js
angyangie: Agreed. If I had to streamline tests and triple the default jest timeout for tests to pass, no way this would work work under any kind of load. Why would this not be a problem inside the worker? Because it'll handle the load appropriately?  I'm fine shipping it since tests are consistent

packages/api/src/store/cloudflare-store.js
angyangie: Love the edits ... thanks for making my code cleaner. I'll change

packages/api/src/controllers/stream.test.js
j0sh: Normally for these types of tests, mocking is used to simulate the behavior of external resources. That way our logic can be tested (and particular failures etc injected to test specific error paths) without introducing a run-time dependency on resources that are slow/expensive/rate-limited/hard-to-access/etc.

packages/api/src/store/cloudflare-store.js
angyangie: 200s from `POST`s and `DELETE`s confirmed. Unusual indeed!

packages/api/src/controllers/stream.test.js
iameli: I hear that, and there might be some value in it eventually, but I've liked so far that we've run pretty much everything as integration tests — it's caught bugs in inconsistent implementations between the two existing stores that I'm not sure we'd have seen if we were mocking everything.

j0sh: Sure, the stuff that directly talks to CF should still be tested separately to ensure that whatever assumptions we are making still hold.

I haven't actually reviewed any of these tests, but are we actually testing different CF behaviors, or the same CF behavior repeatedly but with a slightly different code path on our end? The latter is what would justify mocking, *in addition to* separate tests for specific behaviors we expect from CF.

iameli (approved):
LGTM!

packages/api/src/controllers/stream.test.js
angyangie: Hmmm I'm not sure what you mean @j0sh. We're testing all stores (local, postgres, CloudFlare) in `stream.test.js`. We're running that file three times using all three DB options to ensure consistent results regardless of DB chosen ( see here: https://github.com/livepeer/livepeerjs/pull/504/files#diff-43c6f412a6b721476ad29c6674d36fa6R42-R47). 

Since I'm working on another Cloudflare ticket, we can do additional testing there if necessary: https://app.zenhub.com/workspaces/video-user-team-5dd2c75e4ee08500015994d7/issues/livepeer/livepeerjs/508

PR #515, opened 2019-12-04
Adding the ability to handle passthrough custom profiles

angyangie:
Adding the ability for API to handle passthrough custom profiles.

- Removing presets from Wowza Hydrate
- Adding profile construction from Wowza config 

- End to end test with go-livepeer changes (here: https://github.com/livepeer/go-livepeer/pull/1242) + streaming
- Unit tests

https://github.com/livepeer/go-livepeer/issues/1185 and https://github.com/livepeer/go-livepeer/issues/811

**Checklist:**


packages/api/src/controllers/wowza-hydrate.test.js
```diff
@@ -20,112 +20,248 @@ streamTransrate45.wowza.sourceInfo.fps = 45
...
+        name: 'width_height_test_360p',
+        height: 360,
+        width: 640,
+        bitrate: 850000,
+        fps: 30,
+      },
+      {
+        name: 'width_height_test_240p',
+        height: 240,
+        width: 426.66666666666663,
```
angyangie: This sort of thing is bound to happen when width or height is zero with custom profiles ... do we care?

iameli: We should probably round it to an integer, no such thing as a two-thirds pixel.

packages/api/src/controllers/wowza-hydrate.js
```diff
@@ -58,40 +56,21 @@ export default stream => {
...
-      if (thisDiff < diff || (thisDiff == diff && thisFpsDiff < fpsDiff)) {
-        foundPreset = preset.name
-        diff = thisDiff
-        fpsDiff = thisFpsDiff
-      }
+    const profile = {
+      name: renditionName,
+      height: height,
+      width: width,
+      bitrate: parseInt(videoBitrate),
```
iameli: Pretty sure the Wowza bitrates are in bytes. Ours are in kilobytes, so this should get a `Math.round(parseInt(videoBitrate)) / 1000`.

packages/api/src/controllers/wowza-hydrate.test.js
```diff
@@ -20,112 +20,248 @@ streamTransrate45.wowza.sourceInfo.fps = 45
...
-      'P360p30fps16x9',
-      'P240p30fps16x9',
-      'P144p30fps16x9',
+
+    expect(newStream.profiles).toEqual([
+      {
+        name: 'width_height_test_720p',
+        height: 720,
+        width: 1280,
+        bitrate: 1300000,
```
iameli, line 178: Yeah, all of these should be adjusted — this would be about 1.2 gigabits per second haha

iameli (changes requested):
Looks good, few changes

packages/api/src/controllers/wowza-hydrate.js
j0sh: > Ours are in kilobytes

Kilobits, so if Wowza is actually in bytes (rather than bits) then it would be `parseInt(videoBitrate)*8/1000`  (or equivalently, `parseInt(videoBitrate)/125`).

packages/api/src/controllers/wowza-hydrate.test.js
angyangie: lol 🤦‍♀

packages/api/src/controllers/wowza-hydrate.test.js
angyangie: Sounds good to me!

packages/api/src/controllers/wowza-hydrate.js
angyangie: Thank you both! Good catch Eli

iameli: Ah, I mispoke — both of mine should have been bits, no `8` required here.

Actually I think it should be `/1024`, yeah?

iameli (approved):

Issue #516, opened 2019-12-10
API: Store Endpoint


wohlner:
- Need the ability for an API user to specify the credentials and store into, then when they make a stream to associate it with the store such that it could be returned by the Webhook.
- Endpoint: `/api/objectStores`
- To do: Schema, schema validation, associated with a `userId`
- Use MiddleWare auth user objects. API keys. 

- Example: CBS wants to add in Media Store credentials, and have streams they create upload to that media store. 
- Goal: create stream associated with an object store. New field on `stream` that has that info. Talk to Ivan who is working on webhook protocols. Webhook call with appropriate ManifestId returning the info from `/hook` formatted appropriately for `go-livepeer` consumption.

PR #518, opened 2019-12-16
Adding api store endpoint - ready for review

angyangie:

Adds the ability for an API user to specify the credentials and store into, then when they make a stream to associate it with the store such that it could be returned by the Webhook.

To create an objectStore:

1. POST `/api/object-store` with `{"type": "s3", "path": "us-west-2/my-bucket", "credentials": "abc123/abc123" }` (`path` is `region/bucket` and `credentials` are `ACCESSKEYID/ACCESSKEY`)
2. Get back an object with `{"id": "395a4d55-8f6c-4b18-b6cc-084c29c6e2ad" "type": "s3", "path": "us-west-2/my-bucket", "credentials": "actually let's not send these back to the user"}`

Then, for every stream being streamed to this OS:

1. POST `/api/stream` with`{objectStoreId: "395a4d55-8f6c-4b18-b6cc-084c29c6e2ad", profiles: [... etc ...]}`
2. Get back `{id: "5e58543b-1468-4e50-81a6-c1255ed9b6f0", objectStoreId: "395a4d55-8f6c-4b18-b6cc-084c29c6e2ad", profiles: [...etc...]}`

Then, use `5e58543b-1468-4e50-81a6-c1255ed9b6f0` as an RTMP stream key/ HTTP ingest key.
Stream into `rtmp://some-broadcaster.example.com/live/5e58543b-1468-4e50-81a6-c1255ed9b6f0 `and `go-livepeer` hits `/api/stream/hook`. We send back something like: 
```javascript
{
  manifestId: "5e58543b-1468-4e50-81a6-c1255ed9b6f0",
  objectStore: {
    "type": "s3",
    "path": "us-west-2/my-bucket",
    "credentials": "abc123/abc123"
  },
  profiles: [ ...etc... ]
}
```

- Adding credentials schema & endpoint

Will add tests 

https://github.com/livepeer/livepeerjs/issues/516

**Checklist:**


packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+    next.searchParams.set('cursor', nextCursor)
+    res.links({
+      next: next.href,
+    })
+  }
+  res.json(output)
+})
+
+app.get('/:id', authMiddleware({}), async (req, res) => {
+  const credentials = await req.store.get(`objectstores/${req.params.id}`)
```
iameli: Let's do a bit more work here and validate that `req.user.id === credentials.userId` so users can only download their own objectStores.

iameli: And I'd say maybe this could be called `const os` so we can tell the difference between the `credentials` variable and the `credentials.credentials` sub-field

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+import path from 'path'
+
+const app = Router()
+
+app.get('/', authMiddleware({ admin: true }), async (req, res) => {
+  let limit = req.query.limit
+  let cursor = req.query.cursor
+  logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
+
+  const resp = await req.store.list(`objectstores/`, cursor, limit)
```
iameli: I think for this particular object we're going to want to store OSes at `/objectstores/${userId}/${objectStoreId}`. I can't think of a good scenario to want to list them globally, but we do want users to easily be able to list their _own_ stores, which we can do with a list on `/objectstores/${userId}/`.

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+      id: id,
+      credentials: req.body.credentials,
+      path: req.body.path,
+      userId: req.user.id,
+      type: req.body.type,
+      kind: 'objectstores',
+    })
+
+    const store = await req.store.get(`objectstores/${id}`)
+    store.credentials = 'xxx'
```
iameli: Solid, but I like `null` here, "this field is here but you cannot see it"

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+      next: next.href,
+    })
+  }
+  res.json(output)
+})
+
+app.get('/:id', authMiddleware({}), async (req, res) => {
+  const credentials = await req.store.get(`objectstores/${req.params.id}`)
+  res.status(200)
+  res.json(credentials)
```
iameli: On this endpoint - the one used by individuals to retrieve one set of their credentials - we shouldn't send back the actual credentials. We want them to stay secure even if your account is compromised. Easy enough to accomplish with something like
```javascript
const secureOS = {...os, credentials: null}
res.json(secureOS).
```

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+  let limit = req.query.limit
+  let cursor = req.query.cursor
+  logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
+
+  const resp = await req.store.list(`objectstores/`, cursor, limit)
+  const output = resp.data
+  const nextCursor = resp.cursor
+  res.status(200)
+
+  let baseUrl = new URL(
```
iameli: This process will get repetitive fast - let's get this into a shared helper function that takes `(req, nextCursor)` and returns `next.href`

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,66 @@
...
+    `${req.protocol}://${req.get('host')}${req.originalUrl}`,
+  )
+  if (output.length > 0) {
+    let next = baseUrl
+    next.searchParams.set('cursor', nextCursor)
+    res.links({
+      next: next.href,
+    })
+  }
+  res.json(output)
```
iameli, line 32: We should do an `output.filter()` option here to clear out the `credentials` field of each of these

packages/api/src/middleware/auth.js
```diff
@@ -100,6 +100,16 @@ function authFactory(params) {
       return res.sendStatus(401)
     }
 
+    if (req.body.objectStoreId) {
+      try {
+        const store = await req.store.get(
+          `objectstores/${req.body.objectStoreId}`,
+        )
+      } catch (e) {
+        throw e
```
iameli: If we don't need to do anything else with this error, we can lose the `try{}` block entirely

packages/api/src/middleware/auth.js
```diff
@@ -100,6 +100,16 @@ function authFactory(params) {
       return res.sendStatus(401)
     }
 
+    if (req.body.objectStoreId) {
+      try {
+        const store = await req.store.get(
```
iameli: This is a good check - we want to make sure that the store exists if you're using it in a stream - but we should do this check in the `POST /stream` endpoint rather than here where it'll run for any potential body containing an `objectStoreId` key

packages/api/src/schema/schema.yaml
```diff
@@ -15,6 +15,8 @@ components:
         - id
         - kind
         - name
+        - objectStoreId
```
iameli: It shouldn't be required, lots of streams work just fine without an OS.

packages/api/src/schema/schema.yaml
```diff
@@ -15,6 +15,8 @@ components:
...
+        - presets
```
iameli: I don't think presets should be required either - we're transitioning to `profiles` for most cases anyway.

packages/api/src/schema/schema.yaml
```diff
@@ -115,6 +119,21 @@ components:
...
+        - type
+        - path
+        - credentials
+      additionalProperties: false
+      properties:
+        type:
+          type: string
+        path:
+          type: string
+        credentials:
```
iameli, line 145: Thinking this should be an object with `type`, `path`, and `credentials` fields like this here https://github.com/livepeer/go-livepeer/issues/1263

packages/api/src/schema/schema.yaml
```diff
@@ -162,3 +181,50 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/error'
+
+  /objectstores:
+    tags:
+      - stream
```
iameli: What result does this have on the docs? Thinking maybe it should have an `objectstores` tag or something

iameli (changes requested):
Solid — two meaningful changes (changing the key where we store OSes and a string --> object proposed change for `credentials`), everything else is just polish.

packages/api/src/controllers/object-stores.js
angyangie: Really good point! However, in conjunction with one of your other (great) comments, if we store OS's in `/objectstores/${userId}/${objectStoreId}` ... do we still want to check for `os.userId`? If we store the object this way, the userId in the route will always be the same as the `userId` stored in the `objectStores` object itself. Which leads me to the question: do we then need to store the `userId` in the object itself? 

```javascript
await req.store.create({
      id: id,
      credentials: req.body.credentials,
      path: req.body.path,
      userId: req.user.id,
      type: req.body.type,
      kind: `objectstores/${req.user.id}`,
    })
```

packages/api/src/controllers/object-stores.js
angyangie: You're right. I can't think of a scenario in which we'd want to list them globally either.

packages/api/src/middleware/auth.js
angyangie: Good point ... we don't need to do this check for every endpoint

packages/api/src/schema/schema.yaml
angyangie: What do you mean? I included it in the `objectstores` object here

packages/api/src/schema/schema.yaml
angyangie: haha woops

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const resp = await req.store.list(`objectstores/`, cursor, limit)
+  let output = resp.data
+  const nextCursor = resp.cursor
+  res.status(200)
+
+  if (output.length > 0) {
+    res.links({ next: makeNextHREF(req, nextCursor) })
+  }
+  output = output.filter(x => x.userId == req.user.id)
```
iameli: `===` always. We should get a linter rule for that.

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,77 @@
...
+  output.map(x => (x.credentials = null))
```
iameli: This'll modify the existing records, kinda bad form for a `.map()` call. I like 
```javascript
output = output.map(x => ({
  ...x,
  credentials: null 
}))
```

packages/api/src/schema/schema.yaml
iameli: Oh, so you did. Whoops! JSON-Schema brain fart.

In that case, we should add the rest of the fields onto this — these three are all that are required for a POST, but there's also `id` and `userId` on the object itself.

packages/api/src/controllers/stream.js
```diff
@@ -43,11 +43,22 @@ app.get('/:id', authMiddleware({}), async (req, res) => {
 app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
   const id = uuid()
 
+  let objectStoreID
+  if (req.body.objectStoreId) {
+    try {
+      await req.store.get(`objectstores/${req.body.objectStoreId}`)
+      objectStoreID = req.body.objectStoreId
+    } catch (e) {
+      console.error(e)
```
iameli, line 56: Erg, this is another place where we have to do the annoying pattern where we check for "NotFoundError" and continue if it's of that type but throw otherwise. Let's open a ticket to change that to returning null from `req.store.get` next week?

iameli (changes requested):
Cool, just minor changes.

packages/api/src/controllers/stream.js
angyangie: Yes, that would be helpful. Just created the ticket: https://github.com/livepeer/livepeerjs/issues/525

packages/api/src/controllers/object-stores.test.js
```diff
@@ -0,0 +1,290 @@
...
+
+      const resp = await client.get(
+        `/objectstores/${tokenObject.userId}/${objStore.id}`,
+      )
+      expect(resp.status).toBe(200)
+      const objStoreGet = await resp.json()
+      expect(objStore.path).toEqual(objStoreGet.path)
+      expect(objStore.userId).toBe(objStoreGet.userId)
+
+      // testing right number of object stores returned when multiple created by different users
```
iameli, line 209: :+1:

packages/api/src/controllers/stream.js
```diff
@@ -109,11 +123,17 @@ app.post('/hook', async (req, res) => {
   }
 
   const stream = await req.store.get(`stream/${streamId}`)
+  const store = await req.store.get(`objectstores/${stream.objectStoreId}`)
```
iameli: This'll break for streams that don't have an object store. Which is, at the moment, all of them.

iameli (changes requested):
One line change 👍

packages/api/src/controllers/object-stores.js
```diff
@@ -0,0 +1,80 @@
...
+  async (req, res) => {
+    const id = uuid()
+
+    await req.store.create({
+      id: id,
+      credentials: req.body.credentials,
+      path: req.body.path,
+      userId: req.user.id,
+      type: req.body.type,
+      kind: `objectstores/${req.user.id}`,
```
iameli, line 61: So I tried to be clever with this `kind` field and it ended up being kind of silly. What do you think would be a better way to handle multiple indexes like this? One option would be to ditch the `kind` field entirely and have this thing's `id` be `objectstores/${userId}/${id}`. But I dunno.

angyangie: We could do that. Is having an ID of `objectstores/${userId}/${id}` v. having a kind and ID more intuitive? Are we thinking of doing this only for objectstores? Or, is this important as a standard for future routes for items that might need to be nested under a userId?

angyangie: I'll push up my changes without this one included for now, but will circle back to this once we come to consensus!

packages/api/src/controllers/stream.js
```diff
@@ -123,17 +127,25 @@ app.post('/hook', async (req, res) => {
   }
 
   const stream = await req.store.get(`stream/${streamId}`)
-  const store = await req.store.get(`objectstores/${stream.objectStoreId}`)
+  let objectStore = { type: '', path: '' }
```
iameli:
```suggestion
  let objectStore = undefined
```

iameli (approved):

Issue #525, opened 2020-01-20
Return `null` from `req.store.get` instead of  `NotFoundError`


angyangie:
**What's the problem & current behavior?**

We have a pattern of returning `NotFoundError` when an item isn't found in the database. As a result, the following is repeated often:
```
if (e.type !== 'NotFoundError') {
        throw e
}
```

**Describe a possible solution you've considered (optional)**

We should return `null` from `req.store.get` instead of  `NotFoundError`.

PR #540, opened 2020-01-31
Removing NotFoundErrors from API

angyangie:

We have a pattern of returning NotFoundError when an item isn't found in the database. As a result, the following is repeated often:

```javascript
if (e.type !== 'NotFoundError') {
        throw e
}
```

We should return null from req.store.get instead of `NotFoundError`.

https://github.com/livepeer/livepeerjs/issues/525

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -53,10 +53,7 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
       )
       objectStoreID = req.body.objectStoreId
     } catch (e) {
-      console.error(e)
-      if (e.type !== 'NotFoundError') {
-        throw e
-      }
+      throw e
     }
   }
 
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -43,18 +43,10 @@ app.get('/:id', authMiddleware({}), async (req, res) => {
 app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
   const id = uuid()
 
-  console.log(`idddddd: ${req.body.objectStoreId}`)
-
   let objectStoreID
   if (req.body.objectStoreId) {
-    try {
-      await req.store.get(
-        `objectstores/${req.user.id}/${req.body.objectStoreId}`,
-      )
-      objectStoreID = req.body.objectStoreId
-    } catch (e) {
-      throw e
-    }
+    await req.store.get(`objectstores/${req.user.id}/${req.body.objectStoreId}`)
+    objectStoreID = req.body.objectStoreId
   }
 
   const doc = wowzaHydrate({
--- a/packages/api/src/middleware/auth.js
+++ b/packages/api/src/middleware/auth.js
@@ -37,7 +37,6 @@ async function generateUserAndToken(req, res, next) {
     const user = await req.store.get(`user/${userId}`)
     req.user = user
   } catch (error) {
-    console.error(error)
     throw error
   }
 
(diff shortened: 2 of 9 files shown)


iameli:
image

This indicates to me this was a good idea 😄

packages/api/src/controllers/stream.js
```diff
@@ -53,10 +53,7 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
       )
       objectStoreID = req.body.objectStoreId
     } catch (e) {
-      console.error(e)
-      if (e.type !== 'NotFoundError') {
-        throw e
-      }
+      throw e
```
iameli: This and the other cases like it can have the `try/catch` removed entirely. If the only thing we're doing upon an error is throwing it up the stack, there's no difference there

packages/api/src/middleware/auth.js
```diff
@@ -106,13 +106,14 @@ function authFactory(params) {
       // if tokenObject does not exist, create tokenObject and user
       tokenObject = await req.store.get(`apitoken/${req.token}`)
     } catch (e) {
-      if (e.type !== 'NotFoundError') {
-        throw e
-      }
+      throw e
```
iameli: same here

packages/api/src/middleware/auth.js
```diff
@@ -170,20 +171,20 @@ async function getUserWithGoogleAuth(req, res, next) {
   try {
     user = await req.store.get(`user/${payload.sub}`)
   } catch (error) {
-    if (error.type !== 'NotFoundError') {
-      console.error(error)
-      throw error
-    }
+    console.error(error)
```
iameli: Same - feel free to remove these `console.error()` statements generally too. There should be unhandled error logging that takes care of that above us.

packages/api/src/store/level-store.js
```diff
@@ -74,7 +74,15 @@ export default class LevelStore {
...
   async get(key) {
     await this.ready
-    return JSON.parse(await this.db.get(key))
+    let res
+    try {
+      res = await this.db.get(key)
+    } catch (err) {
+      if (err.name === 'NotFoundError') {
+        return null
+      }
```
iameli, line 84: We should `throw err` here if it was an error other than a NotFoundError

packages/api/src/store/level-store.js
```diff
@@ -86,15 +94,14 @@ export default class LevelStore {
...
-    } catch (err) {
-      if (!err.type === 'NotFoundError') {
-        throw err
+      item = await this.get(`${kind}/${id}`)
+      if (item) {
+        throw new Error(`${id} already exists`)
       }
-      // Not found - that's great!
+    } catch (err) {
+      throw err
```
iameli: block can be removed

packages/api/src/store/postgres-store.js
```diff
@@ -94,19 +93,13 @@ export default class PostgresStore {
       `UPDATE ${TABLE_NAME} SET data = $1 WHERE id = $2`,
       [JSON.stringify(data), key],
     )
-    if (res.rowCount < 1) {
-      throw new NotFoundError()
```
iameli, line 98: Interesting... I think we might still want these two. Calling `update()` on something that doesn't exist is still an error condition

iameli (changes requested):
Cool, mostly just cleanup of instances where we can remove try{} blocks entirely.

I'm thinking there might need to be some extra logic in Cloudflare and LevelDB's `update` and `delete` methods - I think in those cases we do want to throw an error if the record doesn't exist. Probably most of the time if I try and `update()` with an id that doesn't exist something has gone horribly wrong.

packages/api/src/controllers/stream.js
angyangie: I didn't realize that ... thank you!

angyangie:
Thanks for the comments Eli! Working on extra logic for Cloudflare and LevelDB's update and delete methods now

packages/api/src/store/postgres-store.js
```diff
@@ -104,9 +105,6 @@ export default class PostgresStore {
       `DELETE FROM ${TABLE_NAME} WHERE id = $1`,
       [id],
     )
-    if (res.rowCount < 1) {
```
iameli, line 107: Think this one should still be here.

iameli (approved):
One tiny change, feel free to merge after that and CI passes!

packages/api/src/store/postgres-store.js
angyangie: Thanks Eli!

PR #544, opened 2020-02-03
fix ci build errors

adamsoffer:
This PR fixes an error in the CI build caused by the `subgraph`and `explorer-2.0` packages

- adds a prebuild step to the `subgraph` package that runs the script responsible for creating the `subgraph.yaml` file the CI is complaining about
- adds a yarn resolution for the 3box `multicodec` dependency. This fixes another ci build error caused by this open 3box issue https://github.com/3box/3box-js/issues/687 
- Updates yarn.lock file

**Additional Context**
@iameli looks like there's also a build error caused by the `api-frontend` package but perhaps your already aware of that one.
`@livepeer/api-frontend: Error occurred prerendering page "/login": TypeError: Cannot read property 'asPath' of null`
 
**Checklist:**

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -7,7 +7,7 @@
     "create:local": "graph create livepeer/livepeer --node http://127.0.0.1:8020",
     "codegen": "graph codegen --debug --output-dir src/types/",
     "build": "graph build --debug",
-    "prebuild": "yarn prepare:mainnet && yarn codegen",
+    "prepare": "yarn prepare:mainnet && yarn codegen",
     "deploy": "graph deploy $SUBGRAPH --debug --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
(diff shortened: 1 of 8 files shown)


packages/subgraph/package.json
```diff
@@ -7,6 +7,7 @@
     "create:local": "graph create livepeer/livepeer --node http://127.0.0.1:8020",
     "codegen": "graph codegen --debug --output-dir src/types/",
     "build": "graph build --debug",
+    "prebuild": "yarn prepare:mainnet && yarn codegen",
```
iameli: Is `prebuild` the right step for this? If it's something that should be set up every time `yarn install` runes, it feels more like a `prepare` to me.

iameli (commented):
image
Looks like we're currently getting this error - can happen when CircleCI goes OOM.

packages/subgraph/package.json
adamsoffer: @iameli ah yes that makes sense. I'll change to `prepare`.

Issue #533, opened 2020-01-27
explorer: display total fees earned by an O


adamsoffer:
**What's the problem? (required)**
As a tokenholder, I don't have any visibility into how many total fees an orchestrator has earned.

**Describe a possible solution you've considered (optional)**
Capture redeemed winning ticket events on the livepeer subgraph and accrue fees for individual O's that can then be displayed in the explorer. 

**Additional context (optional)**
The total fees should be displayed on the orchestrator list view as well as the campaign view for an orchestrator.
<img width="506" alt="Screen Shot 2020-01-27 at 10 36 15 AM" src="https://user-images.githubusercontent.com/555740/73188447-e239f800-40f0-11ea-8a7a-1886703696b5.png">

PR #551, opened 2020-02-05
explorer: display total fees per O 

adamsoffer:
This PR displays an orchestrator's earned fees in the main list view as well as the campaign view.

Staging url -> https://beta.explorer.livepeer.org/

- Add a `winningTicketRedeemed` mapping to the subgraph for indexing fees
- Adds a fees column to the orchestrator list view
- Adds a fees card component to the orchestrator campaign view
- Some miscellaneous CSS improvements and refactoring

#533 

**Screenshots (optional):**
Screen Shot 2020-02-05 at 9 57 19 AM
Screen Shot 2020-02-05 at 9 56 08 AM

**Checklist:**


packages/explorer-2.0/components/Orchestrators/index.tsx
```diff
@@ -369,6 +363,30 @@ function renderTooltip(title) {
...
+        <>
+          <ReactTooltip
+            id="tooltip-fees"
+            className="tooltip"
+            place="bottom"
+            type="dark"
+            effect="solid"
+          />
+          <Help
+            data-tip="Total fees earned from transcoding."
```
yondonfu: Should we specify ETH as the token here since we specify LPT as the token in the "Stake" tooltip?

packages/explorer-2.0/data/unlock-tokens.mdx
```diff
@@ -0,0 +1,3 @@
+In order to stake your LPT tokens you must first grant permissions to the Livepeer
```
yondonfu: "LPT tokens" -> "LPT"?

yondonfu: Do you think it is important to mention "Proxy" here? Not sure how relevant that will to be for most users.

packages/explorer-2.0/data/unlock-tokens.mdx
```diff
@@ -0,0 +1,3 @@
...
+Bonding Manager Proxy contract. This will allow the contract to transfer the LPT
```
yondonfu: This currently reads like the contract is one that is going to start earnings rewards. WDYT about "and start earning rewards" -> "allowing you to start earning rewards"?

packages/explorer-2.0/layouts/main.tsx
```diff
@@ -65,7 +64,7 @@ export default ({
     items.push({
       name: (
         <Box>
-          {process.env.THREEBOX_ENABLED && threeBoxSpace && threeBoxSpace.name
+          {process.env.THREEBOX_ENABLED && threeBoxSpace?.name
```
yondonfu: Oh cool so `threeBoxSpace?.name` is shorthand for `threeBoxSpace ? threeBoxSpace.name : undefined`?

packages/subgraph/networks.yaml
```diff
@@ -21,6 +21,9 @@ mainnet:
     minter:
       address: "8573f2f5a3bd960eee3d998473e50c75cdbe6828"
       startBlock: 5533890
+    ticketBroker:
+      address: "5b1ce829384eebfa30286f12d1e7a695ca45f5d2"
+      startBlock: 5533890
```
yondonfu: The TicketBroker proxy was deployed at block [9274420](https://etherscan.io/tx/0x1c22c8ac44cc7365de3c4c13f9b08c1bbce61b283ca9686692580b39ec667ce3), but I guess as long as the subgraph can handle the case where the contact doesn't exist in blocks prior to contract deployment then its not a big deal.

One reason I could see for specifying the actual deployment block here is if it makes a difference in subgraph syncing time. I made another comment in this review so feel free to answer my question there.

packages/subgraph/networks.yaml
```diff
@@ -45,4 +48,7 @@ rinkeby:
     minter:
       address: "e5056d1fE48F562dd8AcB1BBFFa147223d448C1B"
       startBlock: 0
+    ticketBroker:
+      address: "40897Ff6e44Ca2e145A691701897844669Bf7cee"
+      startBlock: 0
```
yondonfu: Would subgraph syncing on Rinkeby be faster if we specified a start block that reflected the block at which contracts were deployed?

packages/subgraph/src/mappings/ticketBroker.ts
```diff
@@ -0,0 +1,32 @@
+import { WinningTicketRedeemed } from '../types/TicketBroker/TicketBroker'
+import { WinningTicket, Stat, Transcoder } from '../types/schema'
+import { BigInt } from '@graphprotocol/graph-ts'
+
+export function winningTicketRedeemed(event: WinningTicketRedeemed): void {
+  let winningTicket = new WinningTicket(event.transaction.hash.toHex())
```
yondonfu: Since we talked about if offline figured I'd make a note here:

Worth considering using a different ID here instead of the tx hash since it is possible for multiple tickets to be redeemed in the same tx. In practice, it is unlikely for multiple tickets to be redeemed in the same tx right now because the node only redeems tickets one at a time, but that might not be true in the future.

Some options include:
- The [ticket hash](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/test/helpers/ticket.js#L34). Downside with this option is that we'd need to compute 2 keccak256 hashes for each event - one to compute [recipientRandHash](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/test/helpers/ticket.js#L21) from `recipientRand` and another to compute the ticket hash
- The tx hash + log index. I think this is one of the recommended options from docs for the Graph

If we do use something else besides the tx hash as the ID for tickets, is there an easy way to link a `WinningTicket` entity back to an event/tx? If not, might be useful to include the tx hash in the saved record - that way anyone querying the subgraph could easily link redeemed tickets back to the txs that included the tickets.

packages/subgraph/subgraph.template.yaml
```diff
@@ -215,4 +215,24 @@ dataSources:
...
+      apiVersion: 0.0.3
+      language: wasm/assemblyscript
+      file: ./src/mappings/ticketBroker.ts   
+      entities:
+        - WinningTicket
+      abis:
+        - name: TicketBroker
+          file: ./abis/TicketBroker.json
+      eventHandlers:
+        - event: WinningTicketRedeemed(indexed address,indexed address,uint256,uint256,uint256,uint256,bytes)
```
yondonfu, line 237: Is the indexed keyword necessary for the event signature? I ask because it would result in different event signature hashes, but the subgraph seems to be syncing properly so seems like it either is necessary or ignored?

packages/explorer-2.0/layouts/main.tsx
adamsoffer: Yeah, neat right? The most recent version of Typescript v3.7 introduced optional chaining. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

packages/subgraph/src/mappings/ticketBroker.ts
adamsoffer: I tried option 1 earlier this using the keccak256 function provided in the graph's assemblyscript library, `graph-ts`,  but I was getting an error for some reason.

I'll like option 2. I can add a `txHash` field to the `WinningTicket` record 👍

packages/subgraph/subgraph.template.yaml
adamsoffer: Yeah the graph throws a compilation error if the `indexed` keyword is omitted 🤷‍♂️

packages/subgraph/networks.yaml
adamsoffer: Yeah, it would speed it up. I'll specify start blocks 👍

adamsoffer:
Thanks for the review @yondonfu ! I added your suggested changes and rebased.

packages/explorer-2.0/data/unlock-tokens.mdx
```diff
@@ -0,0 +1,2 @@
+In order to stake your LPT you must first grant permissions to the Livepeer
```
yondonfu, line 1: The BondingManager contract is usually described as one word and is technically the actual name in Solidity so perhaps we can stick with that convention here.

yondonfu:
Looks like API errors are causing CI to fail:

Screen Shot 2020-02-10 at 6 13 43 PM

packages/explorer-2.0/data/unlock-tokens.mdx
adamsoffer: 👍

adamsoffer:
@yondonfu reran build inside circleci and it passed. Must be a race condition like you suggested 🤷‍♂️

yondonfu (approved):
LGTM!

PR #557, opened 2020-02-10
Update SDK to be compatible with Streamflow

adamsoffer:
This PR updates the SDK with the latest changes introduced by streamflow.

- Removes all instances of deprecated JobsManager
- Adds option to return a transaction hash from each method 
- Updates ABI files

- Ran `yarn test`. All unit tests pass.
 
**Checklist:**


kyriediculous (changes requested):
We actually don't need many of these ABIs for the SDK which is why I would be in favor of extracting only the necessary ones into a seperate package and maintaining that. We _could_ set up some CI tool that rebuilds that package when there's changes to the `livepeer/protocol` repo. 

ABIs that aren't used (or most of them, I haven't checked too thorough) : Assertions, Interfaces, Mocks, inherited contracts*

_* eg. TicketBroker ABI contains all of the internal types of eg MReserve , because the encapsulating TicketBroker contract inherits it_ 

I would be in favor of just removing the unused ABIs for now and creating a seperate issue for a more robust way to keep the ABIs in sync with the protocol branch. 

cc @iameli @yondon what do you think?

packages/sdk/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@livepeer/sdk",
   "description": "A module for interacting with Livepeer's smart contracts. The SDK is a core dependency of most LivepeerJS packages.",
-  "version": "1.0.0-streamflow-alpha.17",
```
kyriediculous: Does this get changed automatically on each commit? 
Might be interesting to set up releases or per PR bumps.

packages/sdk/src/index.js
```diff
@@ -1956,19 +1938,14 @@ export async function createLivepeerSDK(
      * // }
      */
     async withdrawStake(
-      unbondingLockId: string,
+      unbondingLockId = null,
```
kyriediculous: I think we shouldn't set up the default to be `null` as that would lead to failure  anyway. 
What do you think about considering a null check and fail early instead?

kyriediculous (changes requested):

packages/sdk/package.json
adamsoffer: It doesn't. I didn't mean to commit that package version change. Reverted. Definitely need to think about how to best handle releases across packages in the monorepo.

packages/sdk/src/index.js
adamsoffer: Agreed. I now perform a null check and throw an error if argument is null.

adamsoffer:
Thanks @kyriediculous. I made the requested change. I also cleaned up the commit history so all the ABI updates are in a separate commit now.

packages/sdk/package.json
kyriediculous: Rethinking this a bit , it doesn't really matter if it isn't published to npm , but might be good taking a proactive approach here. We might also want to consider publishing certain package to npm possibly (again out of scope here, just spewing ideas).

packages/sdk/src/index.js
kyriediculous: Also here it's unfortunate that I can't see the in-between diff between that review and now, only the end-result. Related to https://github.com/livepeer/livepeerjs/pull/570#issuecomment-590313254

packages/sdk/src/index.js
```diff
@@ -2190,29 +1841,20 @@ export async function createLivepeerSDK(
...
+      unbondLockId: string,
       tx = config.defaultTx,
-      unbondingLockId = null,
     ): Promise<TxReceipt> {
-      const {
-        status,
-        withdrawAmount,
-        nextUnbondingLockId,
-      } = await rpc.getDelegator(tx.from)
+      if (!unbondLockId) {
```
kyriediculous, line 1847: This threw me off for a second as I forgot empty strings are falsy values in JS 😄

kyriediculous (approved):
LGTM 👍 

I have created an issue regarding the CI stuff in the protocol repo : https://github.com/livepeer/protocol/issues/371

Issue #534, opened 2020-01-27
explorer: add ability to withdraw earned fees


adamsoffer:
**What's the problem? (required)**
As a tokenholder, I am unable to withdraw my earned fees from the explorer
 
**What's the current behavior? (required)**
Tokenholders must use either MyCrypto or the old explorer to withdraw fees 

**Describe a possible solution you've considered (optional)**
Design and implement a "fees" component with a call to action for withdrawing fees

PR #561, opened 2020-02-10
add "earned fees" view to explorer

adamsoffer:
The PR adds an "earned fees" view to the account profile page, where users can view and withdraw earned fees. It also includes some refactoring to improve the page load time of profile pages, a Fortmatic wallet integration, and a couple fixes described below.

- adds "earned fees" tab to account profile pages
- adds a `withdrawnFees` field to the subgraph `Delegator` entity for indexing total withdrawn fees.
- refactors some data fetching code to clean things up and improve the page load time on profile pages. Specifically, I'm now fetching all profile data in a single graphql request as opposed to 3.
- adds Fortmatic wallet integration. (#552) FYI I'm including this feature as part of this PR because the Fortmatic team asked if we can have it deployed in time before ETHDenver this Friday. 
- fixes incorrect claim earnings and approval amounts in explorer's history tab (#554, #559)

All functionality and fixes tested on [beta.explorer.livepeer.org](https://beta.explorer.livepeer.org)

#534 #552 #554 #559  

**Screenshots (optional):**

Screen Shot 2020-02-10 at 4 42 14 PM

**Checklist:**


packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -151,10 +151,10 @@ export function transcoderSlashed(event: TranscoderSlashed): void {
   let bondingManager = BondingManager.bind(event.address)
   let roundsManager = getRoundsManagerInstance(dataSource.network())
   let currentRound = roundsManager.currentRound()
-  let totalStake = bondingManager.transcoderTotalStake(transcoderAddress)
+  let delegateData = bondingManager.getDelegator(transcoderAddress)
 
   // Update transcoder total stake
-  transcoder.totalStake = totalStake
+  transcoder.totalStake = delegateData.value3
```
yondonfu, line 157: Ah can we not reference named fields (i.e. `delegateData.delegatedAmount`) because the subgraph libraries don't support resolving the Solidity struct named fields? Is the convention used by the libraries `value0`, `value1`, etc.?

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -374,48 +370,50 @@ export function reward(event: Reward): void {
...
-  let claimEarningsEventID = call.transaction.hash.toHex() + '-ClaimEarnings'
-  let claimEarningsEvent = ClaimEarningsEvent.load(claimEarningsEventID)
-
   // The Streamflow release introduced an event emitter for EarningsClaimed, so
-  // we can ignore this call handler henceforth
-  if (claimEarningsEvent != null) {
-    return
+  // we can ignore this call handler henceforth after the block in which the
+  // protocol was paused prior to the streamflow upgrade
+  if (call.block.number.le(BigInt.fromI32(9274600))) {
```
yondonfu: Technically, the updated BondingManager contract with this event didn't exist until the contract was deployed at block [9274414](https://etherscan.io/tx/0x0461475a8daf6b028aa49ae4a7eb283c80b6022bf3514410175c542106898061) - WDYT about using that block number since it is corresponds with contract creation?

packages/subgraph/src/mappings/livepeerToken.ts
```diff
@@ -1,13 +1,12 @@
...
-import { LivepeerToken, Approval } from '../types/LivepeerToken/LivepeerToken'
+import { Approval } from '../types/LivepeerToken/LivepeerToken'
 import { Delegator, ApprovalEvent } from '../types/schema'
 import { getRoundsManagerInstance } from './util'
 
 // Handler for NewRound events
 export function approval(event: Approval): void {
-  let livepeerToken = LivepeerToken.bind(event.address)
   let owner = event.params.owner
-  let spender = event.params.spender
```
yondonfu, line 10: Should we record the `spender`? I'm not sure how this indexed data is used, but at the moment could it be possible to have an incorrect allowance for a user if the same account that sets an allowance for the BondingManager also sets an allowance for trading on a DEX (i.e. the account approves X LPT for the BondingManager and a 0x exchange contract). Probably not too common so if this is the case not the biggest priority to handle, but worth tracking.

packages/explorer-2.0/lib/connectors.tsx
```diff
@@ -17,3 +18,8 @@ export const Portis = new PortisConnector({
   dAppId: '0e9ac0c3-9184-4660-8492-6989cf3dc5d4',
   networks: [1],
 })
+
+export const Fortmatic = new FortmaticConnector({
+  apiKey: 'pk_live_48F5D30F4D931706',
```
yondonfu: Should this be an env var?

packages/explorer-2.0/lib/connectors.tsx
```diff
@@ -17,3 +18,8 @@ export const Portis = new PortisConnector({
...
+  chainId: 1,
```
yondonfu, line 24: Does Fortmatic support other networks besides mainnet?

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from 'graphql-tools'
 
 const subgraphEndpoint =
-  'https://api.thegraph.com/subgraphs/name/livepeer/livepeer'
+  'https://api.thegraph.com/subgraphs/name/livepeer/livepeer-canary'
```
yondonfu: Are we still maintaining the original subgraph?

yondonfu (commented):
Left a few comments and I'm going to play with the staging link this afternoon.

packages/explorer-2.0/lib/createSchema.tsx
adamsoffer: oops, yes we are still maintaining the original subgraph. This endpoint shouldn't change. I changed it to `livepeer-canary` for testing purposes locally. I'll revert back to `livepeer/livepeer`.

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: Yeah, not great for readability. Unfortunately the subgraph lib doesn't support resolving the struct named fields.

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: 👍

packages/subgraph/src/mappings/livepeerToken.ts
adamsoffer: yeah good point. I'll add that to the record 👍

packages/explorer-2.0/lib/connectors.tsx
adamsoffer: It does provide a separate API key for other networks such as rinkeby.

packages/explorer-2.0/lib/connectors.tsx
adamsoffer: 👍

adamsoffer:
Thanks @yondonfu - just pushed up suggested changes.

Unrelated changes I also included:
- Added missing ga tracking snippet
- Added an `.npmrc` file to the project root until [this](https://github.com/yarnpkg/yarn/issues/7902#issuecomment-584874383) is resolved.
- removed some dead code

packages/explorer-2.0/lib/connectors.tsx
adamsoffer: One thing I'd like to address in a separate PR is conditionally passing in rinkeby specific params such as this when the app detects a network change.

yondonfu (approved):
LGTM 🚢

PR #567, opened 2020-02-14
explorer: add orchestrator table sorting

### LINKED ISSUE #556: explorer: Sort orchestrator list by different options

Some users may want to sift through the list based on another metric besides stake, but it is always sorted by stake right now.

The orchestrator list in the explorer is always sorted by stake. 

Add an option to sort the list by different metrics. The options could just be the different columns:
- Stake
- Fees
- Reward cut
- Fee share
- Reward calls

The last one probably is not as important as the others.

adamsoffer:
This PR adds the ability for users to sort table columns in the orchestrators view by clicking column header titles. Available for testing on [staging url](https://beta.explorer.livepeer.org).

- Adds sorting to each orchestrator column. Clicking once will sort ascending, twice descending, thrice will reset.
- Updates `react-table` dependency

Fixes #556

**Checklist:**


kyriediculous:
On initial load we sort by stake but the sorting arrow isn't there in the table header, would it be possible to show this so that there is a visual indication that columns are now sortable? I tried looking in the commit where it's missing but it's been a while since I've looked at react stuff 😅

kyriediculous (changes requested):
On initial load we sort by stake but the sorting arrow isn't there in the table header, would it be possible to show this so that there is a visual indication that columns are now sortable? I tried looking in the commit where it's missing but it's been a while since I've looked at react stuff 😅

adamsoffer:
> On initial load we sort by stake but the sorting arrow isn't there in the table header, would it be possible to show this so that there is a visual indication that columns are now sortable? I tried looking in the commit where it's missing but it's been a while since I've looked at react stuff 😅

@kyriediculous Good call. I added a desc sort icon on the initial page load. It's up on staging.

packages/explorer-2.0/components/NetworkWidget/index.tsx
```diff
@@ -14,14 +14,19 @@ const GET_ROUND = gql`
...
     pollInterval: 20000,
+    ssr: false,
   })
 
   if (loading) {
     return null
   }
 
+  if (error) {
+    console.log(error)
```
kyriediculous: Shouldn't have console logs according to convention I think? Isn't this an ESLINT thing?

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -15,7 +15,7 @@ export default ({ delegator, protocol, currentRound }) => {
   const query = router.query
   const account = query.account as string
   const context = useWeb3React()
-  const isMyAccount = account === context.account
+  const isMyAccount = account.toLowerCase() === context?.account?.toLowerCase()
```
kyriediculous: We should not convert addresses to lowercase but use the checksums of the addresses `web3.utils.toChecksumAddress(address)`

Additionally we should check if both addresses are checksum addresses before comparing them `web3.utils.checkAddressChecksum(address)` 

https://web3js.readthedocs.io/en/v1.2.0/web3-utils.html#tochecksumaddress

kyriediculous (changes requested):

kyriediculous:
Looks good, just one more request regarding address checksums. Should use those across the board, especially in functions that involve `eth_send(Raw)Transaction`

packages/explorer-2.0/components/NetworkWidget/index.tsx
adamsoffer: Yeah it is. I've removed console logs 👍

packages/explorer-2.0/components/StakingView/index.tsx
adamsoffer: I created a helper function `checkAddressEquality(address1, address2)` that converts to checksum addresses first. I'm not using `checkAddressChecksum(address)` because I'm grabbing the address from the url and we can't guarantee that it will be valid checksum.

adamsoffer:
@kyriediculous thanks for the review! Good to squash and merge?

kyriediculous:
Changes look good !

kyriediculous (approved):

Issue #565, opened 2020-02-12
explorer: add rinkeby support


adamsoffer:
**What's the problem? (required)**
The new explorer doesn't currently support rinkeby. 

**What's the current behavior? (required)**
When switching to the rinkeby network on metamask, the explorer continues to display data on mainnet. It should interact with and reflect what's on the rinkeby network.

**Describe a possible solution you've considered (optional)**
Upon detecting a network change:
1. Instantiate the livepeer SDK with the rinkeby controller address
2. Instantiate apollo client with the rinkeby subgraph endpoint
3. Instantiate wallet providers with rinkeby api keys

yondonfu:
Closed by #570

PR #570, opened 2020-02-18
explorer: web3-react refactor and rinkeby support

adamsoffer:
This PR adds Rinkeby support to the explorer via network change detection. Inspired by uniswap's frontend, it also includes a refactor of the way the `web3-react` framework is implemented in an effort to improve how the app persists web3 wallet connections (no more annoying metamask popups!). Available for testing on [staging server](https://beta.explorer.livepeer.org).

- added rinkeby support
- refactored web3-react implementation
- added `react-ga` dependency for simplifying google analytics event tracking

- Tested by changing network from mainnet to rinkeby and back again using metamask.

#565 

**Checklist:**


packages/explorer-2.0/lib/connectors.tsx
```diff
@@ -3,22 +3,29 @@ import { PortisConnector } from '@web3-react/portis-connector'
 import { NetworkConnector } from '@web3-react/network-connector'
 import { FortmaticConnector } from '@web3-react/fortmatic-connector'
 
-const POLLING_INTERVAL = 10000
+const POLLING_INTERVAL = 12000
```
kyriediculous, line 6: any reason for the change ? I don't think this has to approximate the block time. 
We multiply this on L15, perhaps we can just put this at something like 30 seconds then? Or would that feel too unreactive?

packages/explorer-2.0/lib/connectors.tsx
```diff
@@ -3,22 +3,29 @@ import { PortisConnector } from '@web3-react/portis-connector'
...
+const RPC_URLS: { [chainId: number]: string } = {
+  1: process.env.RPC_URL_1 as string,
+  4: process.env.RPC_URL_4 as string,
+}
 
 export const Network = new NetworkConnector({
-  urls: { 1: 'https://mainnet.infura.io/v3/39df858a55ee42f4b2a8121978f9f98e' },
+  defaultChainId: 1,
+  urls: { 1: RPC_URLS[1], 4: RPC_URLS[4] },
```
kyriediculous: since this isn't strongly typed we could just do `urls: RPC_URLS` here

kyriediculous (changes requested):

packages/explorer-2.0/lib/connectors.tsx
adamsoffer: no good reason. I just noticed that's what [the example used](https://github.com/NoahZinsmeister/web3-react/blob/v6/example/connectors.ts#L14) in the `web3-react` framework lol

kyriediculous: I see... I personally think it's a bad example of them to try to approximate the blocktime.
Since blocktimes aren't exactly 12 seconds, the poller will run out of sync anyway. Worst case scenario we are only polling for a discovered block 11.99999... seconds after it has appeared. This could lead to confusion for novice devs building web3 dapps. In addition , Rinkeby has 5 sec blocktimes. 

I think it's more important to optimize for UX and memory here rather than blocktimes. 

What I've done personally in the past is immediately refresh state that pertains to the user itself when the user takes an action. For other state we can just rely on the polling interval and it would be fine to just have this at 30 seconds. Alternatively use pub-sub RPC but this is experimental (which is why it has been removed in go-livepeer as well). 

That being said I don't think an optimization here needs to be part of this PR as regardless of this interval the support for rinkeby works.

kyriediculous:
As a rule of thumb I think it would be good to not force-push until the review process has been completed. This allows the reviewer a diff between the new updates and the last review (by just selecting the added fixup commit) as well as consistent commit history. 

A process we've been using in the protocol team for `go-livepeer` (thanks to Yondon ! ) is as follows: 
1.  we submit `fixup` commits during the review process using 
`git commit --fixup <previous commit hash>` 

3. When all changes have been accepted by the reviewer we leave a comment to rebase the commit history and auto-squash the fixups (but don't approve the PR yet)
`git rebase -i --autosquash <branch, most often master>`

4. After rebasing locally, force-push the branch 
`git push --force-with-lease`

5. The reviewer will now accept the PR for merging. 

As an example, [this last commit](https://github.com/livepeer/livepeerjs/pull/570/commits/85113850927534a036ec8b9c37d023aa3dcb524b) , could have been submitted as fixup to `add rinkeby support` 

As more people start working on this repo it could be a good idea to consider a process everyone working on the repo follows. 

Here's the blog post Yondon sent me to better understand  fixup commits and rebasing https://thoughtbot.com/blog/autosquashing-git-commits

kyriediculous (approved):
LGTM 👍 

As mentioned let's address polling interval optimisation separately if we really want to (observing memory usage might be a good idea here). 

I'd also squash the last commit into `add rinkeby support` but I'll leave that up to you.

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -111,3 +111,70 @@ export const textTruncate = (str, length, ending) => {
...
+const networksTypes = {
+  1: 'mainnet',
+  2: 'morden',
+  3: 'ropsten',
+  42: 'kovan',
+  4: 'rinkeby',
+}
+
+const networksIds = {
+  main: 1,
```
yondonfu, line 127: Is this a duplicate of `mainnet`?

adamsoffer: ah yea it is. I'll remove.

yondonfu:
Looks like the orchestrator table view does not refresh when toggling between mainnet & Rinkeby.

Reproduction steps:
1. Connect with Metamask
2. Connect to mainnet in Metamask
3. Observe that the network indicator says "Mainnet" and that the orchestrator table view reflects the list of orchestrators on mainnet
4. Connect to Rinkeby in Metamask
5. Observe that the network indicator says "Rinkeby"
6. Observe that the orchestrator table view still reflects the list of orchestrators on mainnet

I had to manually refresh the page in order for the orchestrator table view to be populated with the list of orchestrators on Rinkeby.

I don't this issue should block this PR though and we can track it in a separate issue.

packages/explorer-2.0/lib/utils.tsx
adamsoffer: This utility function is based on [this package ](https://github.com/miguelmota/web3-detect-network). Wondering if that's actually intended.

adamsoffer:
> As a rule of thumb I think it would be good to not force-push until the review process has been completed. This allows the reviewer a diff between the new updates and the last review (by just selecting the added fixup commit) as well as consistent commit history.
> 
> A process we've been using in the protocol team for `go-livepeer` (thanks to Yondon ! ) is as follows:
> 
> 1. we submit `fixup` commits during the review process using
>    `git commit --fixup <previous commit hash>`
> 2. When all changes have been accepted by the reviewer we leave a comment to rebase the commit history and auto-squash the fixups (but don't approve the PR yet)
>    `git rebase -i --autosquash <branch, most often master>`
> 3. After rebasing locally, force-push the branch
>    `git push --force-with-lease`
> 4. The reviewer will now accept the PR for merging.
> 
> As an example, [this last commit](https://github.com/livepeer/livepeerjs/pull/570/commits/85113850927534a036ec8b9c37d023aa3dcb524b) , could have been submitted as fixup to `add rinkeby support`
> 
> As more people start working on this repo it could be a good idea to consider a process everyone working on the repo follows.
> 
> Here's the blog post Yondon sent me to better understand fixup commits and rebasing https://thoughtbot.com/blog/autosquashing-git-commits

This process looks great. Let's use it moving forward. Question - is it necessary to autosquash at the end since github has the "squash and merge" button?

yondonfu:
> Looks like the orchestrator table view does not refresh when toggling between mainnet & Rinkeby.

As discussed offline, it turns out that the orchestrator table view *does* refresh when toggling between mainnet & Rinkeby, the refresh is just delayed. Tracking in #576 

> Question - is it necessary to autosquash at the end since github has the "squash and merge" button?

I think this comes down to personal preference. IIUC, GH's "squash and merge" button will always squash all commits into a single commit. This is nice and easy if the contributor wants everything to be in a single commit. The benefit of doing the local rebase when squashing fixup commits is that the contributor has control over the commit history so the contributor can choose to merge the PR with commits X, Y and Z (each commit would contain its own distinct set of changes that the contributor wants to keep separate from each other) as opposed to merging the PR with a single commit. It also gives the maintainer/reviewer the opportunity to do a final code check + commit history check prior to merging.

adamsoffer:
> > Looks like the orchestrator table view does not refresh when toggling between mainnet & Rinkeby.
> 
> As discussed offline, it turns out that the orchestrator table view _does_ refresh when toggling between mainnet & Rinkeby, the refresh is just delayed. Tracking in #576
> 
> > Question - is it necessary to autosquash at the end since github has the "squash and merge" button?
> 
> I think this comes down to personal preference. IIUC, GH's "squash and merge" button will always squash all commits into a single commit. This is nice and easy if the contributor wants everything to be in a single commit. The benefit of doing the local rebase when squashing fixup commits is that the contributor has control over the commit history so the contributor can choose to merge the PR with commits X, Y and Z (each commit would contain its own distinct set of changes that the contributor wants to keep separate from each other) as opposed to merging the PR with a single commit. It also gives the maintainer/reviewer the opportunity to do a final code check + commit history check prior to merging.

ah yeah makes sense!

Issue #571, opened 2020-02-18
subgraph: Index TicketBroker deposit/reserve events


yondonfu:
**What's the problem? (required)**
Describe the problem or need in exactly 1 sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

The subgraph currently does not index deposit/reserve events.

**Describe a possible solution you've considered (optional)**

The relevant events:

- [DepositFunded](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/contracts/pm/mixins/interfaces/MTicketBrokerCore.sol#L16)
- [ReserveFunded](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/contracts/pm/mixins/interfaces/MReserve.sol#L13)
- [WinningTicketRedeemed](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/contracts/pm/mixins/interfaces/MTicketBrokerCore.sol#L18)
- [ReserveClaimed](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/contracts/pm/mixins/interfaces/MReserve.sol#L15)
- [Withdrawal](https://github.com/livepeer/protocol/blob/d6bcb03337839071f1a90e391760ef1f7660f836/contracts/pm/mixins/interfaces/MTicketBrokerCore.sol#L34)

Sketch of indexing logic:

- A broadcaster entity has a `deposit` and a `reserve`
- When `DepositFunded` is emitted, add `amount` to `deposit` for address `sender`
- When `ReserveFunded` is emitted, add `amount` to `reserve` for address `reserveHolder`
- When `ReserveClaimed` is emitted, subtract `amount` from `reserve` for address `reserveHolder`
- When `WinningTicketRedeemed` is emitted:
    - If `faceValue` is greater than `deposit`, set `deposit` to 0
    - If `faceValue` is less than or equal to `deposit`, subtract `faceValue` from `deposit`
- When `Withdrawal` is emitted, set `deposit` and `reserve` to 0

**Additional context (optional)**
Add any other context or screenshots about the enhancement or feature request here.

yondonfu:
Closed by #575

PR #575, opened 2020-02-24
subgraph: index TicketBroker deposit/reserve events

adamsoffer:
This PR includes an update to the subgraph for indexing TicketBroker deposit/reserve events as well as some other miscellaneous updates detailed below.

- adds `Broadcaster` entity with `deposit` and `reserve` fields
- adds `totalGeneratedFees` `totalNewStake` and `totalMovedStake` to `Round` entity
- renames `accruedFees` to `totalGeneratedFees` in `Transcoder` entity to be consistent
- updates README with instructions on how to run subgraph locally
- patches graph-cli dep (the graph pushed up a fix to this package but haven't yet published to npm)

Successfully deployed and synced on [livepeer-canary](https://thegraph.com/explorer/subgraph/livepeer/livepeer-canary) subgraph

#571 

**Checklist:**


packages/subgraph/README.md
```diff
@@ -21,3 +21,19 @@ required by The Graph.
...
 variable or directly by choosing a different "prepare" script. See [package.json](./package.json) for all options.
+
+### Deploy the Livepeer Subgraph to the Graph's hosted service (recommended)
+Follow the instructions documented [here](https://thegraph.com/docs/deploy-a-subgraph).
+
+### Deploy the Livepeer Subgraph locally
+
+1. Install [Docker](https://docs.docker.com) and [Docker Compose](https://docs.docker.com/compose/install/)
+2. Inside `docker-compose.yml` set the ethereum network to an archive node that has tracing enabled. If you don't have access to an archive node we recommend using [Alchemy](https://alchemyapi.io/). 
+3. In the root of this project run `docker-compose up`. This command will look for the `docker-compose.yml` file and automatically provision a server with rust, postgres, and ipfs, and spin up a graph node with a GraphiQL interface at `http://127.0.0.1:8000/`.)
```
yondonfu: Extra parens at the end?

packages/subgraph/README.md
```diff
@@ -21,3 +21,19 @@ required by The Graph.
...
 network (Mainnet, Goerli, Kovan, Rinkeby, Ropsten). The network can be changed via the `NETWORK_NAME` environment
 variable or directly by choosing a different "prepare" script. See [package.json](./package.json) for all options.
+
+### Deploy the Livepeer Subgraph to the Graph's hosted service (recommended)
+Follow the instructions documented [here](https://thegraph.com/docs/deploy-a-subgraph).
+
+### Deploy the Livepeer Subgraph locally
+
+1. Install [Docker](https://docs.docker.com) and [Docker Compose](https://docs.docker.com/compose/install/)
+2. Inside `docker-compose.yml` set the ethereum network to an archive node that has tracing enabled. If you don't have access to an archive node we recommend using [Alchemy](https://alchemyapi.io/).
```
yondonfu: Might be helpful to note the key name (i.e. `ethereum` under the `environment` section) within the YAML file for which a value should be specified.

packages/subgraph/schema.graphql
```diff
@@ -115,11 +121,15 @@ type Delegator @entity {
 }
 
 """
-The possible statuses of a transcoder
+The Livepeer protocol is round based and each round is represented by some number of Ethereum blocks.
```
yondonfu: Seems like this comment is in the wrong place (doesn't describe the `Broadcaster` entity)?

packages/subgraph/src/mappings/ticketBroker.ts
```diff
@@ -16,20 +33,92 @@ export function winningTicketRedeemed(event: WinningTicketRedeemed): void {
   winningTicket.txHash = event.transaction.hash.toHex()
   winningTicket.save()
 
+  let winningTicketRedeemedEvent = new WinningTicketRedeemedEvent(
+    event.transaction.hash.toHex() + 'WinningTicketRedeemedEvent',
```
yondonfu: We currently use [tx hash + log index](https://github.com/livepeer/livepeerjs/blob/f80f47baa02f8a09ec1fe00b7d8acc67d5c5e723/packages/subgraph/src/mappings/ticketBroker.ts#L7) in the identifier since there could be multiple events in the same transaction - should we do the same here?

packages/subgraph/schema.graphql
```diff
@@ -140,6 +150,32 @@ type UnbondingLock @entity {
   withdrawRound: BigInt
 }
 
+type WinningTicket @entity {
```
yondonfu: Why do we need an `WinningTicket` entity and a `WinningTicketRedeemedEvent` entity? Seems like `WinningTicketRedeemedEvent` contains a superset of the fields of `WinningTicket`.

packages/explorer-2.0/queries/orchestratorsView.gql
```diff
@@ -8,13 +8,11 @@
...
   ) {
     id
-    active
-    accruedFees
+    totalGeneratedFees
     feeShare
     activationRound
     deactivationRound
     rewardCut
-    active
```
yondonfu, line 17: We don't need the `active` field anymore?

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -200,6 +200,11 @@ export function bond(event: Bond): void {
     delegator = new Delegator(delegatorAddress.toHex())
   }
 
+  let bondedAmount = delegator.bondedAmount as BigInt
+
+  // additional amount == new bondedAmount minus previous bonded amount
+  let additionalAmount = delegatorData.value0.minus(bondedAmount)
```
yondonfu: I think this will include the new LPT staked by the delegator as well as the delegator's unclaimed rewards since the last time its `bondedAmount` field was updated in the DB. I don't think we should count the delegator's unclaimed rewards to `additionalAmount` because:

1. The unclaimed rewards do not consist of new LPT being staked by the delegator (rather they consist of new LPT that is minted and automatically staked)
2. The `Bond` event handler in [bondingManager_LIP11.ts](https://github.com/livepeer/livepeerjs/pull/575/files#diff-f99a2db018ec7d0de146285dc8f28e3cR85) has access to an `additionalAmount` field that only counts new LPT being staked by the delegator and not unclaimed rewards. So, as-is, this `Bond` event handler mapping's behavior will be different from the behavior of the `bondingManager_LIP11` `Bond` event handler mapping.

Handling this might not be that straightforward...

- One solution could be to index LPT `Transfer` events and when this `Bond` event handler fires check if we have an LPT `Transfer` event with `from = delegator's address` and `to = Minter address` in the same transaction then the [value field](https://github.com/livepeer/protocol/blob/cbcdf50b5e7cb93b61da009824ca6dae9fe34ae7/contracts/bonding/BondingManager.sol#L333) will be the new LPT staked by the delegator. This isn't full proof though since in theory an address could send LPT directly to the Minter without it being staked with the BondingManager

- Another solution could be to call `pendingStake()` at the block number right before the block in which the new `Bond` event is included in. The delegator's last recorded `bondedAmount` could be subtracted from the value returned to get the unclaimed rewards. Then, the unclaimed rewards could be subtracted from the `delegatorData.value0.minus(bondedAmount)` to get the new LPT staked by the delegator. I think this might work, but I'm not sure about the practicality of calling `pendingStake()` at a past block number in the event handler

yondonfu (changes requested):

packages/subgraph/schema.graphql
```diff
@@ -93,6 +93,9 @@ type Round @entity {
   totalGeneratedFees: BigInt
 }
 
+"""
+Bonded accounts who have delegated their stake towards a transcoder candidate
```
kyriediculous, line 97: Let's be proactive about adopting ubiquitous streamflow languageh ere. 

"towards an orchestrator"

packages/subgraph/schema.graphql
```diff
@@ -121,7 +124,7 @@ type Delegator @entity {
 }
 
 """
-The Livepeer protocol is round based and each round is represented by some number of Ethereum blocks.
+Broadcasters pay transcoders to do the work of transcoding in exchange for fees
```
kyriediculous, line 127: "pay orchestrators for transcoding work"

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: Confirmed `additionalAmount` included unclaimed rewards which we don't want. I pushed a fix that replaces the `Bond` event handler with a call handler which gives us access to a delegator's inputted amount.

packages/subgraph/src/mappings/bondingManager_LIP12.ts
```diff
@@ -39,24 +39,24 @@ export function transcoderUpdated(event: TranscoderUpdate): void {
...
+  transcoderUpdated.gasUsed = event.transaction.gasUsed
+  transcoderUpdated.gasPrice = event.transaction.gasPrice
+  transcoderUpdated.timestamp = event.block.timestamp
+  transcoderUpdated.from = event.transaction.from.toHex()
+  transcoderUpdated.to = event.transaction.to.toHex()
+  transcoderUpdated.round = currentRound.toString()
+  transcoderUpdated.rewardCut = rewardCut
+  transcoderUpdated.feeShare = feeShare
+  transcoderUpdated.delegate = transcoderAddress.toHex()
+  transcoderUpdated.save()
```
kyriediculous, line 60: Kinda of out of scope and more of a question than a remark as this goes for any type similar to this one. 

Why are we not simply passing the event into the constructor ? 

Alternatively we can use a method that calls the constructor instead, validates the necessary fields and can return an error if necessary.

packages/subgraph/src/mappings/bondingManager_LIP12.ts
```diff
@@ -39,24 +39,24 @@ export function transcoderUpdated(event: TranscoderUpdate): void {
   transcoder.save()
 
   // Store transaction info
-  let transcoderUpdatedEvent = new TranscoderUpdatedEvent(
-    event.transaction.hash.toHex() + '-TranscoderUpdate',
+  let transcoderUpdated = new TranscoderUpdated(
+    event.transaction.hash.toHex() + '-' + event.logIndex.toString(),
```
kyriediculous: we can extract this into a helper , something like `GetEventID` ?  This goes for every one of these constructor calls that set the ID.

kyriediculous (changes requested):

packages/explorer-2.0/queries/orchestratorsView.gql
adamsoffer: As of streamflow, in the Explorer, we only use `activationRound` and `deactivationRound` to determine if an O is active so we no longer need to ask for the `active` field. 

These are the ways I can query the active and pending active sets in th explorer as of now:
1. Current active set: `activationRound <= currentRound && deactivationRound > currentRound`
2. Pending active: `activationRound > currentRound  && deactivationRound > currentRound`
3. Pending inactive: `activationRound <= currentRound &&  deactivationRound == currentRound + 1`

I pushed an update that sets a transcoder's `active` state whenever a `transcoderActivated` or `transcoderDeactivated` event gets emitted, so we can query transcoders by the `active` field again if we want and without needing the current round ahead of time, but important to note this query would exclude transcoders who are active and set to become inactive in the following round, and it includes transcoders who are inactive but are set to become active in the following round.  I believe the only way to keep track of the current active set using the `active` field would be to loop through all registered transcoders upon new round events and call `isActiveTranscoder` but there's no easy way to query all registered transcoders in a mapping right now and it also wouldn't scale very well.

packages/subgraph/schema.graphql
adamsoffer: Yeah, I was going to use "orchestrator" but decided to stick with "transcoder" simply because the smart contracts still refer to them as transcoders and I want to maintain parity with them. I think if we use the term Orchestrator in one place we should update it everywhere else in the subgraph, which maybe we want to do. Worth a discussion.

adamsoffer: My philosophy right now is to use "orchestrator" in all user-facing UIs but maintain parity with smart contracts in code.

kyriediculous: The problem with the contracts is that we cant' just change function signatures whenever we want to conform to a certain language, well we _could_, but that would be ill advised. Upgrading contracts is a risky endeavour and it would change the existing API surface for ABI bindings (eg JS contract bindings or Go contract bindings). 

Additionally in streamflow a transcoder is by definition not protocol aware.

Therefore I think not maintaining parity is fine but since I don't actively work on this repo I will leave the end decision up to you. There's some overlap in go-livepeer as well still.

adamsoffer: Yeah, agreed. Let's aim to rename all instances of the subgraph `Transcoder` entity to `Orchestrator`. Want to create an issue for this?

packages/subgraph/schema.graphql
adamsoffer: This is indeed duplicative. I've removed `WinningTicket` in favor of a single `WinningTicketRedeemed` entity that implements the Transaction interface defined in `schema.graphql`

packages/subgraph/src/mappings/ticketBroker.ts
adamsoffer: Yeah, we should do the same here. Fixed and removed the duplicative entity.

packages/subgraph/README.md
adamsoffer: removed

packages/subgraph/README.md
adamsoffer: I've updated readme to make this more clear.

packages/subgraph/schema.graphql
adamsoffer: Updated comment to describe `Broadcaster` entity

packages/subgraph/schema.graphql
adamsoffer: 👍I'll replace all instances of "transcoder" with "orchestrator" in another PR.

packages/subgraph/src/mappings/bondingManager_LIP12.ts
adamsoffer: Yeah, it's a good question. It appears the graph only allows you to instantiate an entity with an id. https://thegraph.com/docs/define-a-subgraph#writing-mappings

kyriediculous: Ahh these Classes are generated? Gotcha !

packages/subgraph/schema.graphql
kyriediculous: okay sounds good to me

packages/subgraph/src/mappings/bondingManager_LIP12.ts
adamsoffer: Created a `makeEventId` helper 👍

packages/subgraph/src/mappings/roundsManager.ts
```diff
@@ -28,7 +31,10 @@ export function newRound(event: NewRound): void {
   while (EMPTY_ADDRESS.toHex() != currentTranscoder.toHex()) {
     // Update transcoder active state
     active = bondingManager.isActiveTranscoder(currentTranscoder, roundNumber)
-    transcoder.active = active
+    transcoder.active = bondingManager.isActiveTranscoder(
```
yondonfu: Looks like there is already a `bondingManager.isActiveTranscoder()` call in the previous line.

adamsoffer: woops

adamsoffer: Fixed.

yondonfu (approved):
LGTM 🚢. Let's squash into a single commit before merging.

Nit: I noticed that 4b4f1dc contains a few changes that are not related to subgraph indexing (i.e. README updates and the graph-cli patch). For future reference, it could be helpful to keep those changes in separate commits. IMO this makes review easier and also helps with navigating commit history when trying to figure out where a change/bug was introduced. No need to do anything for this PR, but just wanted to make the suggestion.

adamsoffer:
> LGTM 🚢. Let's squash into a single commit before merging.
> 
> Nit: I noticed that [4b4f1dc](https://github.com/livepeer/livepeerjs/commit/4b4f1dc860e284fbf22f458f5f5611f86d3a7d08) contains a few changes that are not related to subgraph indexing (i.e. README updates and the graph-cli patch). For future reference, it could be helpful to keep those changes in separate commits. IMO this makes review easier and also helps with navigating commit history when trying to figure out where a change/bug was introduced. No need to do anything for this PR, but just wanted to make the suggestion.

Awesome. For sure. Commit readability will be top of mind on all future commits 👍

PR #578, opened 2020-03-02
fix 3box external account signature verification

adamsoffer:
This PR fixes an issue causing the 3box signature verification to fail when adding an external account. This bug was introduced when we switched the web3 library we pass into `web3-react` (`ethers.js` instead of `web3.js`) since ether.js doesn't have a `ecrecover` method. 

It also prevents the account page from reloading and displaying the loading indicator after successfully linking an external account.

- Instantiates web3 instance and uses `ecRecover` method for signature verification
- Checks if data has already been fetched and if so prevents loading indicator from taking over the page upon data refetching.

**Checklist:**

### DIFF
--- a/packages/explorer-2.0/components/EditProfile/index.tsx
+++ b/packages/explorer-2.0/components/EditProfile/index.tsx
@@ -117,8 +117,12 @@ export default ({ threeBoxSpace, refetch, account }: Props) => {
       if (hasExistingProfile(profile)) {
         setHasProfile(true)
       }
+
+      const Web3 = require('web3') // use web3 lib for ecRecover method
+      const web3 = new Web3(context.library._web3Provider)
+
       if (signature && ethereumAccount) {
-        const verifiedAccount = await context.library.eth.personal.ecRecover(
+        const verifiedAccount = await web3.eth.personal.ecRecover(
           message.replace(/<br ?\/?>/g, '\n'),
           signature,
         )
--- a/packages/explorer-2.0/components/EditProfile/index.tsx
+++ b/packages/explorer-2.0/components/EditProfile/index.tsx
@@ -18,6 +18,7 @@ import { useDebounce } from 'use-debounce'
 import ThreeBoxSteps from '../ThreeBoxSteps'
 import Spinner from '../Spinner'
 import { useWeb3React } from '@web3-react/core'
+import { ethers } from 'ethers'
 
 interface Props {
   account: string
@@ -118,11 +119,8 @@ export default ({ threeBoxSpace, refetch, account }: Props) => {
         setHasProfile(true)
       }
 
-      const Web3 = require('web3') // use web3 lib for ecRecover method
-      const web3 = new Web3(context.library._web3Provider)
-
       if (signature && ethereumAccount) {
-        const verifiedAccount = await web3.eth.personal.ecRecover(
+        let verifiedAccount = ethers.utils.verifyMessage(
           message.replace(/<br ?\/?>/g, '\n'),
           signature,
         )
--- a/packages/explorer-2.0/pages/accounts/[account]/[slug].tsx
+++ b/packages/explorer-2.0/pages/accounts/[account]/[slug].tsx
@@ -66,7 +66,10 @@ export default withApollo(() => {
     }
   }, [context.chainId])
 
-  if (loading || loadingMyAccount) {
+  if (
+    ((loading || loadingMyAccount) && !data) ||
+    (context.active && !dataMyAccount)
+  ) {
     return (
       <Layout>
         <Flex


packages/explorer-2.0/components/EditProfile/index.tsx
```diff
@@ -117,8 +117,12 @@ export default ({ threeBoxSpace, refetch, account }: Props) => {
       if (hasExistingProfile(profile)) {
         setHasProfile(true)
       }
+
+      const Web3 = require('web3') // use web3 lib for ecRecover method
```
kyriediculous: Ooh I see what's happening, no need to wrap the provider even here. That's just because web3 requires a provider to get the object that exposes the API. 

In ethersjs the EC functions are in utilities : https://docs.ethers.io/ethers.js/html/api-utils.html#elliptic-curve

kyriediculous:
````
let recovered = ethers.utils.verifyMessage(messageHashBytes, flatSig);
console.log(recovered === wallet.address);
// true
```

adamsoffer: ah good find thanks!

packages/explorer-2.0/components/EditProfile/index.tsx
```diff
@@ -118,11 +119,8 @@ export default ({ threeBoxSpace, refetch, account }: Props) => {
         setHasProfile(true)
       }
 
-      const Web3 = require('web3') // use web3 lib for ecRecover method
-      const web3 = new Web3(context.library._web3Provider)
-
       if (signature && ethereumAccount) {
-        const verifiedAccount = await web3.eth.personal.ecRecover(
+        let verifiedAccount = ethers.utils.verifyMessage(
```
kyriediculous, line 123: We need to pass in the message digest here as a uint8 array

```
let messageHash = arrayify(solidityKeccak256(message.replace(/<br ?\/?>/g, '\n')));
```

packages/explorer-2.0/components/EditProfile/index.tsx
```diff
@@ -18,6 +18,7 @@ import { useDebounce } from 'use-debounce'
 import ThreeBoxSteps from '../ThreeBoxSteps'
 import Spinner from '../Spinner'
 import { useWeb3React } from '@web3-react/core'
+import { ethers } from 'ethers'
```
kyriediculous, line 21: I think we can even do `import {arrayify ,solidityKeccak256, verifyMessage} from 'ethers/utils'` but I'll leave that up to you (see other comment)

kyriediculous (changes requested):

packages/explorer-2.0/components/EditProfile/index.tsx
adamsoffer: @kyriediculous hmm is this necessary? This works as is:
```js
let verifiedAccount = verifyMessage(
  message.replace(/<br ?\/?>/g, '\n'),
  signature,
)
```

kyriediculous: oh yes it will work as is too, my bad.

packages/explorer-2.0/components/EditProfile/index.tsx
adamsoffer: Oddly ethers throws a weird `Can't resolve 'child_process'` error when I import individual methods like that so I'll leave as is.

packages/explorer-2.0/components/EditProfile/index.tsx
kyriediculous: Do we have unit testing? If we extract it into a helper then we can easily unit test it.

kyriediculous: Or can we do a sanity check at least to see that this works as expected ?

adamsoffer: Not yet :/ Been meaning to set up unit testing. I can create an issue to set that up. I did a sanity test locally and then on https://beta.explorer.livepeer.org

kyriediculous: Ok , looks good then !

kyriediculous: Could we squash the second commit into the first one ?

adamsoffer: Great! Will do.

BTW in addition to unit tests, another thing I can set up to help us test new features/fixes is "deployment previews" with Now. Preview urls for every commit and PR  👉https://zeit.co/features/deployment-previews

kyriediculous (approved):

PR #580, opened 2020-03-04
fix ga page tracking

adamsoffer:
This PR fixes GA page tracking. When I refactored our GA implementation to use the `react-ga` library our page tracking stopped working.

- Adds missing line for tracking pages
- Moves GA logic out of `_app.tsx` and into `layout.tsx` so page tracking fires when route changes client-side 
- Hardcoded google tracking ID (doesn't really need to be an env variable).
- Moves now.json to the root of the project to fix Now deployment preview URLs.

Pushed to prod and confirmed GA realtime page tracking is working again.

**Checklist:**


packages/explorer-2.0/layouts/main.tsx
```diff
@@ -7,12 +7,19 @@ import Orchestrators from '../public/img/orchestrators.svg'
...
 import Header from '../components/Header'
 import Router from 'next/router'
 import useWindowSize from 'react-use/lib/useWindowSize'
 import WalletModal from '../components/WalletModal'
 import { useQuery } from '@apollo/react-hooks'
+import ReactGA from 'react-ga'
+import { isMobile } from 'react-device-detect'
+
+if (process.env.NODE_ENV === 'production') {
+  ReactGA.initialize('UA-111259858-1')
```
yondonfu: Shouldn't this be stored in a env variable? If the GA ID is exposed here couldn't anyone use it to send data to the LP GA account?

yondonfu (changes requested):

packages/explorer-2.0/layouts/main.tsx
adamsoffer: It will be publicly available whether it's an env variable or not. You can find the GA ID through the source code of any site that uses Google Analytics. I have a filter set up so that only traffic coming from our domain is permitted. For consistency, though, probably best to store this as an env variable along with everything else.

adamsoffer: added 👍

yondonfu (approved):
LGTM! Looks like the commits can be squashed into a single one.

Issue #582, opened 2020-03-05
api: email/password registration backend


iameli:
If you're working on this, ask Eli for the code behind the time he did this before on a Cloudflare worker

PR #592, opened 2020-03-11
ar/adding email password registration backend

angyangie:
First pass at https://github.com/livepeer/livepeerjs/issues/582

https://github.com/livepeer/livepeerjs/issues/582

https://github.com/livepeer/livepeerjs/issues/582

**Checklist:**


packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+import logger from '../logger'
+import Crypto from 'node-webcrypto-ossl'
+import util from 'util'
+import uuid from 'uuid/v4'
+import jwt from 'jsonwebtoken'
+import validator from 'email-validator'
+
+const app = Router()
+const crypto = new Crypto()
+const iterations = 10000
```
iameli, line 13: Maybe let's go with a all caps `ITERATIONS` to signify it's a constant

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+    }
+    const token = jwt.sign({ sub: user.id }, 'secret', {
+      algorithm: 'HS256',
+    })
+    res.status(201)
+    res.json({ id: user.id, email: user.email, token: token })
+  },
+)
+
+function makeNextHREF(req, nextCursor) {
```
iameli, line 138: Let's move this into a shared file and import it from there rather than duplicating it in all the different controllers.

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+
+  const userEmail = await req.store.get(`user-emails/${req.body.email}`)
+  if (userEmail) {
+    res.status(403)
+    res.json({ error: 'user already exists' })
+    return
+  }
+
+  const [hashedPassword, salt] = await hash(req.body.password)
+  //   TODO: if (hashedPassword.length != req.body.password.length) {
```
iameli, line 63: This can be enforced via the JSON-schema

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+  async (req, res) => {
+    const email = req.body.email
+    const password = req.body.password
+    if (!email || !password) {
+      res.status(422)
+      res.json({ error: 'missing email or password' })
+      return
+    }
+
+    const userEmail = await req.store.get(`user-emails/${email}`)
```
iameli, line 110: `const userEmail = await req.store.find("user", {email})` post-refactor

It'll do a list("user-emails/eli@iame.li"), which'll return
```
[user-emails/eli@iame.li/ABC123]
```

And then we can retrieve userId ABC123

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+    if (!userEmail) {
+      res.status(404)
+      res.json({ error: `user ${email} not found` })
+      return
+    }
+    const user = await req.store.get(`user/${userEmail.userId}`)
+    if (!user) {
```
iameli, line 117: Refactor note: this data integrity stuff can be enforced on the new model layer

Eli DevOps note: let's add a Papertrail alert for "data integrity exception"

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+      })
+      return
+    }
+    const [hashedPassword] = await hash(password, user.salt)
+    if (hashedPassword !== user.password) {
+      res.status(403)
+      res.json({ error: 'incorrect password' })
+      return
+    }
+    const token = jwt.sign({ sub: user.id }, 'secret', {
```
iameli, line 130: Let's add `--jwt-secret=` as a parameter in cli.js

iameli: Let's also at  `--jwt-audience=` as a CLI parameter and add it to the JWT here with `{sub: user.id, aud: cliParamters.jwtAudience}`

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+function makeNextHREF(req, nextCursor) {
+  let baseUrl = new URL(
+    `${req.protocol}://${req.get('host')}${req.originalUrl}`,
+  )
+  let next = baseUrl
+  next.searchParams.set('cursor', nextCursor)
+  return next.href
+}
+
+async function hash(password, salt) {
```
iameli, line 147: Let's move this function and its helpers into its own file so it can be imported by the API frontend

packages/api/src/controllers/user.test.js
```diff
@@ -0,0 +1,317 @@
...
+    password: 'mypassword',
+  }
+})
+
+afterEach(async () => {
+  await clearDatabase(server)
+})
+
+describe('controllersuser', () => {
+  describe('basic CRUD with google auth', () => {
```
iameli, line 22: Don't think we probably need to test this

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,228 @@
...
+import util from 'util'
+import uuid from 'uuid/v4'
+import jwt from 'jsonwebtoken'
+import validator from 'email-validator'
+
+const app = Router()
+const crypto = new Crypto()
+const iterations = 10000
+
+app.get('/', authMiddleware({ admin: true }), async (req, res) => {
```
iameli, line 15: Other things we might do on `authMiddleware`: 

Enforce that there is a user, but not anything else:

`authMiddleware({})`

Let's say users can't change their password with an API key, that'd be like:

`authMiddleware({ apiKey: false })`

packages/api/src/schema/schema.yaml
```diff
@@ -153,6 +157,25 @@ components:
...
+      required:
+        - email
+        - password
+      additionalProperties: false
+      properties:
+        email:
+          type: string
+          description: user email address
+          example: useremail@gmail.com
+        password:
```
iameli, line 171: We can do the `minLength` and `maxLength` enforcement here - should be exactly 64 characters

iameli (changes requested):

angyangie:
Closing this branch - all changes were included in https://github.com/livepeer/livepeerjs/pull/604

PR #594, opened 2020-03-12
add round status component

### LINKED ISSUE #537: explorer: display blocks and time estimate until next round

adamsoffer:
This PR adds a round status component to the explorer so tokenholders can view how many blocks and how much time is left in the current round, and whether it's been initialized. This component can be viewed by tapping on the little round widget on the bottom of the sidebar.

Closes #537

**Additional Context**
This component relies on data from the Livepeer SDK as opposed to the subgraph because at the moment the subgraph only indexes new rounds when they're initialized, but this component requires displaying whether a round has been initialized or not. I plan on submitting a feature request to the Graph that would allow us to create a dynamic handler that gets triggered, for example, 5760 blocks after the `currentRoundStartBlock`, so we can eventually swap out the data in this component with subgraph data.

**Screenshots (optional):**
Screen Shot 2020-03-11 at 11 27 14 PM

**Checklist:**


packages/explorer-2.0/components/RoundStatus/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+        id,
+        initialized,
+        lastInitializedRound,
+        length,
+        startBlock,
+        blockNumber: number,
+        blocksRemaining,
+        percentage,
+        blocksSinceCurrentRoundStart,
+        timeRemaining: ethGasStationResult.block_time * blocksRemaining,
```
kyriediculous, line 85: Do you know how this is optimised when compiling either by webpack or JIT browser engines?

Because this deconstruction pattern is an anti-pattern if not optimised, we're essentially assigning new memory for each key in the returned object and then using the value instead of just using the value directly, this uses double the memory space. 

I suspect this to be optimised, I just thought it might be interesting to point out. I remember reading an article whereby deconstructing in some cases lead to a 4.5x increase in bytecode size.

kyriediculous (commented):
The PR overall looks good to me.

I personally have a preference to have more condensed information on the modal rather than full sentences. I think one good option is to put more emphasis on time as a unit rather than block count. The latter can be too abstract at times to do quick math when the remaining number of blocks is still fairly large. 

Therefore I would perhaps use time indication in the circle and block numbers in the informative text area on the right.

adamsoffer:
> The PR overall looks good to me.
> 
> I personally have a preference to have more condensed information on the modal rather than full sentences. I think one good option is to put more emphasis on time as a unit rather than block count. The latter can be too abstract at times to do quick math when the remaining number of blocks is still fairly large.
> 
> Therefore I would perhaps use time indication in the circle and block numbers in the informative text area on the right.

Yeah, that's a fair point. I used blocks for the chart since it's deterministic, whereas the time estimate per round can change based on average block times. I'll look into ways we can put more emphasis on time in another design iteration.

packages/explorer-2.0/components/RoundStatus/index.tsx
adamsoffer: Interesting. Yeah reading about this [here](https://codeburst.io/es6s-function-destructuring-assignment-is-not-free-lunch-19caacc18137). Looks like the TLDR is this is optimized by JIT browser engines and we don't have to worry about this in most of the cases.

Issue #437, opened 2019-07-29
Add Network entity to the subgraph for displaying protocol pauses


iameli:
The Livepeer protocol can become paused in response to security issues - see for example https://forum.livepeer.org/t/protocol-paused-for-bug-fix-upgrade-5-26-18-9-00pm-edt-update-resumed-at-5-27-18-2-30pm-edt/265

This would be a good piece of information to expose in the frontend and the subgraph, perhaps through a Network entity.

yondonfu:
Closed by #595

PR #595, opened 2020-03-12
update subgraph Protocol entity with new fields

adamsoffer:
**What does this pull request do? Explain your changes.**
This PR adds a new set of fields to the Protocol entity and introduces 4 new entities.

**Specific updates**
- Adds the following fields to the Protocol entity: 
`currentRound, inflation, inflationChange, lastInitializedRound, lastRoundLengthUpdateRound, lastRoundLengthUpdateStartBlock, lockPeriod, maxEarningsClaimsRounds, numActiveTranscoders, paused, roundLength, roundLockAmount, targetBondingRate, unbondingPeriod`
- Introduces the following new entities:
`SetCurrentRewardTokens, Pause, Unpause, ParameterUpdate`

The changes haven been deployed and synced on the Livepeer Canary subgraph

#437 

**Checklist:**

### DIFF
--- a/packages/subgraph/src/mappings/livepeerToken.ts
+++ b/packages/subgraph/src/mappings/livepeerToken.ts
@@ -7,17 +7,8 @@ export function approval(event: ApprovalEvent): void {
   let owner = event.params.owner
   let amount = event.params.value
   let spender = event.params.spender
-
-  let protocol = Protocol.load('0')
-  if (protocol == null) {
-    protocol = new Protocol('0')
-  }
-
-  // Create delegator if it does not yet exist
-  let delegator = Delegator.load(owner.toHex())
-  if (delegator == null) {
-    delegator = new Delegator(owner.toHex())
-  }
+  let protocol = Protocol.load('0') || new Protocol('0')
+  let delegator = Delegator.load(owner.toHex()) || new Delegator(owner.toHex())
 
   delegator.allowance = amount
   delegator.save()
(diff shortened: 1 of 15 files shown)


packages/subgraph/src/mappings/livepeerToken.ts
```diff
@@ -1,16 +1,17 @@
...
   let owner = event.params.owner
   let amount = event.params.value
   let spender = event.params.spender
 
-  let roundsManager = getRoundsManagerInstance(dataSource.network())
-  let currentRound = roundsManager.currentRound()
+  let protocol = Protocol.load('0')
+  if (protocol == null) {
+    protocol = new Protocol('0')
+  }
```
kyriediculous: We can shorten this to  `const protocol = Protocol.load('0') || new Protocol('0')` , that goes for every file in this PR where we do this. 

```
const a = null || undefined || "foo" 
console.log(a)
// "foo"
```

kyriediculous (changes requested):
Looks good overall although my subgraph working knowledge is limited. One minor detail.

packages/subgraph/src/mappings/livepeerToken.ts
adamsoffer: @kyriediculous yeah good call. I've updated each file with this shortened syntax.

kyriediculous:
Changes look good, can we squash the commits on merge?

kyriediculous (approved):

PR #604, opened 2020-03-23
Ar/adding new db model to api

angyangie:
Adding new DB model to API

https://github.com/livepeer/livepeerjs/issues/599

Need to fix all tests, but working locally

https://github.com/livepeer/livepeerjs/issues/599

**Checklist:**


packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,89 @@
...
+})
+
+app.post('/token', validatePost('user'), async (req, res) => {
+  const userIds = await req.store.getPropertyIds(`useremail/${req.body.email}`)
+  const user = await req.store.get(`user/${userIds.data[0]}`, false)
+
+  const [hashedPassword] = await hash(req.body.password, user.salt)
+  if (hashedPassword !== user.password) {
+    res.status(403)
+    res.json({ error: 'incorrect password' })
```
iameli: Let's do `errors: ["incorrect password"]` here for consistency

packages/api/src/controllers/object-stores.js
```diff
@@ -14,38 +15,49 @@ const app = Router()
...
-    res.status(200)
-    res.json(secureOS)
-  } else {
-    res.status(200)
-    res.json({})
+  const objStoreIds = await req.store.getPropertyIds(
+    `objectstoresuserId/${userId}`,
+  )
+
+  if (objStoreIds.includes(id)) {
```
iameli, line 55: Need an `else` condition here. Actually, for multiple checks like this, I usually like to flip the condition and flatten out all the failure cases? Like...

```javascript
if (!objStoreIds.includes(id)) {
  res.status(403);
  return res.json({errors: [etc etc]})
}

if (req.user.admin !== true && req.user.id !== objStore.userId) {
  res.status(403);
  return res.json //etc...
}

res.status(200)
res.json(objStore)
```

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,89 @@
...
+  const emailValid = validator.validate(req.body.email)
+  if (!emailValid) {
+    res.status(422)
+    res.json({ errors: ['invalid email'] })
+    return
+  }
+
+  const [hashedPassword, salt] = await hash(req.body.password)
+  const id = uuid()
+  const admin = 'admin' in req.body ? req.body.admin : false
```
iameli: Hmm, this would let anyone creating an account declare themselves to be an admin. Let's have all users set to `false` for now, and we can manually change folks in the database for now.

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,89 @@
...
+    res.json(user)
+  } else {
+    res.status(403)
+    res.json({ errors: ['user not created'] })
+  }
+})
+
+app.post('/token', validatePost('user'), async (req, res) => {
+  const userIds = await req.store.getPropertyIds(`useremail/${req.body.email}`)
+  console.log(`USERUDSSSS: ${JSON.stringify(userIds)}`)
```
iameli: Pronounced "useroods", perhaps?

packages/api/src/controllers/user.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const user = await req.store.get(`user/${id}`)
+
+  if (user) {
+    res.status(201)
+    res.json(user)
+  } else {
+    res.status(403)
+    res.json({ errors: ['user not created'] })
+  }
```
iameli: Same as above, we can flatten out the logic here:

```suggestion
  if (!user) {
    res.status(403)
    return res.json({ errors: ['user not created'] })
  }
  
  res.status(201)
  res.json(user)
```

packages/api/src/controllers/user.test.js
```diff
@@ -0,0 +1,317 @@
...
+      const deleted = await server.store.get(`user/${resGet.id}`)
+      expect(deleted).toBe(null)
+
+      // TO DO: test for deletion of `user-email` object as well
+      // it should return a NotFound Error when trying to delete a record that doesn't exist
+      try {
+        await server.store.deleteKey(`user/${userRes.id}`)
+      } catch (err) {
+        expect(err.status).toBe(404)
+      }
```
iameli: This pattern will cause a false positive when the `deleteKey` _doesn't_ return an error —  the `catch` block will never run and nothing will ever be `expect`ed. An alternative:

```suggestion
      let error;
      try {
        await server.store.deleteKey(`user/${userRes.id}`)
      } catch (err) {
        error = err
      }
      expect(error.status).toBe(404)
```

packages/api/src/controllers/user.test.js
```diff
@@ -0,0 +1,317 @@
...
+
+      // it should return a 500 Error when trying to use replace method
+      try {
```
iameli, line 181: same error pattern

packages/api/src/store/model.js
```diff
@@ -0,0 +1,191 @@
...
+    const record = await this.get(key)
+    if (!record) {
+      throw new NotFoundError()
+    }
+    return await this.backend.delete(key)
+  }
+
+  async delete(id) {
+    const [properties, kind] = this.getSchema(id)
+    if (!properties || properties.length) {
```
iameli: Does this ever happen? Seems like we'd want everything to have properties.

iameli (changes requested):
Sweet!

Something's making me nervous about our join keys just being concatenations of kind and field name... `useremail/eli@iame.li` means "the `user` that has the `email` field set to `eli@iame.li` but it could also mean we have something with `kind: useremail`. What if we did `user+email/eli@iame.li`? That way we can make plusses illegal in `kind` fields and we never have to worry about a collision.

packages/api/src/store/model.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  getCleanKind(kind) {
+    let cleanKind = kind.charAt(0) === '/' ? kind.substring(1) : kind
+    return cleanKind.indexOf('/') > -1
+      ? cleanKind.substr(0, cleanKind.indexOf('/'))
+      : cleanKind
+  }
+
+  cleanWriteOnlyResponses(id, responses) {
```
iameli, line 182: I like this!

packages/api/src/store/model.js
```diff
@@ -0,0 +1,191 @@
...
+  async list(prefix, cursor, limit, cleanWriteOnly = true) {
+    const responses = await this.backend.list(prefix, cursor, limit)
+
+    if (responses.data.length > 0 && cleanWriteOnly) {
+      return this.cleanWriteOnlyResponses(prefix, responses)
+    }
+    return responses
+  }
+
+  async getPropertyIds(prefix, cursor, limit, cleanWriteOnly) {
```
iameli: I think the signature here could be improved. It would be nice to not have to know about how the keys are stored internally (e.g. `user+email/eli@iame.li/ABC123`) to work with this. How about something like this:

```javascript
const [userId] = await req.store.query("user", {email: "eli@livepeer.org"})
```

And then we can parse that out into something that can be passed to listKeys, something like

```javascript
async query(kind, queryObj, cursor, limit, cleanWriteOnly) {
  const [queryKey, ...rest] = Object.keys(queryObj);
  if (rest.length > 0) {
    throw new Error("you may only query() by one key")
  }
  const queryValue = queryObj[queryKey];
  const prefix = `${kind}+${queryKey}/${queryValue}`
  const [keys] = await this.backend.listKeys(prefix, cursor, limit, cleanWriteOnly)
```

packages/api/src/store/model.js
angyangie: You are right! I just removed it. Good catch

packages/api/src/store/model.js
angyangie: This is much cleaner! I have just edited

PR #605, opened 2020-03-24
wip: mist-proxy

iameli:
This is a little proxy server that can intercept requests from Mist using the "custom API url" feature. It's designed to run alongside MistServer in a sidecar container. Right now, it does nothing, just proxying connections through to the API and the broadcasters. But, it's designed to work by intercepting incoming stream requests and combining them into a single multipart transcode. This allows for multiple rendition support and multipart return support without any changes to MistProcLivepeer.


iameli:
image

PR #609, opened 2020-03-26
www: make html, body, and __next height: 100%

iameli:
Just getting my feet wet with this particular front-end stack. Also this is kind of an old-school way to solve this — anything better come to mind?

Basically I want to avoid the case where the page is kinda small and the footer shows up right after the header.

After: 
image

I don't have a before screenshot but imagine the footer smooshed right up against the text there

### DIFF
--- a/packages/livepeer.com/www/components/Layout/index.tsx
+++ b/packages/livepeer.com/www/components/Layout/index.tsx
@@ -1,7 +1,7 @@
 import { DefaultSeo } from "next-seo";
 import Navigation from "../Navigation";
 import Footer from "../Footer";
-import { Box } from "@theme-ui/components";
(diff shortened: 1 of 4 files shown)


adamsoffer:
Nice. Yeah, I believe this is referred to as the "holy grail layout" in CSS land lol. Flexbox definitely works, but a more elegant "new school" solution would be to use CSS Grid. https://dev.to/matchboxhero10/grid-tutorial-the-holy-grail-of-layout-40ef

iameli:
I thought that Holy Gail was using floats for sidebars and stuff 😅

Googling around a bit it looks like `100vh` might be a better fit for what I'm looking for here. I'll poke around at it some more.

iameli:
closed in favor of #610

PR #610, opened 2020-03-27
www: implement api login & password reset

iameli:
@adamsoffer This is ready for an initial review. If you use the top-level `yarn run dev`, you should be able to get at everything with `http://localhost:3004/`. Thoughts:

* Is there anything I'm doing with `sx` properties that should get added to the theme instead?
* I like how `useApi()` works in the components... `const {user, logout} = useApi()` feels very nicely decoupled. I don't at all like the _implementation_ of `useApi`, where I'm adding a bunch of functions onto a `useState` object to create the context. Aside from it being generally messy, I'm pretty sure it can cause all sorts of data inconsistencies because there's no guarantee that the context doesn't change in the middle of `async` functions. Yuck. If you have a better idea/library for implementing the same pattern, I'm super open to it.
* "Forgot password" doesn't do anything yet, but we'll wire that up before going live.
* This includes a different solution for https://github.com/livepeer/livepeerjs/pull/609 using `min-height: 100vh`.
* This pattern will be very common, and should probably get its own `useLoggedIn()` hook to avoid duplication: 
```javascript
  const router = useRouter();
  const { token } = useApi();
  useEffect(() => {
    if (!token) {
      router.replace("/login");
    }
  }, [token]);
```
* I threw a little space for the error text between the form and the button, but we should probably do something more prominent than a little line of text there
* We gotta do something about nav on mobile, it's all squooshed and the logo is off the page now. Hamburger time?
* Still needs some sort of `localStorage` implementation to persist the API token across reloads


adamsoffer (approved):
This is looking solid. The sign up and user log in flow is working great after running on my machine. 🚀

One pedantic piece of feedback - can we rename the `/register` route `/signup` to match the call to action text?

 > We gotta do something about nav on mobile, it's all squooshed and the logo is off the page now. Hamburger time? 

Yeah, it's 🍔time. I can design something quick in figma.

iameli:
Sweet. I'm going to make a couple of the changes I described to the hooks API and merge to staging. Hamburgers can be another PR off of that.

Issue #601, opened 2020-03-16
polling: Add tests for the poll tallying


yondonfu:
**What's the problem? (required)**
Describe the problem or need in exactly 1 sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

Poll tallying involves summing the voting stake of orchestrators and delegators. It will be important to confirm that the poll tallying logic is implemented correctly.

**Describe a possible solution you've considered (optional)**

We should investigate adding tests for the poll tallying logic in the subgraph. This would be an e2e test of sorts since it would involve running the subgraph locally. An initial thought is that we could write a shell script that runs a local instance of ganache-cli with the protocol contracts deployed. Then, we could submit transactions that would update the state of the subgraph. Finally, we could run a few GraphQL queries for poll tally results and compare them against the expected results.

Looks like TheGraph also supports a [time traveling query](https://blocklytics.org/blog/ethereum-blocks-subgraph-made-for-time-travel/) feature. If we rely on this feature we could also run additional queries at past blocks in the tests to confirm that the poll tally results are correct even when the poll is still in progress.

yondonfu:
Closed by #611

Issue #600, opened 2020-03-16
polling: Real time tallying in subgraph


yondonfu:
[**What's the problem? (required)**
Describe the problem or need in exactly 1 sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

There are two approaches to poll tallying:
- Update the poll tally at a regular interval i.e. whenever a `NewRound` is emitted
- Update the poll tally in real-time whenever any on-chain action updates the poll tally (i.e. when a vote is cast, whenever a delegator/orchestrator's stake is updated)

Regardless of the approach to poll tallying, the subgraph will need to perform an update for a list of voters/delegators.

- If we update the poll tally at a regular interval, we will need to perform an update for each voter in a poll whenever a `NewRound` is emitted
- If we update the poll tally in real-time, we will need to update the stake of all of an orchestrator's delegators when `Reward` is emitted so that the delegator's voting stake can be fetched

An inefficient method of accomplishing these updates is to loop through a list in a single event handler. This is known to cause problems (the event handler could timeout).

**Describe a possible solution you've considered (optional)**
A clear and concise description of what you want to happen. '

We can investigate a new feature in TheGraph for dynamic data sources/handlers that would create N dynamic handlers that are fired when a parent event handler is fired instead of going through N iterations in a loop in a single event handler. 

After we investigate this new feature we can decide whether we should do poll tallying at a interval or in real-time.

yondonfu:
Closed by #611

PR #611, opened 2020-03-31
add real-time poll tallying to subgraph

adamsoffer:
This PR introduces real-time poll tallying to the subgraph as per the spec outlined in [LIP16](https://github.com/livepeer/LIPs/issues/16). 

#600 #601 

**Todos**
- [x] Complete integration test assertions
- [x] Add `PollCreator` rinkeby contract address to networks.yaml
- [x] Spin up graph-node and ganache instance as part of integration tests
- [x] Add `PollCreator` mainnet contract address to networks.yaml


kyriediculous:
CI fails because the VM it runs on does not have truffle installed. 

There's two ways to fix this 
- Add truffle to the dependencies for the project and instead of `truffle test` use `./node_modules/.bin/truffle test`
- Add a job to the circle yaml file that installs truffle globally 

I'd prefer the first option as it is the simplest

Edit: see this comment about not using truffle https://github.com/livepeer/livepeerjs/pull/611#discussion_r409202110

packages/subgraph/docker-compose.yml
```diff
@@ -16,7 +18,7 @@ services:
       postgres_pass: let-me-in
       postgres_db: graph-node
       ipfs: 'ipfs:5001'
-      ethereum: 'https://eth-mainnet.alchemyapi.io/jsonrpc/[API_KEY]'
+      ethereum: 'development:http://host.docker.internal:8545'
```
kyriediculous, line 21: Is this a permanent change or was this meant for development?

packages/subgraph/schema.graphql
```diff
@@ -179,6 +180,55 @@ enum TranscoderStatus @entity {
...
+type PollTally @entity {
+  id: ID!
+  yes: BigInt
+  no: BigInt
+}
+
+type Vote @entity {
+  "voter address + poll address"
+  id: ID!
+  "The voter that casted the vote"
```
kyriediculous: how about "vote caster" for brevity ?

packages/subgraph/schema.graphql
```diff
@@ -179,6 +180,55 @@ enum TranscoderStatus @entity {
...
+  voter: Voter!
+  "Stake weighted vote"
+  voteStake: BigInt
+  "This will be non-zero if voter is an orchestrator and any of the its delegators voted"
```
kyriediculous, line 214: Since there are other cases where this can be negative I think, ie any stake that is not bonded. How about rewording this to "stake  weight not eligible for voting"

If not rewording "any of the its delegators" -> "any of its delegators"

packages/subgraph/schema.graphql
```diff
@@ -179,6 +180,55 @@ enum TranscoderStatus @entity {
...
+  nonVoteStake: BigInt
+  choice: PollChoice
+  poll: Poll
+}
+
+type Voter @entity {
+  "The voter address"
```
kyriediculous: is writing "The voted address" instead of "voter address" a subgraph convention?

packages/subgraph/truffle.js
```diff
@@ -0,0 +1,17 @@
...
+      port: 8545,
+      network_id: '*', // Match any network id
+      gas: 6600000,
+    },
+  },
+  mocha: {
+    enableTimeouts: false,
+    before_timeout: 120000,
+  },
+}
```
kyriediculous: I don't think we need truffle for this because this package isn't set up to fully utilise truffle suite anyway. The only thing we really get right now is web3 provider injection but this is quite trivial to set up. 

We can just use vanilla mocha and replace `contract` blocks with `describe` blocks. 

Then we also don't need to run and install truffle for CI but can just run `npm run test`

kyriediculous (changes requested):

packages/subgraph/docker-compose.yml
adamsoffer: This is a permanent change. We only use the docker file for development purposes. If a developer wants to deploy the subgraph using their own archive node I mention they can change the `ethereum` value in the README.

packages/subgraph/schema.graphql
adamsoffer: Sounds good

packages/subgraph/schema.graphql
adamsoffer: hmm, I'm not aware that this can ever be negative. `voteStake` will always match a delegators pending stake, and `nonVoteStake` will always be the sum of all it's delegators voteStake that voted.

> If not rewording "any of the its delegators" -> "any of its delegators"
Will fix typo

packages/subgraph/schema.graphql
adamsoffer: lol no it's not a convention. Not sure why I prefaced with "the". I can change to "voter address".

packages/subgraph/schema.graphql
kyriediculous: Oh sorry I meant non-zero but somehow wrote 'negative'

packages/subgraph/docker-compose.yml
kyriediculous: this was previously running against a mainnet ethereum node but now we expect users  to run this docker image against their own local node? 

Could we update the `README.md` as well if that's the case?

packages/subgraph/schema.graphql
kyriediculous: Oh okay I see this is per orchestrator, description looks good then !

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -31,12 +31,6 @@ export function updatePollTallyOnReward(event: RewardEvent): void {
       let protocol = Protocol.load('0') || new Protocol('0')
       let voteId = makeVoteId(delegator.id, poll.id)
       let vote = Vote.load(voteId)
-
-      // return if voter didn't vote in this specific poll
-      if (!vote) {
```
kyriediculous: I think we should keep this, no ? 

Once an orchestrator votes we can use its stake weight that was updated by the subgraph up until then.

This should reduce the amount of updates needed for orchestrators not participating in governance.

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -77,13 +71,7 @@ export function updatePollTallyOnReward(event: RewardEvent): void {
...
 export function updatePollTallyOnBond(event: BondEvent): void {
   let voterAddress = dataSource.context().getString('voter')
-  let delegator = Delegator.load(voterAddress) as Delegator
-
-  // If voter is a delegator and shares the same delegate attached to the event
-  if (
-    delegator != null &&
-    delegator.delegate == event.params.newDelegate.toHex()
-  ) {
+  if (voterAddress == event.params.delegator.toHex()) {
```
kyriediculous: Nit: instead of nesting the entire function body in this if statement, how about adopting some go idiomatics here of always trying to end with the happy case ? 

```
if (voterAddress !== event.params.delegator.toHex()) return;
// function body ... 
```

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -157,12 +140,7 @@ export function updatePollTallyOnBond(event: BondEvent): void {
 
 export function updatePollTallyOnUnbond(event: UnbondEvent): void {
   let voterAddress = dataSource.context().getString('voter')
-  let delegator = Delegator.load(voterAddress) as Delegator
-  // If voter is a delegator and shares the same delegate attached to the event
-  if (
-    delegator != null &&
-    delegator.delegate == event.params.delegate.toHex()
-  ) {
+  if (voterAddress == event.params.delegator.toHex()) {
```
kyriediculous: Nit: instead of nesting the entire function body in this if statement, how about adopting some go idiomatics here of always trying to end with the happy case ? 

```
if (voterAddress !== event.params.delegator.toHex()) return;
// function body ... 
```

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -212,12 +185,7 @@ export function updatePollTallyOnUnbond(event: UnbondEvent): void {
 
 export function updatePollTallyOnRebond(event: RebondEvent): void {
   let voterAddress = dataSource.context().getString('voter')
-  let delegator = Delegator.load(voterAddress) as Delegator
-  // If voter is a delegator and shares the same delegate attached to the event
-  if (
-    delegator != null &&
-    delegator.delegate == event.params.delegate.toHex()
-  ) {
+  if (voterAddress == event.params.delegator.toHex()) {
```
kyriediculous: Nit: instead of nesting the entire function body in this if statement, how about adopting some go idiomatics here of always trying to end with the happy case ? 

```
if (voterAddress !== event.params.delegator.toHex()) return;
// function body ... 
```

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -93,11 +81,6 @@ export function updatePollTallyOnBond(event: BondEvent): void {
       let voteId = makeVoteId(voterAddress, pollAddress)
       let vote = Vote.load(voteId)
 
-      // return if voter didn't vote in this specific poll
```
kyriediculous: same comment as above

packages/subgraph/src/mappings/dynamic.ts
```diff
@@ -172,11 +150,6 @@ export function updatePollTallyOnUnbond(event: UnbondEvent): void {
...
```
kyriediculous: same question as above

packages/subgraph/test/integration/test.js
```diff
@@ -305,54 +305,54 @@ contract('Subgraph Integration Tests', accounts => {
...
-    await Poll.methods.no().send({ from: delegator1 })
+    await Poll.methods.vote(1).send({ from: delegator1 })
 
     // delegator 2 votes yes
-    await Poll.methods.yes().send({ from: delegator2 })
+    await Poll.methods.vote(0).send({ from: delegator2 })
 
     // delegator 3 votes no
-    await Poll.methods.no().send({ from: delegator3 })
+    await Poll.methods.vote(1).send({ from: delegator3 })
```
kyriediculous: nit: we don't have to do this serially , we can use `Promise.all()`. 

```
await Promise.all([
  Poll.methods.vote(0).send({ from: transcoder1 }),
  Poll.methods.vote(1).send({ from: delegator1 })
  Poll.methods.vote(0).send({ from: delegator2 })
  Poll.methods.vote(1).send({ from: delegator3 })
])
```

kyriediculous: Additionally how about creating an object here first that we can then re-use throughout this test

```
const voters = {}
voters[transcoder1]=0
voters[delegator1]=1
voters[delegator2]=0
voters[delegator3]=1

await Promise.all(
  function() {
    let calls = []
    for voter, choiceID in voters {
       calls.push(Poll.methods.vote(choiceID).send({ from: voter }))
     }
    return calls
  }
)
```

packages/subgraph/test/integration/test.js
```diff
@@ -305,54 +305,54 @@ contract('Subgraph Integration Tests', accounts => {
...
-          choice
+          choiceID
         }
       }`,
     })
 
     assert.equal(
-      subgraphPollData.data.transcoder1VoteChoice[0].choice,
-      'Yes',
+      subgraphPollData.data.transcoder1VoteChoice[0].choiceID,
```
kyriediculous: If we also return the voter from the subgraph query on L324, L327, L330 and L333 and we adopt the object described in the comment above then this can be shortened to 

```
for (k, vote in subgraphPollData.data) {
  assert.equal(vote[0].choiceID, voters[vote.voter], "incorrect vote choice")
}

**Review by kyriediculous — CHANGES_REQUESTED** [review]

File: packages/subgraph/schema.graphql [hunk 14]
```diff
@@ -179,6 +180,55 @@ enum TranscoderStatus @entity {
   Registered
 }
 
+"""
+The possible poll choice
+"""
+enum PollChoice @entity {
```
kyriediculous, line 186: the CLI has an "Abstain" option. 

Should we add it here as well ? This has been a discussion point in LIP16. 

Personally I think it won't be used much if at all , but then again it doesn't hurt adding it for completeness. 

I'm wondering what @yondonfu thinks about this.

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -51,15 +51,12 @@ export function transcoderUpdated(event: TranscoderUpdateEvent): void {
...
-  let pendingPricePerSegment = event.params.pendingPricePerSegment
 
   // Update transcoder
   transcoder.delegator = transcoderAddress.toHex()
-  transcoder.pendingRewardCut = pendingRewardCut
-  transcoder.pendingFeeShare = pendingFeeShare
-  transcoder.pendingPricePerSegment = pendingPricePerSegment
+  transcoder.pendingRewardCut = event.params.pendingRewardCut
+  transcoder.pendingFeeShare = event.params.pendingFeeShare
+  transcoder.pendingPricePerSegment = event.params.pendingPricePerSegment
```
kyriediculous, line 59: Why are we re-adding setting this state considering these fields are deprecated?

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -78,8 +75,8 @@ export function transcoderUpdated(event: TranscoderUpdateEvent): void {
   transcoderUpdated.to = event.transaction.to.toHex()
   transcoderUpdated.round = protocol.currentRound
   transcoderUpdated.delegate = transcoderAddress.toHex()
-  transcoderUpdated.rewardCut = pendingRewardCut
-  transcoderUpdated.feeShare = pendingFeeShare
+  transcoderUpdated.rewardCut = event.params.pendingRewardCut
+  transcoderUpdated.feeShare = event.params.pendingFeeShare
```
kyriediculous, line 79: Why are we re-adding setting this state considering these fields are deprecated?

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -238,8 +239,7 @@ export function bond(call: BondCall): void {
     transcoder.totalStake = delegateData.value3
     delegate.delegatedAmount = delegateData.value3
 
-    // no existing delegate && has bondedAmount == rebonding
-    // Subtract bondedAmount from unbonded
+    // no existing delegate && delegator has bondedAmount then it is rebonding
```
kyriediculous: can we just make this comment  `// delegator rebonding` , I don't think we need to repeat the if statement expressed below in the comment. Let's be brief !

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -335,9 +338,12 @@ export function reward(event: RewardEvent): void {
   let delegate = Delegator.load(transcoderAddress.toHex())
   let delegateData = bondingManager.getDelegator(transcoderAddress)
   let protocol = Protocol.load('0') || new Protocol('0')
-  let poolId = makePoolId(transcoderAddress, protocol.currentRound)
+  let totalActiveStake = bondingManager.getTotalBonded()
```
kyriediculous: we can just assign thi return values straight to `protocol.totalActiveStake`  at L345  and remove this line I believe ?

packages/subgraph/src/mappings/bondingManager_LIP11.ts
```diff
@@ -33,25 +37,22 @@ export function bond(event: BondEvent): void {
...
-
   let delegate =
     Delegator.load(newDelegateAddress.toHex()) ||
     new Delegator(newDelegateAddress.toHex())
-
   let delegator =
     Delegator.load(delegatorAddress.toHex()) ||
     new Delegator(delegatorAddress.toHex())
 
+  protocol.totalActiveStake = totalActiveStake
```
kyriediculous: same comment as above about just assigning directly instead of first declaring `totalActiveStake`

packages/subgraph/src/mappings/bondingManager_LIP11.ts
```diff
@@ -220,8 +221,10 @@ export function rebond(event: RebondEvent): void {
   let delegate = Delegator.load(delegateAddress.toHex())
   let delegator = Delegator.load(delegatorAddress.toHex())
   let delegateData = bondingManager.getDelegator(delegateAddress)
-
   let protocol = Protocol.load('0') || new Protocol('0')
+  let totalActiveStake = bondingManager.getTotalBonded()
+
+  protocol.totalActiveStake = totalActiveStake
```
kyriediculous: same comment as above about just assigning directly instead of first declaring totalActiveStake

packages/subgraph/src/mappings/bondingManager_LIP11.ts
```diff
@@ -136,21 +137,20 @@ export function unbond(event: UnbondEvent): void {
...
-
   let delegate =
     Delegator.load(delegateAddress.toHex()) ||
     new Delegator(delegateAddress.toHex())
-
   let unbondingLock =
     UnbondingLock.load(uniqueUnbondingLockId) ||
     new UnbondingLock(uniqueUnbondingLockId)
 
+  protocol.totalActiveStake = totalActiveStake
```
kyriediculous: same comment as above about just assigning directly instead of first declaring totalActiveStake

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function vote(event: VoteEvent): void {
+  let protocol = Protocol.load('0') || new Protocol('0')
+  let delegator = Delegator.load(event.params.voter.toHex())
+  let voter =
+    Voter.load(event.params.voter.toHex()) ||
+    new Voter(event.params.voter.toHex())
+  let poll = Poll.load(event.address.toHex()) as Poll
+  let voteId = makeVoteId(event.params.voter.toHex(), poll.id)
+  let vote = Vote.load(voteId) || new Vote(voteId)
```
kyriediculous: In the case a user votes twice on the same poll , could we remove the old vote from the tally and just create a new vote instead of mutating the old one? That should simplify/shorten this function a little bit. I'm not sure if this is actually possible though , eg `vote.remove()` ?

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function tallyVotes(poll: Poll): void {
+  let pollTally = new PollTally(poll.id)
+  let votes = poll.votes as Array<string>
+  let v: Vote
+  pollTally.yes = BigInt.fromI32(0)
+  pollTally.no = BigInt.fromI32(0)
+  for (let i = 0; i < votes.length; i++) {
+    v = Vote.load(votes[i]) as Vote
+    if (v.choiceID == 0) {
```
kyriediculous: WDYT about turning this into a `switch` statement?

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+
+import { makeVoteId, EMPTY_ADDRESS } from '../../utils/helpers'
+import { tallyVotes } from './poll'
+
+export function updatePollTallyOnReward(event: RewardEvent): void {
+  let voterAddress = dataSource.context().getString('voter')
+  let delegator = Delegator.load(voterAddress) as Delegator
+
+  // If voter is a delegator and the caller that invoked reward is its delegate,
+  if (
```
kyriediculous, line 22: What do you think about reversing this `if`statement to check for the bad case and `return`  so we don't have to encapsulate the entire function body in the `if` statement

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+    delegator != null &&
+    delegator.delegate == event.params.transcoder.toHex()
+  ) {
+    let pollAddress = dataSource.context().getString('poll')
+    let poll = Poll.load(pollAddress) as Poll
+
+    // If poll is active then update delegator's vote stakes and poll tally
+    if (poll.endBlock > event.block.number) {
+      let protocol = Protocol.load('0') || new Protocol('0')
```
kyriediculous: what do you think about moving this top-level and doing 

```
if (poll.endBlock <= event.block.number) return;
```

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+      let voteId = makeVoteId(delegator.id, poll.id)
+      let vote = Vote.load(voteId)
+      let transcoder = Transcoder.load(event.params.transcoder.toHex())
+
+      // update vote stakes
+      if (delegator.id == event.params.transcoder.toHex()) {
+        vote.voteStake = transcoder.totalStake
+      } else {
```
kyriediculous: if the changes above are implemented we can unnest  this as well and do 

```
if (delegator.id !== event.params.transcoder.toHex()) return;
// rest of function body here 
```

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+        vote.voteStake = pendingStake
+      }
+
+      vote.save()
+      tallyVotes(poll)
+    }
+  }
+}
+
+export function updatePollTallyOnBond(event: BondEvent): void {
```
kyriediculous, line 69: see comment above about preventing `if else` nesting by first checking the bad case and possibly returning, which allows the happy case of the function to be unnested, improving readability. We can apply the same logic here.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export function updatePollTallyOnUnbond(event: UnbondEvent): void {
+  let voterAddress = dataSource.context().getString('voter')
+  if (voterAddress == event.params.delegator.toHex()) {
+    let pollAddress = dataSource.context().getString('poll')
+    let poll = Poll.load(pollAddress) as Poll
+
+    // Check if poll is active
+    if (poll.endBlock > event.block.number) {
```
kyriediculous: see comment above about preventing if else nesting by first checking the bad case and possibly returning, which allows the happy case of the function to be unnested, improving readability. We can apply the same logic here.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+      }
+      vote.save()
+      tallyVotes(poll)
+    }
+  }
+}
+
+export function updatePollTallyOnRebond(event: RebondEvent): void {
+  let voterAddress = dataSource.context().getString('voter')
+  if (voterAddress == event.params.delegator.toHex()) {
```
kyriediculous: see comment above about preventing if else nesting by first checking the bad case and possibly returning, which allows the happy case of the function to be unnested, improving readability. We can apply the same logic here.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+    let pollAddress = dataSource.context().getString('poll')
+    let poll = Poll.load(pollAddress) as Poll
+
+    // Check if poll is active
+    if (poll != null && poll.endBlock > event.block.number) {
```
kyriediculous: see comment above about preventing if else nesting by first checking the bad case and possibly returning, which allows the happy case of the function to be unnested, improving readability. We can apply the same logic here.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,286 @@
...
+  // If voter is a delegator and it shares the same delegate attached to the event
+  if (
+    delegator != null &&
+    delegator.delegate == event.params.delegate.toHex()
+  ) {
+    let pollAddress = dataSource.context().getString('poll')
+    let poll = Poll.load(pollAddress) as Poll
+
+    // Check if poll is active
+    if (poll != null && poll.endBlock > event.block.number) {
```
kyriediculous: see comment above about preventing if else nesting by first checking the bad case and possibly returning, which allows the happy case of the function to be unnested, improving readability. We can apply the same logic here.

packages/subgraph/src/mappings/roundsManager.ts
```diff
@@ -87,6 +90,7 @@ export function newRound(event: NewRoundEvent): void {
   let protocol = Protocol.load('0') || new Protocol('0')
   protocol.lastInitializedRound = roundsManager.lastInitializedRound()
   protocol.currentRound = roundNumber.toString()
+  protocol.totalActiveStake = totalActiveStake
```
kyriediculous: let's assign this directly instead of first declaring `totalActiveStake`

packages/subgraph/src/mappings/roundsManager_streamflow.ts
```diff
@@ -67,6 +74,7 @@ export function newRound(event: NewRoundEvent): void {
   let protocol = Protocol.load('0') || new Protocol('0')
   protocol.lastInitializedRound = roundsManager.lastInitializedRound()
   protocol.currentRound = roundNumber.toString()
+  protocol.totalActiveStake = totalActiveStake
```
kyriediculous: let's assign this directly instead of first declaring totalActiveStake

kyriediculous (changes requested):

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: ah yeah this file is outdated. Sorry. I renamed this file to `pollTallyHandlers.ts` where I make this check

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: This is also outdated and has been changed.

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: This is also outdated and has been changed. See `pollTallyHandlers.ts`

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: outdated

packages/subgraph/test/integration/test.js
adamsoffer: Nice yeah I like that.

packages/subgraph/test/integration/test.js
adamsoffer: Nice will do

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: I removed these fields in a previous commit because they're deprecated, then realized I'll need them in the future to calculate historical reward shares.

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: same as above

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: totally

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: yeah let's do that

packages/subgraph/patches/@graphprotocol+graph-cli+0.18.0.patch
```diff
@@ -1,11 +1,11 @@
...
 --- a/node_modules/@graphprotocol/graph-cli/src/codegen/schema.js
 +++ b/node_modules/@graphprotocol/graph-cli/src/codegen/schema.js
-@@ -129,7 +129,7 @@ module.exports = class SchemaCodeGenerator {
-                             'value',
-                             fieldValueType
-                           )}`
+@@ -127,7 +127,7 @@ module.exports = class SchemaCodeGenerator {
+ 
+ 
+     let getNonNullable = `return ${typesCodegen.valueToAsc('value', fieldValueType)}`
```
yondonfu, line 8: We have this patch here because it hasn't landed in the latest release of graph-cli?

packages/subgraph/schema.graphql
```diff
@@ -179,6 +180,48 @@ enum TranscoderStatus @entity {
...
```
yondonfu, line 186: I know that the details of how the enum is represented internally (as integers for example) are not leaked to the client, but might be good for consistency to put `Yes` first to match the description in the LIP-16 spec and the ordering in [livepeer_cli](https://github.com/livepeer/go-livepeer/blob/acea1c17be1cc319e808d1bf70f58bf6de8adc48/eth/types/contracts.go#L104).

packages/subgraph/networks.yaml
```diff
@@ -27,6 +27,9 @@ mainnet:
     ticketBroker:
       address: '5b1ce829384eebfa30286f12d1e7a695ca45f5d2'
       startBlock: 9274420
+    pollCreator:
```
yondonfu, line 30: Adding this note as a reminder to myself that this is a placeholder that needs to be updated once the contract is deployed on mainnet.

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -182,17 +179,21 @@ export function bond(call: BondCall): void {
     let amount = call.inputs._amount
     let delegatorData = bondingManager.getDelegator(delegatorAddress)
     let delegateData = bondingManager.getDelegator(newDelegateAddress)
+    let totalActiveStake = bondingManager.getTotalBonded()
```
yondonfu: The value returned by `bondingManager.getTotalBonded()` will only be updated when a new round is initialized. The BondingManager will keep track of stake that is pending activation and deactivation during a round and when a new round is initialized it will update the active stake for the new round to take into account stake that was pending activation or deactivation in the previous round. So, I think we only need to update the `totalActiveStake` field in the Protocol entity when a new round is initialized.

packages/subgraph/src/mappings/pollCreator.ts
```diff
@@ -0,0 +1,37 @@
+import { PollCreated as PollCreatedEvent } from '../types/PollCreator/PollCreator'
+import { Protocol, Poll, PollCreated } from '../types/schema'
+import { makeEventId } from '../../utils/helpers'
+import { Poll as PollTemplate } from '../types/templates'
+
+export function pollCreated(event: PollCreatedEvent): void {
+  let poll = new Poll(event.params.poll.toHex())
+  poll.tally = event.params.poll.toHex()
```
yondonfu, line 8: Just to clarify - we should just be setting `tally` to the poll address because the ID of the PollTally instance will be the poll address?

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,119 @@
...
+          event.params.voter,
+          BigInt.fromI32(protocol.currentRound as i32),
+        )
+        vote.voteStake = pendingStake
+      }
+
+      // If delegate is a registered orchestrator and not delegated to self
+      if (
+        delegate.status == 'Registered' &&
+        event.params.voter.toHex() != delegator.delegate
```
yondonfu: `delegator.delegate` -> `delegate` since `delegate` is already declared?

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,119 @@
...
+    vote.choiceID = 'Yes'
+  } else if (event.params.choiceID.equals(BigInt.fromI32(1))) {
+    vote.choiceID = 'No'
+  } else {
+    return
+  }
+
+  // If first time voting in this poll
+  let pollVotes = poll.votes ? poll.votes : new Array<string>()
+  if (pollVotes.indexOf(voteId) == -1) {
```
yondonfu: IIUC we are not just checking if there is already a Vote saved here i.e. `Vote.load(voteId) == null` because if a delegator votes then a Vote will be saved for its orchestrator even if the orchestrator itself has not voted yet. This `indexOf` call will get more expensive as the # of voters increases which likely won't be an issue at this stage, but I wonder if there is an alternative that doesn't involve `indexOf`.

Since choiceID is nullable, could we check if the old choiceID (before updating it starting on L24) was null? If the old choiceID was null then this would be the first time the voter is voting in the poll.

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,119 @@
...
+export function vote(event: VoteEvent): void {
+  let protocol = Protocol.load('0') || new Protocol('0')
+  let delegator = Delegator.load(event.params.voter.toHex())
+  let poll = Poll.load(event.address.toHex()) as Poll
+  let voteId = makeVoteId(event.params.voter.toHex(), poll.id)
+  let vote = Vote.load(voteId) || new Vote(voteId)
+  let delegate: Transcoder
+
+  if (event.params.choiceID.equals(BigInt.fromI32(0))) {
+    vote.choiceID = 'Yes'
```
yondonfu, line 38: Could we use the `PollChoice` enum here? i.e. `vote.choiceID = PollChoice.Yes`

packages/subgraph/subgraph.template.yaml
```diff
@@ -58,21 +56,21 @@ dataSources:
           handler: withdrawFees
         - event: ParameterUpdate(string)
           handler: parameterUpdate
-      callHandlers:
-        - function: claimEarnings(uint256)
-          handler: claimEarnings
-        - function: bond(uint256,address)
-          handler: bond
+      # callHandlers:
```
yondonfu: Don't we need these call handlers for the period of time on mainnet when the necessary data wasn't available in events yet?

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+        let delegateVoteId = makeVoteId(
+          event.params.transcoder.toHex(),
+          poll.id,
+        )
+        let delegateVote = Vote.load(delegateVoteId) || new Vote(delegateVoteId)
+        delegateVote.voter = event.params.transcoder.toHex()
+
+        // if delegate voted, update its vote stake
+        if (delegateVote.voteStake != null) {
+          delegateVote.voteStake = transcoder.totalStake
```
yondonfu: Isn't this case handled already?

- If a delegate has voted already, then this handler will fire for the delegate and L38 will update its vote stake
- If a delegator has not voted yet, then its vote stake will be updated by the vote handler when the delegate casts a vote

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+        let delegateVoteId = makeVoteId(
+          event.params.newDelegate.toHex(),
+          poll.id,
+        )
+        let delegateVote = Vote.load(delegateVoteId) || new Vote(delegateVoteId)
+        delegateVote.voter = event.params.newDelegate.toHex()
+
+        // if delegate voted, update its vote stake
+        if (delegateVote.voteStake != null) {
+          delegateVote.voteStake = transcoder.totalStake
```
yondonfu: Isn't this case handled already?

- If a delegate has voted already, then this handler will fire for the delegate and L06 will update its vote stake
- If a delegator has not voted yet, then its vote stake will be updated by the vote handler when the delegate casts a vote

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+          BigInt.fromI32(protocol.currentRound as i32),
+        )
+
+        let delegateVoteId = makeVoteId(event.params.delegate.toHex(), poll.id)
+        let delegateVote = Vote.load(delegateVoteId) || new Vote(delegateVoteId)
+        delegateVote.voter = event.params.delegate.toHex()
+
+        // if delegate voted, update its vote stake
+        if (delegateVote.voteStake != null) {
+          delegateVote.voteStake = transcoder.totalStake
```
yondonfu: Ditto other comments from other handlers about already handling this case.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+    // Check if poll is active
+    if (poll.endBlock > event.block.number) {
+      let transcoder = Transcoder.load(event.params.delegate.toHex())
+      let voteId = makeVoteId(voterAddress, pollAddress)
+      let vote = Vote.load(voteId)
+
+      if (event.params.delegator.toHex() == event.params.delegate.toHex()) {
+        vote.voteStake = transcoder.totalStake
+      } else {
+        let protocol = Protocol.load('0') || new Protocol('0')
```
yondonfu: A lot of the logic for this case seems to be the same across all the handlers in this file i.e. call pendingStake, update the delegate's non vote stake, update the delegator's vote stake. Perhaps we can extract this into a helper?

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+}
+
+export function updatePollTallyOnRebond(event: RebondEvent): void {
+  let voterAddress = dataSource.context().getString('voter')
+  if (voterAddress == event.params.delegator.toHex()) {
+    let pollAddress = dataSource.context().getString('poll')
+    let poll = Poll.load(pollAddress) as Poll
+
+    // Check if poll is active
+    if (poll != null && poll.endBlock > event.block.number) {
```
yondonfu: Would the poll ever be null since this handler is only setup if the voter voted in the poll already? Also don't see this check in the other handlers.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,291 @@
...
+      let transcoder = Transcoder.load(event.params.delegate.toHex())
+      let voteId = makeVoteId(voterAddress, pollAddress)
+      let vote = Vote.load(voteId)
+
+      // return if voter didn't vote in this specific poll
+      if (!vote) {
```
yondonfu: Wouldn't this always be non-null since this handler is only setup if the voter voted?

packages/subgraph/test/integration/test.js
```diff
@@ -0,0 +1,429 @@
...
+    delegator2StartStake = 3000
+    delegator3StartStake = 3000
+
+    // Register transcoder 1
+    await Token.methods
+      .approve(bondingManagerAddress, transcoder1StartStake)
+      .send({ from: transcoder1 })
+    await BondingManager.methods
+      .bond(transcoder1StartStake, transcoder1)
+      .send({ gas: 1000000, from: transcoder1 })
```
yondonfu: Looks like web3.js allows you to set a default gas limit for all txs sent for a contract instance ([docs](https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#options)). That could help avoid having to set the gas limit for all the txs in this test file.

yondonfu (changes requested):

packages/subgraph/src/mappings/dynamic.ts
kyriediculous: It seems like the if-else structure persisted across renaming but I left similar comments on the updated version as well. Note that this won't actually change the logic just code readability so it's really nitpicking and might be personal opinion as well so I'll leave it up to you whether you'd like to address the code structure or not.

packages/subgraph/docker-compose.yml
adamsoffer: Updated README 👍

packages/subgraph/schema.graphql
adamsoffer: Changed to vote caster in latest

packages/subgraph/schema.graphql
adamsoffer: Updated all comments and removed "the" prefixes for brevity

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: I've adopted this go idiomatic across the board in latest commit 👍

packages/subgraph/src/mappings/dynamic.ts
adamsoffer: done

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: Changed in latest 👍

packages/subgraph/src/mappings/bondingManager_LIP11.ts
adamsoffer: Done. Also, per Yondon this only needs to get set on NewRound events, so I removed from this handler altogether.

packages/subgraph/src/mappings/bondingManager_LIP11.ts
adamsoffer: Same as above.

packages/subgraph/src/mappings/bondingManager_LIP11.ts
adamsoffer: Same comment as above.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: I dig it. I've adopted this go idiomatic throughout.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/src/mappings/roundsManager.ts
adamsoffer: done

packages/subgraph/src/mappings/roundsManager_streamflow.ts
adamsoffer: done

packages/subgraph/patches/@graphprotocol+graph-cli+0.18.0.patch
adamsoffer: Yeah exactly. I need to follow up with Leo from the Graph about landing a fix for this in `graph-cli` so we can remove this patch.

packages/subgraph/schema.graphql
adamsoffer: Done

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: Makes sense. I've removed it from all the handlers with the exception of `newRound`.

packages/subgraph/src/mappings/pollCreator.ts
adamsoffer: Yeah exactly.

packages/subgraph/src/mappings/poll.ts
adamsoffer: Updated to use `delegate.id`.

packages/subgraph/src/mappings/poll.ts
adamsoffer: Yeah good call that would be more performant. I pushed up an update that instead checks if the old `choiceID` is null.

packages/subgraph/src/mappings/poll.ts
adamsoffer: For some reason the autogenerated exported types don't include enums, so I have to hard code 🤷‍♂️

packages/subgraph/subgraph.template.yaml
adamsoffer: Yeah we do. I temporarily commented them out because ganache and rinkeby don't support them. I just pushed up an update that conditionally outputs the callHandlers in the subgraph.yaml manifest based on the network name.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Good call. Removed.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: ditto. removed.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: ditto. removed.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Yeah good point. I pushed up an update that abstracts this logic into its own function called `updatePollTally` used by the `updatePollTallyOnUnbond` and `updatePollTallyOnRebond`.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Yeah you're right the poll should never be null. Removed.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Yup. Removed this check in latest.

packages/subgraph/src/mappings/poll.ts
adamsoffer: I found that switch case statements don't compile with the assembly script compiler for some reason 🤷‍♂️

packages/subgraph/test/integration/test.js
adamsoffer: done

packages/subgraph/test/integration/test.js
adamsoffer: done

packages/subgraph/src/mappings/poll.ts
adamsoffer: We could but not sure it would shorten the function all that much. The only thing we change at this point if a user votes a second time is the votes choiceID.

packages/subgraph/test/integration/test.js
```diff
@@ -38,6 +38,8 @@ const fetchSubgraph = createApolloFetch({
   uri: 'http://127.0.0.1:8000/subgraphs/name/livepeer/livepeer',
 })
 
+const defaults = { gas: 1000000 }
```
kyriediculous: Can we use this value as a fallback as per [this comment](https://github.com/livepeer/livepeerjs/pull/611#discussion_r418158455) 

That way we don't have to deconstruct `...defaults`  when we make a `send` we can just overwrite the default `from` address.  

___________

We can add this default to the `web3.eth.Contract()` call 

`web3.eth.Contract(ABI, address, options)`  where `options` is the object of transaction options of which some will be used as fallbacks for calls/sends 

https://web3js.readthedocs.io/en/v1.2.5/web3-eth-contract.html#new-contract

kyriediculous (changes requested):
Changes look good so far :+1:  , one minor comment regarding setting a default `gas` value on the contract instance when  we create it.

packages/subgraph/src/mappings/poll.ts
```diff
@@ -0,0 +1,125 @@
...
+        )
+        let bondingManager = BondingManager.bind(
+          Address.fromString(bondingManagerAddress),
+        )
+        let protocol = Protocol.load('0') || new Protocol('0')
+        let pendingStake = bondingManager.pendingStake(
+          event.params.voter,
+          BigInt.fromI32(protocol.currentRound as i32),
+        )
+        vote.voteStake = pendingStake
```
yondonfu, line 74: I think the code in this else block will give vote stake to a delegator even if its delegate is not a registered orchestrator. I think this code only be executed if the delegate is a registered orchestrator (along with the code starting at L75)?

kyriediculous:
Looks like CI Is failing because it can't find a remote ethereum node. 

I think we need an additional job in CircleCI that spins up the ganache with protocol image before we run the tests.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,276 @@
...
+  if (delegateVote != null && !!delegateVote.choiceID) {
+    updateTally = true
+    let newDelegate = Transcoder.load(event.params.delegate.toHex())
+    delegateVote.voteStake = newDelegate.totalStake
+    delegateVote.save()
+  }
+
+  // if delegator voted, update its voteStake
+  let voterAddress = dataSource.context().getString('voter')
+  if (voterAddress == event.params.delegator.toHex()) {
```
yondonfu: If the voter is a registered orchestrator I think the code starting on L234 will run and then the code in this block will run overwriting the vote stake with the pending stake of the registered orchestrator as opposed to its total stake?

yondonfu (changes requested):

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: There's a check on L248 that only runs if the voter isn't delegated to itself (a registered O). We can probably combine the conditionals on L241 and 248 to make this more clear.

yondonfu: Ah woops missed that check. Combining the conditionals sounds good.

packages/subgraph/src/mappings/poll.ts
adamsoffer: Do we only want to count stake with registered orchestrators? I'm actually not sure if that explicitly mentioned in LIP-16. We make the distinction between inactive and active stake, but is stake with an unregistered delegate considered inactive?

adamsoffer: My assumption was only registered orchestrators, in which case this needs to be addressed, but just want to double check!

yondonfu: Hm LIP-16 actually just states that stake (both inactive and active) receives voting power. Under that definition, all stake, regardless of whether it is delegated to a registered orchestrator or not receives voting power which seems fine to me then. Nvm!

adamsoffer:
@yondonfu hooray all checks have passed ✅. I believe this is ready to merge once we have the mainnet contract address. Should we deploy latest rinkeby contract with the state variable name change (`threshold` to `quota`) and resync subgraph? Then I can update the graphql query in the UI with the new field name.

yondonfu:
> Should we deploy latest rinkeby contract with the state variable name change (threshold to quota) and rsync subgraph?

Deployed at [0x149805EF90FaDA12D27e8a020b6c138F3d86A9a3](https://rinkeby.etherscan.io/address/0x149805ef90fada12d27e8a020b6c138f3d86a9a3).

packages/subgraph/test/integration/test.js
adamsoffer: done!

kyriediculous:
Thanks for fixing up CI @yondonfu !

packages/subgraph/test/integration/test.js
adamsoffer: done!

packages/subgraph/truffle.js
adamsoffer: I noticed the graph uses truffle in their integration tests which is what I based this on. https://github.com/graphprotocol/graph-node/tree/master/tests/integration-tests

I think using truffle could come in handy at some point and since it's already working probably best to leave as is at this point.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: done

packages/subgraph/truffle.js
kyriediculous: I'm fine leaving it as is although I find it a weird decision on their part. 

Truffle really shines as a smart contract development suite but is very opinionated about things like directory structure and to fully leverage truffle your deployed addresses need to be in the ABI files. 

Right now it just injects web3 and auto-connects the provider to the right network through the truffle config file. Since we just use a single network to run the tests that's a lot of package overhead. 

The main reason I'm fine with it is that it's a dev dependency and not included in the production build. Perhaps interesting features could come of it in the future, we'll see. It doesn't hurt keeping it either.

adamsoffer: yeah that makes sense. I'm down to remove in a follow-up PR.

adamsoffer:
Heads up I’ve updated and deployed subgraph + UI with latest Rinkeby PollCreator contract and created a new test poll.

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -0,0 +1,302 @@
...
+      }
+    }
+
+    newDelegateVote.voter = event.params.newDelegate.toHex()
+    newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
+      .minus(vote.voteStake as BigInt)
+      .plus(event.params.bondedAmount)
+    newDelegateVote.save()
+
+    vote.voteStake = pendingStake
```
yondonfu: I think pendingStake == event.params.bondedAmount here so we could just use event.params.bondedAmount and remove the pendingStake variable/function call?

packages/subgraph/test/integration/test.js
```diff
@@ -0,0 +1,622 @@
...
+  })
+
+  it('correctly tallies poll after delegator with zero voting power votes', async () => {
+    let subgraphPollData = await fetchSubgraph({
+      query: `{ polls { id } }`,
+    })
+    const pollAddress = subgraphPollData.data.polls[0].id
+    const Poll = new web3.eth.Contract(PollABI, pollAddress, defaults)
+
+    Poll.methods.vote(1).send({ from: delegator6 })
```
yondonfu: Missing an `await` here

yondonfu: Proposed change in #689

packages/subgraph/src/mappings/pollTallyHandlers.ts
yondonfu: Proposed change in #689

adamsoffer: ah nice one.

adamsoffer:
I've updated the subgraph manifest with the PollCreator mainnet address and startblock and deployed to our mainnet "canary" subgraph. It's currently syncing and should be up to date in a few hours or so. https://thegraph.com/explorer/subgraph/livepeer/livepeer-canary?version=pending

yondonfu (approved):
LGTM :shipit: after resolving conflicts with master

PR #636, opened 2020-04-18
explorer: batchClaimEarnings gas optimisations

### LINKED ISSUE #597: Pre-fill claimRewards transaction object with gas price
Metamask sometimes incorrectly calculates the gasPrice for `claimRewards` causing very high gas being sent with the transaction. In [this](https://github.com/livepeer/protocol/issues/372) case even the block gas limit 

Metamask calculates the gas needed

We can prefill the transaction object being sent to the web3 provider with a gas amount 

The gas cost for `claimRewards` should be linear with the amount of rounds to claim for, more specifically `O(1 + n)`

so if we can figure out the gas cost for 1 round we can multiply it by the number of rounds to claim through. We already apply an overhead because the amount we multiply includes the cost to invoke a transaction as well as the constant gas cost for the execution up until the loop in `claimRewards`. 

https://github.com/livepeer/protocol/issues/372

kyriediculous:
This PR optimises gas usage for `batchClaimEarnings`. 

User reports have stated that when using metamask it would often set a very high tx gas limit or even default to the block gas limit. 

- 3a3d415 - Add a gas estimation by doing an `eth_estimategas` RPC call before sending the transaction. This should prevent metamask from defaulting to high values.
- 9b5c485 - Change `MAX_EARNINGS_CLAIMS_ROUNDS` to `50`
  This will increase the number of transactions to be approved in metamask if there's a high number of rounds to claim for but it makes a tradeoff to better play into miner economics for gas price optimisation. Transactions with high gas limits (> 10%-20% of the block gas limit) are less likely to be included in blocks if they don't have a higher gas prices attached to them than what the average for a bunch of smaller transactions would be. However people are more inclined to commit to a higher gas price for smaller transactions to get faster confirmations. Leaving large , low gas price transactions in the memory pool for a long time.  

Fixes #597

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/explorer-2.0/apollo/resolvers/Mutation.tsx
+++ b/packages/explorer-2.0/apollo/resolvers/Mutation.tsx
@@ -1,4 +1,4 @@
-import { MAX_EARNINGS_CLAIMS_ROUNDS } from '../../lib/utils'
+import { MAX_BATCH_CLAIM_ROUNDS } from '../../lib/utils'
 
 /**
  * Approve an amount for an ERC20 token transfer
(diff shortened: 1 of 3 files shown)


packages/explorer-2.0/apollo/resolvers/Mutation.tsx
```diff
@@ -68,6 +68,13 @@ export async function batchClaimEarnings(_obj, _args, _ctx) {
     calls.push(
       bondingManager.methods.claimEarnings(
         (parseInt(lastClaimRound) + i * MAX_EARNINGS_CLAIMS_ROUNDS).toString(),
+        {
```
adamsoffer: You're passing the `gas` value as a parameter to the `claimEarnings` method which it doesn't accept. Try passing it as an option in the `request` method inside the `makeBatchRequest` function. See [here](https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#methods-mymethod-send).

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -65,7 +65,7 @@ export const getDelegatorStatus = (
 export const MAXIUMUM_VALUE_UINT256 =
   '115792089237316195423570985008687907853269984665640564039457584007913129639935'
 
-export const MAX_EARNINGS_CLAIMS_ROUNDS = 100
+export const MAX_EARNINGS_CLAIMS_ROUNDS = 50
```
adamsoffer: I use this constant to determine whether to show the claim earnings banner when it's been over 100 rounds. Let's add a separate constant here called `BATCH_TOTAL` and use that instead.

adamsoffer: Sidenote: eventually I should really fetch `MAX_EARNINGS_CLAIMS_ROUNDS` from contract storage instead of hard coding.

adamsoffer (changes requested):

packages/explorer-2.0/apollo/resolvers/Mutation.tsx
kyriediculous: I had to rewrite the function to make it work because of a couple of things:

- We can't estimate the gas within the loop, this will estimate the gas correctly only for the first iteration (for the next iterations it would estimate the gas as if we were claiming `i*MAX_BATCH_CLAIM_ROUNDS` rounds). 

- Previously there were two loops where one could be used 

- The previous closure function did not provide us access to the arguments used for a certain call we add to the batch to estimate the gas. While this might be retrievable from the object there was still the first issue to deal with

packages/explorer-2.0/lib/utils.tsx
kyriediculous: introduced `MAX_BATCH_CLAIM_EARNINGS = 50`

kyriediculous:
I rebased and force pushed already because the diff is so vastly  different from the original commits in this PR.

packages/explorer-2.0/apollo/resolvers/Mutation.tsx
adamsoffer: Nice

adamsoffer (approved):

adamsoffer:
Just tested on the preview URL. Looks and works great! 🚢

PR #638, opened 2020-04-20
Ar/user password reset

angyangie:
https://github.com/livepeer/livepeerjs/issues/589

https://github.com/livepeer/livepeerjs/issues/589

To-do:
- write tests

https://github.com/livepeer/livepeerjs/issues/589

**Checklist:**


packages/api/src/controllers/user.js
```diff
@@ -158,7 +159,143 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
...
+    }
+
+    const id = uuid()
+    let resetToken = uuid()
+    await req.store.create({
+      kind: 'password-reset-token',
+      id: id,
+      userId: userId,
+      resetToken: resetToken,
+      expiration: Math.floor(Date.now() / 1000 + 48 * 60 * 60),
```
iameli: Let's just store timestamps in milliseconds; more precise and less error-prone. [There's a library I like for this sort of thing called `ms`](https://www.npmjs.com/package/ms), which would let us just do `expiration: Date.now() + ms("48 hours")`.

packages/api/src/controllers/user.js
```diff
@@ -158,7 +159,143 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
...
+  '/password/reset-token',
+  validatePost('password-reset-token'),
+  async (req, res) => {
+    const email = req.body.email
+    const userIds = await req.store.query('user', { email: email })
+    if (userIds.length < 1) {
+      res.status(404)
+      return res.json({ errors: ['user not found'] })
+    }
+    const userId = userIds[0]
```
iameli: Nitpick, but somewhat more terse syntax for this:
```suggestion
    const [userId] = await req.store.query('user', { email: email })
    if (!userId) {
      res.status(404)
      return res.json({ errors: ['user not found'] })
    }
```

packages/livepeer.com/www/pages/reset-password.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useApi, useLoggedIn } from "../hooks";
+import { useRouter } from "next/router";
+
+export default () => {
+  useLoggedIn(false);
+  const [errors, setErrors] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const { resetPassword } = useApi();
+  const router = useRouter();
+  const { email, resetToken } = router.query;
```
iameli, line 15: Let's handle the case where someone types this URL in manually and these two variables are undefined. Something like...
```javascript
useEffect(() => {
  if (!email || !resetToken) {
    router.replace("/forgot-password")
  }
}, [email, resetToken])

File: packages/api/src/controllers/user.js [hunk 4]
```diff
@@ -158,7 +159,143 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
...
+      const token = await req.store.get(
+        `password-reset-token/${tokens[i]}`,
+        false,
+      )
+
+      if (token.resetToken === resetToken) {
+        dbResetToken = token
+      }
+    }
+
```
iameli, line 200: I know we decided it doesn't actually matter, but I'd feel a lot better with a check for `if (dbResetToken.userId !== userId) { send back 403 }` kind of deal.

iameli (changes requested):
Cool, minor changes

packages/api/src/controllers/user.js
angyangie: Ooo didn't know about ms. I'll use it!

packages/api/src/controllers/user.js
angyangie: Can do! I didn't include that check since here we're ensuring all tokens in consideration do belong to `user.id` or `userId` : 
```js
const tokens = await req.store.query('password-reset-token', {
    userId: user.id
})
```
Would you feel more comfortable with this: 
```js
if (!dbResetToken || dbResetToken.expiration < Date.now()) {
      res.status(403)
      return res.json({
        errors: ['incorrect or expired user validation token'],
      })
    }

    // change user password
    const [hashedPassword, salt] = await hash(password)
    await req.store.replace({
      ...user,
      password: hashedPassword,
      salt: salt,
      emailValid: true,
    })

    user = await req.store.get(`user/${userId}`)

    // delete all reset tokens associated with user
    for (const t of tokens) {
      await req.store.delete(`password-reset-token/${t}`)
    }

    res.status(201)
    return res.json(user)
```

packages/livepeer.com/www/pages/reset-password.tsx
angyangie: hmm that automatically redirects to `/reset-password` to `/forgot-password` ... no chance to put in the new password

iameli: Odd, it should only happen when those variables are undefined... meh, no big deal

packages/api/src/controllers/user.js
iameli: Eh yeah, as long as it's tested

iameli (approved):
LGTM!

Can you give the PR a more descriptive name than "Ar/user password reset" for the commit message?

iameli:
nvm i did it

Issue #618, opened 2020-04-06
Simple event tracking for core api release


wohlner:
- API key created, stream created (and by which API key), segments created, transcoding result returned.

- Must answer which keys are associated with actually streaming so that Mark can associate an email with the key and reach out.

iameli:
* New flag on API server `--hubspot-key` or something that can log certain user actions to HubSpot

PR #639, opened 2020-04-22
Tracking API Token's `lastSeen`

ya7ya:
Tracks last time an API token was used

- updated schema.yaml
- added non-blocking `tracking.record` function

Fixes https://github.com/livepeer/livepeerjs/issues/618

**Screenshots (optional):**

**Checklist:**


ya7ya:
This is good to go, I'll remove the `console.logs` once it's gets reviewed :smile:

packages/api/src/schema/schema.yaml
```diff
@@ -181,6 +181,9 @@ components:
         name:
           type: string
           example: Example Token
+        lastSeen:
+          type: string
+          example: 2020-04-22T12:35:13.297Z
```
iameli: Let's do these as numbers — in the password reset example we've already got a `Date.now()`-style "milliseconds since epoch" time in the database. I think that's probably the simplest solution unless we needed timezone-aware stuff for some reason

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,31 @@
...
+//     console.log('tracking: ', req.user, req.headers.authorization)
+//     return next()
+//   }
+// }
+
+// export default trackingFactory
+
+const tracking = {
+  record: function record (store, tokenObject) {
+    console.log('debug: tracking record trigger')
```
iameli: This can go away

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,31 @@
...
+    tokenObject.lastSeen = new Date()
+    store.replace(tokenObject).then((_) => {
+      // all good
+      console.log('all good')
```
iameli: Prolly don't need the catch block at all, and the error block can just log something

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+// function trackingFactory(params) {
```
iameli: Can clean up a little here

iameli (changes requested):
Cool, just some cleanup and a tweak to the timestamp format.

Do you want to find a "time ago" library we can use for that column on the front-end? A quick Googling pointed me toward this one but I'm sure there's many out there https://www.npmjs.com/package/timeago.js

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+const tracking = {
+  record: function record (store, tokenObject) {
+    tokenObject.lastSeen = Date.now()
+    store.replace(tokenObject).then((_) => {
```
iameli, line 6: I think we should log an error on the `.catch()` block here — otherwise that'll result in an uncaught promise that crashes the server!

packages/api/src/schema/schema.yaml
```diff
@@ -181,6 +181,9 @@ components:
         name:
           type: string
           example: Example Token
+        lastSeen:
+          type: string
```
iameli:
```suggestion
          type: number
```

packages/livepeer.com/www/components/TokenTable/index.tsx
```diff
@@ -168,7 +168,8 @@ export default ({ userId }) => {
           <Box>Last Active</Box>
         </TableRow>
         {tokens.map(token => {
-          const { id, name } = token;
+          const { id, name, lastSeen } = token;
+          const formattedLastSeen = `${new Date(lastSeen).toLocaleDateString()} ${new Date(lastSeen).toLocaleTimeString()}`
```
iameli, line 172: Cool, good enough for now

iameli (changes requested):
very minor changes

packages/livepeer.com/www/components/TokenTable/index.tsx
ya7ya: I prefer this way to `ago` because when you're debugging an API webhook, you wanna see exact absolute timestamps.

iameli (approved):
LGTM!

PR #640, opened 2020-04-22
refactoring geolocate and `/api/broadcaster` for the new setup

ya7ya:

- 
- 
- 

**Screenshots (optional):**

**Checklist:**


packages/api/src/middleware/geolocate.js
```diff
@@ -0,0 +1,45 @@
+'use strict'
+
+function geoLocateFactory (params) {
```
iameli, line 3: I'd like to have individual controllers opt in to the `first` behavior in the `params` — we'll want `/api/broadcaster` to do `geolocate({first: true})` but `/api/geolocate` should give us information about every region. I like the query parameter, though! `/api/geolocate?first=true` will be useful.

packages/api/src/middleware/geolocate.js
```diff
@@ -0,0 +1,45 @@
...
+  return async (req, res, next) => {
+    let first = req.query.first || true
```
iameli: This won't produce the expected results with `GET /api/geolocate?first=false`, as the string "false" is truthy. Perhaps we have explicit checks for `"true"` and `"false"`?

packages/api/src/middleware/geolocate.js
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      data = await Promise.all(promises)
+    }
+    const ret = {
+      chosenServer: smallestServer,
+      servers: data,
+    }
+
+    req.region = ret
+    console.log('region populated: ', ret)
```
iameli: can remove

packages/api/src/parse-cli.js
```diff
@@ -160,6 +160,12 @@ export default function parseCli(argv) {
             '[DO NOT USE EXCEPT FOR TESTING] token that test harness can use to bypass validation and access the database',
           type: 'string',
         },
+        ingests: {
+          describe:
+            'list of ingest endpoints to use as options for /api/geolocate',
+          type: 'array',
+          default: ['esh.livepeer.live', 'chi.livepeer-ac.live'],
```
iameli: I don't think we need a default value here, we wouldn't want external users hitting our API servers anyway. And we want to use the hardcoded list if this isn't provided.

packages/api/src/parse-cli.js
```diff
@@ -160,6 +160,12 @@ export default function parseCli(argv) {
             '[DO NOT USE EXCEPT FOR TESTING] token that test harness can use to bypass validation and access the database',
           type: 'string',
         },
+        ingests: {
```
iameli: Hmmmmm contemplating the name here. Right now they're mostly used for ingest, but they might be used later for other things as well... maybe `region`? So it can be called like
`node api.js --region chi.livepeer-ac.live --region esh.livepeer.live`

packages/api/src/parse-cli.js
```diff
@@ -160,6 +160,12 @@ export default function parseCli(argv) {
...
+          describe:
+            'list of ingest endpoints to use as options for /api/geolocate',
+          type: 'array',
```
iameli, line 166: So one trick here is that we want this to be able to be configured entirely with environment variables in the worker context. (And elsewhere.) `yargs` doesn't support this for its array types by default, but [I wrote a workaround in this ticket](https://github.com/yargs/yargs/issues/821#issuecomment-614434134). I think it'll work if you just include that `coerce` function verbatim.

iameli (changes requested):
Sweet, few tweaks

packages/api/src/app-router.js
```diff
@@ -117,6 +117,7 @@ export default async function makeApp(params) {
   // Add a controller for each route at the /${httpPrefix} route
   const prefixRouter = Router() // amalgamates our endpoints together and serves them out
   for (const [name, controller] of Object.entries(controllers)) {
+    console.log('debug: hooking up route ', name)
```
iameli: cleanup

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+// function trackingFactory(params) {
```
iameli: cleanup

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,30 @@
...
+// }
+
+// export default trackingFactory
+
+const tracking = {
+  record: function record (store, tokenObject) {
+    tokenObject.lastSeen = new Date()
+    store.replace(tokenObject).then((_) => {
+      // all good
+      console.log('all good')
```
iameli: cleanup

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,30 @@
...
+    }).catch((e) => {
+      // TODO Log error
```
iameli:
```suggestion
      console.error("error updating user API token", e)
```

packages/api/src/middleware/tracking.js
```diff
@@ -0,0 +1,30 @@
...
+//     console.log('tracking: ', req.user, req.headers.authorization)
+//     return next()
+//   }
+// }
+
+// export default trackingFactory
+
+const tracking = {
+  record: function record (store, tokenObject) {
+    tokenObject.lastSeen = new Date()
```
iameli:
```suggestion
    tokenObject.lastSeen = Date.now()
```

packages/api/src/schema/schema.yaml
```diff
@@ -181,6 +181,9 @@ components:
         name:
           type: string
           example: Example Token
+        lastSeen:
+          type: string
+          example: 2020-04-22T12:35:13.297Z
```
iameli:
```suggestion
          type: number
          example: 1587673332547
```

iameli (changes requested):
Cool, just a little cleanup and a switch from string to number for dates

iameli:
Oh oops, looks like I reviewed the `lastSeen` stuff here — want to either merge the two branches or rebase out `lastSeen` from this one?

ya7ya:
I rebased the tracking features out to keep the PR clean :+1: , this is ready for review

packages/api/src/schema/schema.yaml
```diff
@@ -182,8 +182,13 @@ components:
           type: string
           example: Example Token
         lastSeen:
+<<<<<<< HEAD
```
iameli: Ruh roh

ya7ya: sorry :facepalm:, :coffee:  fixed

iameli (approved):
LGTM!

PR #642, opened 2020-04-23
add MarkdownSection

iameli:
Just outputs totally uncustomized vanilla Markdown for now but I think this might be all that's necessary on the Sanity side of things


adamsoffer (approved):
Looks good 💪 🚢

PR #649, opened 2020-04-27
stream status

darkdarkdragon:
Adds stream-info service that should be run in the same pod as broadcaster.
It pulls broadcaster's `/status` endpoint constantly, pulls manifests of streams
that broadcaster currently transcoding and updates
`lastSeen`, `sourceSegments` and `transcodedSegments` fields of according Stream object.

Also shows list of user's streams in the UI.

- add separate stream-info-service
- adds `lastSeen`, `sourceSegments` and `transcodedSegments` fields to the Stream object
- In the UI (TokenTable) also pulls user's stream list and shows it

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/add-indexes.ts
+++ b/packages/api/src/add-indexes.ts
@@ -1,0 +1,114 @@
+
+/*
+  Adds index for `userId` field for existing stream objects
+*/
+
+import yargs from 'yargs'
+import path from 'path'
+import os from 'os'
+import {
+  createStore
+} from './app/store'
+import {
+  IStore
+} from './types/common'
+
+
+function parseCli() {
+  return (
+    yargs
+      .usage(
+        `
+    Livepeer Stream Info fetcher
+
+    Options my also be provided as LP_ prefixed environment variables, e.g. LP_PORT=5000 is the same as --port=5000.
+
+    `,
+      )
+      .env('LP_')
+      .options({
+        storage: {
+          describe: 'storage engine to use',
+          default: 'level',
+          demandOption: true,
+          type: 'string',
+          choices: [
+            'level',
+            'postgres',
+            'cloudflare',
+            'cloudflare-cluster',
+            'firestore',
+          ],
+        },
+        'db-path': {
+          describe: 'path to LevelDB database',
+          default: path.resolve(os.homedir(), '.livepeer', 'api'),
+          type: 'string',
+        },
+        'postgres-url': {
+          describe: 'url of a postgres database',
+          type: 'string',
+        },
+        'cloudflare-namespace': {
+          describe: 'namespace of a cloudflare database',
+          type: 'string',
+        },
+        'cloudflare-account': {
+          describe: 'account id of a cloudflare database',
+          type: 'string',
+        },
+        'cloudflare-auth': {
+          describe: 'auth of a cloudflare database',
+          type: 'string',
+        },
+        'firestore-credentials': {
+          describe:
+            'JSON string of service account credentials for a GCP account',
+          type: 'string',
+        },
+        'firestore-collection': {
+          describe:
+            'name of the top-level firestore collection for storing our data',
+          type: 'string',
+        },
+      })
+      .help()
+      .parse()
+  )
+}
+
+
+async function addIndex(kind: string, fieldName: string, store: IStore) {
+  console.log('adding userId index to stream objects:')
+  let data, cur, limit = 1
+  while (true) {
+    let { data, cursor } = await store.list('stream/', cur, limit, false)
+    console.log(`got: cursor: ${cursor}`, data)
+    if (!cursor || !data.length) {
+      break
+    }
+    cur = cursor
+    data = data.map(o => o[Object.keys(o)[0]])
+    for (let obj of data) {
+      const key = `${kind}+${fieldName}/${obj[fieldName]}/${obj['id']}`
+      console.log(`key: ${key}`)
+      try {
+        await store.backend.create(key, {})
+      } catch (e) {
+        console.log(e)
+      }
+    }
+  }
+}
+
+async function main() {
+  require('dotenv').config()
+  const params = parseCli()
+  const store = createStore(params)
+  await store.ready
+
+  await addIndex('stream', 'userId', store)
+  await store.close()
+}
+
+main()
--- a/packages/api/src/app/store.ts
+++ b/packages/api/src/app/store.ts
@@ -1,0 +1,47 @@
+
+import {
+  LevelStore,
+  PostgresStore,
+  CloudflareStore,
+  CloudflareClusterStore,
+  FirestoreStore,
+} from '../store'
+
+import {
+  IStore
+} from '../types/common'
+
+export function createStore(params: any): IStore {
+  let store
+  const {
+    storage,
+    dbPath,
+    postgresUrl,
+    cloudflareNamespace,
+    cloudflareAccount,
+    cloudflareAuth,
+    firestoreCredentials,
+    firestoreCollection,
+  } = params
+
+  if (storage === 'level') {
+    store = LevelStore({ dbPath })
+  } else if (storage === 'postgres') {
+    store = PostgresStore({ postgresUrl })
+  } else if (storage === 'cloudflare') {
+    store = CloudflareStore({
+      cloudflareNamespace,
+      cloudflareAccount,
+      cloudflareAuth,
+    })
+  } else if (storage === 'cloudflare-cluster') {
+    store = CloudflareClusterStore({
+      cloudflareNamespace,
+    })
+  } else if (storage === 'firestore') {
+    store = FirestoreStore({ firestoreCredentials, firestoreCollection })
+  } else {
+    throw new Error('Missing storage information')
+  }
+  return store
+}
--- a/packages/api/src/app/stream-info/stream-info-app.ts
+++ b/packages/api/src/app/stream-info/stream-info-app.ts
@@ -129,7 +129,9 @@ class statusPoller {
     try {
       status = await this.getStatus(this.broadcaster)
     } catch (e) {
-      console.log(`got error fetch status: `, e)
+      if (e.code !== 'ECONNREFUSED') {
+        console.log(`got error fetch status: `, e)
+      }
       return
     }
     // console.log(`got status: `, status)
--- a/packages/api/src/app/stream-info/stream-info-app.ts
+++ b/packages/api/src/app/stream-info/stream-info-app.ts
@@ -1,0 +1,278 @@
+
+import express, { Router } from 'express'
+import 'express-async-errors' // it monkeypatches, i guess
+import morgan from 'morgan'
+import {
+  createStore
+} from '../store'
+import { healthCheck } from '../../middleware'
+import logger from '../../logger'
+import {
+  Stream,
+} from '../../schema/types'
+
+import {
+  IStore
+} from '../../types/common'
+
+import {
+  StatusResponse,
+  MasterPlaylist
+} from './livepeer-types'
+
+const pollInterval = 5000
+const updateInterval = 60 * 1000
+const deleteTimeout = 5 * 60 * 1000
+const seenSegmentsTimeout = 60 * 1000
+
+async function makeRouter(params) {
+  // Storage init
+  const bodyParser = require('body-parser')
+  let store = createStore(params)
+  await store.ready
+
+  // Logging, JSON parsing, store injection
+
+  const app = Router()
+  app.use(healthCheck)
+  app.use(bodyParser.json())
+  app.use((req, res, next) => {
+    req.store = store
+    req.config = params
+    next()
+  })
+
+  // If we throw any errors with numerical statuses, use them.
+  app.use(async (err, req, res, next) => {
+    if (typeof err.status === 'number') {
+      res.status(err.status)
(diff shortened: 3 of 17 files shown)


packages/api/src/controllers/stream.js
```diff
@@ -11,13 +11,29 @@ import path from 'path'
 
 const app = Router()
 
-app.get('/', authMiddleware({ admin: true }), async (req, res) => {
+app.get('/', authMiddleware({}), async (req, res) => {
   let limit = req.query.limit
   let cursor = req.query.cursor
   logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
+  if (req.user.admin !== true) {
```
angyangie: I see what you're doing here ... if a user does not have admin privileges, but still hits this endpoint, they'll get a list of their own streams. What do you think of making this a separate endpoint, something like `stream/user/:userId` (wording up to you)? Might be nice to not have to do the JWT check on line 30, since we're already doing it in the auth file

packages/api/src/controllers/stream.js
```diff
@@ -30,12 +46,18 @@ app.get('/', authMiddleware({ admin: true }), async (req, res) => {
     res.links({
       next: next.href,
     })
+    output = output.map(o => o[Object.keys(o)[0]])
```
angyangie: Good catch on removing keys from the response from `list`

packages/api/src/controllers/stream.js
```diff
@@ -30,12 +46,18 @@ app.get('/', authMiddleware({ admin: true }), async (req, res) => {
...
   } // CF doesn't know what this means
   res.json(output)
 })
 
 app.get('/:id', authMiddleware({}), async (req, res) => {
   const output = await req.store.get(`stream/${req.params.id}`)
+  if (!output || output.userId !== req.user.id && !(req.user.admin && req.authTokenType == 'JWT')) {
```
angyangie, line 60: 👍

packages/api/src/controllers/stream.test.js
```diff
@@ -64,12 +64,49 @@ afterEach(async () => {
   await clearDatabase(server)
 })
 
+async function setupUsers(server) {
```
angyangie, line 67: Thank you! I was going to circle back and do something like this for test files. Too much repeated code in tests

packages/api/src/controllers/stream.test.js
```diff
@@ -64,12 +64,49 @@ afterEach(async () => {
...
+}
+
 describe('controllers/stream', () => {
   describe('basic CRUD with JWT authorization', () => {
-    let client
-    let adminUser
+    let client, adminUser, adminToken, nonAdminUser, nonAdminToken
 
     beforeEach(async () => {
+      let { _client, _adminUser, _adminToken, _nonAdminUser, _nonAdminToken } = await setupUsers(server)
```
angyangie: Why do you first capture variables as `_client`, then set equal to their equivalent non-`_` name? Want to learn this one

packages/api/src/controllers/stream.test.js
```diff
@@ -176,34 +214,42 @@ describe('controllers/stream', () => {
       }
     })
 
-    it('should not get all streams with non-admin user', async () => {
+    it('should get only own streams with non-admin user', async () => {
       // setting up non-admin user
+      /*
```
angyangie: I think we can just get rid of all this commented out code everywhere in the test files

packages/api/src/controllers/stream.test.js
```diff
@@ -176,34 +214,42 @@ describe('controllers/stream', () => {
...
       nonAdminUser = { ...nonAdminUserRes, emailValid: true }
       await server.store.replace(nonAdminUser)
+      */
 
-      for (let i = 0; i < 3; i += 1) {
+      for (let i = 0; i < 5; i += 1) {
         const document = {
           id: uuid(),
           kind: 'stream',
+          userId: i < 3 ? nonAdminUser.id : undefined,
```
angyangie, line 201: Love this! So much less code to write

packages/api/src/controllers/stream.test.js
```diff
@@ -221,22 +267,30 @@ describe('controllers/stream', () => {
     })
   })
 
-  describe('object stores endpoint with api key', () => {
```
angyangie, line 224: 😬

packages/api/src/controllers/stream.test.js
```diff
@@ -250,25 +304,43 @@ describe('controllers/stream', () => {
...
       client.apiKey = nonAdminApiKey
       let res = await client.get(`/stream`)
-      expect(res.status).toBe(403)
+      expect(res.status).toBe(200)
+      const streams = await res.json()
+      expect(streams.length).toEqual(3)
+      expect(streams[0].userId).toEqual(nonAdminUser.id)
+    })
 
+    it('should not get all streams for admin user', async () => {
```
angyangie, line 296: "should not get all streams for admin user with APIKey" perhaps?

packages/api/src/store/postgres-store.js
```diff
@@ -45,8 +45,8 @@ export default class PostgresStore {
 
     if (cursor) {
       res = await this.pool.query(
-        `SELECT * FROM ${TABLE_NAME} WHERE id LIKE $1 AND id > $2 LIMIT $3`,
-        [`${prefix}%`, `${prefix}${cursor}`, `${limit}`],
+        `SELECT * FROM ${TABLE_NAME} WHERE id LIKE $1 LIMIT $3 OFFSET $2`,
```
angyangie: Why do we OFFSET here?

packages/api/src/store/postgres-store.js
```diff
@@ -62,10 +62,13 @@ export default class PostgresStore {
     })
 
     if (data.length < 1) {
-      return { data: data, cursor: null }
+      return { data, cursor: null }
+    }
+    if (limit) {
+      cursor = (cursor | 0) + limit
```
angyangie: Does the ID of the last element always equal the limit? Maybe I am not understanding what this line is doing

packages/api/src/add-indexes.ts
```diff
@@ -0,0 +1,114 @@
+
+/*
+  Adds index for `userId` field for existing stream objects
```
angyangie, line 3: Awesome! Is this a script that we'd only have to run once, to add userId fields to stream objects? Useful to have for future backfilling of other fields

packages/api/src/app/store.ts
```diff
@@ -0,0 +1,47 @@
...
+  CloudflareStore,
+  CloudflareClusterStore,
+  FirestoreStore,
+} from '../store'
+
+import {
+  IStore
+} from '../types/common'
+
+export function createStore(params: any): IStore {
```
angyangie, line 14: Is this something we can use in `app-router.js`? Haven't worked with ts before, so not sure what that would look like when used in a js file

packages/api/src/types/common.ts
```diff
@@ -0,0 +1,48 @@
...
+  replace(data: StoredObject): Promise<void>
+  list(prefix: string, cursor: any, limit: number, cleanWriteOnly?: boolean): Promise<{ data: Array<StoredObject>, cursor: any }>
+  listKeys(prefix: string, cursor: any, limit: number): Promise<[Array<string>, any]>
+  query(kind: string, queryObj: object, cursor: any, limit: number, cleanWriteOnly?:boolean) : Promise<Array<string>>
+  deleteKey(key: string) : Promise<void>
+  delete(id: string): Promise<void>
+  create(data: StoredObject): Promise<StoredObject>
+}
+
+export interface IStoreBackend {
```
angyangie, line 40: Do we use this anywhere? Or is it necessary when we're using a DB backend?

packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -0,0 +1,278 @@
...
+        si.sourceSegmentsLastUpdated = si.sourceSegments
+        si.transcodedSegmentsLastUpdated = si.transcodedSegments
+      }
+    }
+    for (const [mid, si] of this.seenStreams) {
+      const now = Date.now()
+      if (!(mid in status.Manifests)) {
+        const notSeenFor: number = now - si.lastSeen.valueOf()
+        if (notSeenFor > deleteTimeout) {
+          this.seenStreams.delete(mid)
```
angyangie, line 170: Nice. Clever way of deleting streams from "active" broadcaster streams.

packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -0,0 +1,278 @@
...
+} from '../../types/common'
+
+import {
+  StatusResponse,
+  MasterPlaylist
+} from './livepeer-types'
+
+const pollInterval = 5000
+const updateInterval = 60 * 1000
+const deleteTimeout = 5 * 60 * 1000
```
angyangie, line 25: Out of curiosity, how did we decide on these timeouts?

angyangie (commented):
You are a typescript ninja 🦘

packages/livepeer.com/www/pages/app/user.tsx
```diff
@@ -24,7 +25,8 @@ export default () => {
           alignItems: "center"
         }}
       >
-        <TokenTable id="API Token Table" userId={user.id} />
+        <TokenTable id="Streams Table" userId={user.id} />
+        <StreamsTable id="API Token Table" userId={user.id} />
```
angyangie: `TokenTable` and `StreamsTable` might be switched around here

packages/api/src/controllers/stream.test.js
darkdarkdragon: API key mentioned in `describe` earlier,  so error message looks like:

`controllers/stream › stream endpoint with api key › should not get all streams for admin user

packages/api/src/controllers/stream.js
darkdarkdragon: Moved it into separate endpoint

packages/api/src/controllers/stream.test.js
darkdarkdragon: This is because I've thought that we can't destructure into existing variables. But I've rechecked this and turned out I've missed brackets in the syntax - I've changed code, take a look.

packages/api/src/controllers/stream.test.js
darkdarkdragon: fixed

packages/api/src/add-indexes.ts
darkdarkdragon: Yes, it's a 'database schema' migration script. Should be run once.

packages/api/src/app/store.ts
darkdarkdragon: Not with current config. I've just used babel's preset for adding typescript, and it is compiles js files to the 'dist' folder, so compiled files can't be used in the pure js sources.
This can be done by changing configuration of typescript compiler to put compiled js files alongside ts ones - that way they can be used from js source. But don't think we need to do this now - this PR already big 😃

packages/api/src/types/common.ts
darkdarkdragon: Nope. I've actually wrote it first as IStore interface and only then noticed that backend is wrapped by Model. Left it here for possible future use.

packages/livepeer.com/www/pages/app/user.tsx
darkdarkdragon: fixed

packages/api/src/app/store.ts
angyangie: Really good to know. Thanks for explaining! Did a little more reading on the subject to get even more informed - https://iamturns.com/typescript-babel/

packages/api/src/types/common.ts
angyangie: Cool!

packages/api/src/controllers/stream.test.js
angyangie: Nice!

darkdarkdragon:
@angyangie about my postgres change - my mistake. It didn't work for me, but I didn't put enough thought into why it is was not working. I've reverted my change, will tackle this problem with separate PR

packages/api/src/app/stream-info/stream-info-app.ts
darkdarkdragon: My thoughts was like these:
- polling interval - we want to see that stream has started in browser in around 10 seconds - so I'm polling broadcaster every 5 sec and if see new stream then updating database
- update interval - but I don't want to make write to database every 5 secs (every write request gets billed) - so I'm updating db record only once in a minute
- delete timeout - we need to clear memory from old streams, but we don't want to do it too soon - some broadcasters can get some segments just occasionally, so I'm keeping data for some time

angyangie: Nice!  Writing to the DB once a minute makes a lot of sense, as does deleting every 5 min. Is the reason for the polling interval being every 5 sec, so that we can capture streams that are started and maybe stopped before the minute interval is over? So that we ensure all streams longer than 5 seconds get logged here?

darkdarkdragon: Reason for 5 sec pulling is for fast update of UI when stream started

angyangie (approved):

PR #651, opened 2020-04-27
Combine steps 3 and 4

ericxtang:

Based on @j0sh's comment, we can pipe the POST result into `munpack`, reducing a step in the tutorial :)

### DIFF
--- a/packages/livepeer.com/www/docs/index.mdx
+++ b/packages/livepeer.com/www/docs/index.mdx
@@ -80,7 +80,7 @@ curl https://livepeer.com/api/broadcaster \
 You will get back a list of `json` objects that represent a list of broadcaster addresses. Pick any of the broadcaster addresses to send the source `.ts` segment.
 
 **Step 3:** Now it's time to POST your .ts segment. to
-`{broadcaster_address}/live/{stream_id}/{media_number}.ts`
+`{broadcaster_address}/live/{stream_id}/{media_number}.ts`, and get your results.
 
 Set `Accept: multipart/mixed` headers. A continuous livestream should start with
 the `media_number` set to `1`.ts and count upward to `2`.ts, `3`.ts and so on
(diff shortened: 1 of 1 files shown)


packages/livepeer.com/www/docs/index.mdx
```diff
@@ -92,20 +92,23 @@ You'll need to replace that URL a broadcaster address from Step 2.
...
 
-When you integrate your own video platform with Livepeer, your platform should
-split the response into 3 different video segments based on the text delimiter,
-and insert them into a HLS playlist for playback. 
+_if you don't see `1080p_1.ts`, `720p_1.ts`, and `360p_1.ts` most likely there 
+was an issue with the request you sent.  Please visit our [support channel on Discord](https://discord.gg/98g8Dqn), 
+or email support@livepeer.org._
 
-The content of the file looks something like:
+_We used [`munpack`](https://linux.die.net/man/1/munpack) to unpack the multipart
```
wohlner: What do you think about moving the munpack instructions to line 84, at the start of step 3. This is directly before the user is required to munpack. Otherwise, many will surely get errors and not make it to the bottom of the final step.

packages/livepeer.com/www/docs/index.mdx
```diff
@@ -92,20 +92,23 @@ You'll need to replace that URL a broadcaster address from Step 2.
...
+was an issue with the request you sent.  Please visit our [support channel on Discord](https://discord.gg/98g8Dqn), 
+or email support@livepeer.org._
 
-The content of the file looks something like:
+_We used [`munpack`](https://linux.die.net/man/1/munpack) to unpack the multipart 
+return.  To install `munpack`, use `brew install mpack` for macos, 
+and `apt install mpack` for Linux. Alternatively, you can save the response directly
+to a file. The content of the API response looks something like:_
 ```
 --27bee784e2f2672487f5
```
j0sh: If we do keep this section in, this probably also needs the original HTTP response headers (eg, full output of `curl -i ...`) in order to identify and extract the `boundary` ? If we don't want to bother describing the response in full, then we might as well remove this section.

wohlner (commented):
I suggest moving munpack directions up to above where the command is used.

packages/livepeer.com/www/docs/index.mdx
wohlner: Also, combining steps 3 and 4 combines the transcoding, unpacking and playback into one step. That's a lot for a tutorial. I fear we are missing out on the "aha" moment of unpacking and playing back the segment. If others feel differently, I won't be stubborn about it. But let's talk about it before deploying this change to prod.

j0sh: The fewer steps, the better, especially for a tutorial that's designed to get folks up and running quickly.

Unpacking isn't particularly enlightening unless you're looking to actively develop against the intermediate results. And even then, the format should be properly documented, which it is not right now. Incomplete / confusing documentation is worse than no documentation, because it just wastes people's time.

ericxtang: Moved the `munpack` instructions based on @wohlner's suggestion.  I think the user can still play back the result after the combined step, and have the "aha" moment.

packages/livepeer.com/www/docs/index.mdx
```diff
@@ -82,17 +82,17 @@ You will get back a list of `json` objects that represent a list of broadcaster
 **Step 3:** Now it's time to POST your .ts segment. to
 `{broadcaster_address}/live/{stream_id}/{media_number}.ts`, and get your results.
 
+_We used [`munpack`](https://linux.die.net/man/1/munpack) to unpack the multipart
```
wohlner: Make it future tense: "We _will_ use munpack to unpack the multipart API response." Otherwise it looks good.

wohlner (changes requested):
Make it future tense, but otherwise all good.

packages/livepeer.com/www/docs/index.mdx
j0sh: Or just "We use munpack" ... ?

ericxtang:
Updated according to comments by @wohlner and @j0sh.

ericxtang:
@wohlner and @j0sh can you give a review today so we can merge it quickly?

j0sh (approved):

PR #658, opened 2020-04-29
Add segment script

adamsoffer:
Adds [segment](https://segment.com/) script to the head per Mark


adamsoffer:
@iameli after this is merged into `master`, I can then merge to `latest`, push up and it will automatically get deployed to livepeer.com right?

Are these the correct steps after a merge into master?

1. `git pull origin master`
2. `git checkout latest`
3. `git rebase master`
4. `git push origin latest`

iameli (changes requested):
This is going to happen in the development and staging environments too, which has caused problems before. Can we get that `o7EtavalnTN8dqxDzpEiyIpF1Xae2RoJ` in an environment variable in the Now build like we do with the other tracking stuff, and only render this `<script>` if it's present?

adamsoffer:
@iameli I moved the key into an env var and only load the script while in prod. Also, I noticed there was an example integration for [segment + nextjs](https://github.com/zeit/next.js/tree/canary/examples/with-segment-analytics) which I ended up following.

iameli (approved):
LGTM!

PR #661, opened 2020-04-30
fixing token page flashing for unverified users 

### LINKED ISSUE #654: www: see a flash of the token page when email is unverified
The logic for this is in `www/hooks/use-logged-in.tsx`

Maybe an initial `display: none` or `opacity: 0` or something

ya7ya:
avoid flashing the `TokenTable` component when the user hasn't verified their email. 

Fixes #654

### DIFF
--- a/packages/livepeer.com/www/pages/app/user.tsx
+++ b/packages/livepeer.com/www/pages/app/user.tsx
@@ -10,7 +10,7 @@ import TokenTable from "../../components/TokenTable";
 export default () => {
   useLoggedIn();
   const { user, logout } = useApi();
-  if (!user) {
+  if (!user || !user.emailValid) {
     return <Layout />;
   }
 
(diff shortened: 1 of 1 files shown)


packages/livepeer.com/www/pages/app/user.tsx
```diff
@@ -10,7 +10,7 @@ import TokenTable from "../../components/TokenTable";
 export default () => {
   useLoggedIn();
   const { user, logout } = useApi();
-  if (!user) {
+  if (!user || !user.emailValid) {
```
iameli: Let's do `user.emailValid === false` here — I think this field will be set in all cases but if it's unset it probably means email verification is disabled

iameli (changes requested):
tiny tweak

iameli (approved):
LGTM!

Issue #653, opened 2020-04-28
Fix HubSpot Integration


iameli:
We want this to happen in prod but not dev/staging

PR #668, opened 2020-05-01
ar/register hubspot contact tracking and notifications

angyangie:
- Sends SenGrid sales email when user signs up and verifies.
- Logs a "page view" on individual HubSpot contact record when user does any of the following: logs in, registers, verifies email, requests password reset token, resets password, and creates api token.

Part of this one: https://github.com/livepeer/livepeerjs/issues/653
This one: https://github.com/livepeer/livepeerjs/issues/676

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/user.js
+++ b/packages/api/src/controllers/user.js
@@ -150,6 +150,40 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
 
   let user = await req.store.get(`user/${userIds[0]}`, false)
   if (user.emailValidToken === req.body.emailValidToken) {
+    // alert sales of new verified user
+    const { supportAddr, sendgridTemplateId, sendgridApiKey } = req.config
+    const protocol =
+      req.headers['x-forwarded-proto'] === 'https' ? 'https' : 'http'
+    const buttonUrl = `${protocol}://${req.headers.host}/login`
+    const unsubscribeUrl = `${protocol}://${req.headers.host}/#contactSection`
+    const salesEmail = 'sales@livepeer.org'
+
+    try {
+      // send email verification message to user using SendGrid
+      await sendgridEmail({
+        email: salesEmail,
+        supportAddr,
+        sendgridTemplateId,
+        sendgridApiKey,
+        subject: `User ${user.email} signed up with Livepeer!`,
+        preheader: 'We have a new verified user',
+        buttonText: 'Log into livepeer',
+        buttonUrl: buttonUrl,
+        unsubscribe: unsubscribeUrl,
+        text: [
+          `User ${user.email} has signed up and verified their email with Livepeer!`,
+        ].join('\n\n'),
+      })
+    } catch (err) {
+      res.status(400)
+      return res.json({
+        errors: [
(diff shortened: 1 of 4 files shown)


angyangie:
LOTS of issues with the tracking code. @adamsoffer and I chatted about them. We'll have to take a closer look to get it to work perfectly. Whether we want to fix these issues depends on how we want to do tracking - is Segment enough, etc. etc. So for now, we'll table them.

**Known Issues:**
1. When a user has multiple Livepeer accounts, logs in with `email1`, logs out, then logs in with `email2`, all user activity on the site gets logged on `email1`'s hubspot contact page.
2. Forms (which are probably irrelevant now that we're using SendGrid to alert Mark that a user has signed up) do not get submitted consistently.
3. A view of a specific page only gets logged when the user first lands on livepeer or first logs in. If you click on a different tab in our navigation tabs, for example `Documentation`, that doesn't get logged into HubSpot. It does get logged into hubspot if you refresh the page. 

**Solutions:**
1. I suspect this is a cookie issue. However, only an issue for users with multiple accounts.
2. If we at any point want forms to be logged consistently, we have to remove javascript calls to the livepeer API, etc. from `onSubmit` on each form. Forms get logged very inconsistently the way things are set up now.
3. We need to get the hubspot script to load upon tab change - right now we're treating livepeer like a single page application re: hubspot.

packages/api/src/controllers/user.js
```diff
@@ -150,6 +150,40 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
 
   let user = await req.store.get(`user/${userIds[0]}`, false)
   if (user.emailValidToken === req.body.emailValidToken) {
+    // alert sales of new verified user
+    const { supportAddr, sendgridTemplateId, sendgridApiKey } = req.config
+    const protocol =
+      req.headers['x-forwarded-proto'] === 'https' ? 'https' : 'http'
+    const buttonUrl = `${protocol}://${req.headers.host}/login`
+    const unsubscribeUrl = `${protocol}://${req.headers.host}/#contactSection`
+    const salesEmail = 'sales@livepeer.org'
```
ya7ya, line 159: It's probably better not to hard code a `salesEmail` address, just in case we switch domains. But it's not a big deal ( it's an optional tweak i mean )

ya7ya (approved):
Very nice work

LGTM :+1:

packages/livepeer.com/www/hooks/use-api.tsx
```diff
@@ -27,6 +33,16 @@ const storeToken = token => {
   }
 };
 
+const trackPageView = (email, path = null) => {
+  var _hsq = (window._hsq = window._hsq || []);
+  _hsq.push(["identify", { email: email }]);
+  if (path) {
+    console.log(`here is pathhh: ${path}`);
```
ya7ya: clean that up first :D

packages/api/src/controllers/user.js
angyangie: From our discord convo, agree with you @ya7ya . We should merge this to have records of users on our site, and open more PRs later for optimization. Especially since this is just our first step into tracking a user. There will be a lot more to do with Hubspot + other tools we integrate

PR #712, opened 2020-05-12
Increase precision in voting percentages

adamsoffer:
This PR increases the total support and total participation precision from 4 to 5 units, so that even if you vote with very little stake, you can still see a change in the total support or total participation percentage before it rounds up.

**Screenshots**
Notice, the total support now displays as 99.999% rather than 100% to reflect the single "No" vote with 54.43 LPT at stake.

image

**Checklist:**


yondonfu (approved):
LGTM

Issue #705, opened 2020-05-11
Log referral information for each visit


angyangie:

angyangie:
- Capture where we were just before … from google search? Landing page?
- Example: multiple LinkedIn campaigns, ideally, which LI campaign?

Issue #703, opened 2020-05-11
Add livepeer.com UI tracking


angyangie:
- User visits a page on livepeer.com and what page(s)
- Livepeer.com sessions (info on sessions)
- Livepeer.com Unique sessions

PR #717, opened 2020-05-13
Ar/adding identity to livepeer.com ui tracking

angyangie:
https://github.com/livepeer/livepeerjs/issues/703
https://github.com/livepeer/livepeerjs/issues/705

- Adds identity to every page view when a user signs up, which associates anonymous records with specific `emails` / `userIds` that can then be used to aggregate all interactions with Livepeer a user might have much more accurately. This info can then be used to pump data back into Hubspot or any other service.
- Tracks referral information
- Removes HubSpot tracking code to prevent duplicates

Segment.com and Hubspot.com

https://github.com/livepeer/livepeerjs/issues/703
https://github.com/livepeer/livepeerjs/issues/705

### DIFF
--- a/packages/api/src/controllers/helpers.js
+++ b/packages/api/src/controllers/helpers.js
@@ -147,11 +147,12 @@ export async function sendgridEmail({
 }
 
 export async function trackAction(userId, email, event, segmentApiKey) {
-  var analytics = new SegmentAnalytics(segmentApiKey)
-  // note: for development, add `{ flushAt: 1 }` as a second parameter to SegmentAnalytics
+  var analytics = new SegmentAnalytics(segmentApiKey, { flushAt: 1 })
   analytics.identify({
     userId: userId,
-    traits: { email },
+    traits: {
+      email: email,
+    },
   })
 
   const properties = {}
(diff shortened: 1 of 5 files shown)


angyangie:
Confirmed with Mark that it is safe to remove HubSpot code and lose the HubSpot page view records going forward. They will live in Segment.

packages/api/src/controllers/helpers.js
```diff
@@ -144,3 +145,27 @@ export async function sendgridEmail({
...
+    traits: {
+      email: email,
+    },
+  })
+
+  const properties = {}
+  if ('properties' in event) {
+    for (const key in Object.keys(event.properties)) {
+      properties[key] = event.properties[key]
+    }
```
iameli: Easier way to write this:
```suggestion
    properties = {...properties, ...event.properties}
```

packages/api/src/controllers/helpers.js
```diff
@@ -144,3 +145,27 @@ export async function sendgridEmail({
   SendgridMail.setApiKey(sendgridApiKey)
   await SendgridMail.send(msg)
 }
+
+export async function trackAction(userId, email, event, segmentApiKey) {
+  var analytics = new SegmentAnalytics(segmentApiKey, { flushAt: 1 })
```
iameli, line 153: Do nothing if no segmentApiKey

iameli (approved):
Let's just make sure we don't track if we don't have a segment code and whatnot

PR #723, opened 2020-05-18
Stream delete endpoint

### LINKED ISSUE #675: Mark a stream as deleted

### LINKED ISSUE #679: Model.query doesn't return cursor
Model.query doesn't return cursor, but here
https://github.com/livepeer/livepeerjs/blob/master/packages/api/src/controllers/object-store.js#L38
`.cursor` property of `query`'s response is used

**To Reproduce (required)**

**Device (required):**
 - Device: [e.g. MacBook Pro, iPhone6]
 - OS: [e.g. iOS, Linux]
 - Browser [e.g. Chrome, Safari]
 
**Screenshots (optional)**

**Console logs (optional)**
```
Paste related logs and/or error messages from the console here.
Network issues, such as the cURL output from failing requests, can also be pasted here.
```

darkdarkdragon:
Improves data storage abstraction
Adds `deleted` field to `stream` object and delete stream endpoint

Fixes #675
Fixes #679 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -42,21 +42,15 @@ app.get('/user/:userId', authMiddleware({}), async (req, res) => {
     })
   }
 
-  const { data: streamIds, cursor: cursorOut } = await req.store.query({
+  const { data: streams, cursor: cursorOut } = await req.store.queryObjects({
     kind: 'stream',
     query: { userId: req.params.userId },
     cursor,
-    limit
+    limit,
+    filter: (o => !o.deleted)
   })
-  const streams = []
-  for (let i = 0; i < streamIds.length; i++) {
-    const stream = await req.store.get(`stream/${streamIds[i]}`, false)
-    if (!stream.deleted) {
-      streams.push(stream)
-    }
-  }
   res.status(200)
-  if (streamIds.length > 0 && cursorOut) {
+  if (streams.length > 0 && cursorOut) {
     res.links({ next: makeNextHREF(req, cursorOut) })
   }
   res.json(streams)
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -102,7 +102,7 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
 app.delete('/:id', authMiddleware({}), async (req, res) => {
   const { id } = req.params
   const stream = await req.store.get(`stream/${id}`, false)
-  if (!stream || stream.userId !== req.user.id && !(req.user.admin && req.authTokenType == 'JWT')) {
+  if (!stream || stream.deleted || stream.userId !== req.user.id && !req.isUIAdmin) {
     res.status(404)
     return res.json({ errors: ['not found'] })
   }
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -17,7 +17,7 @@ app.get('/', authMiddleware({ admin: true }), async (req, res) => {
   let cursor = req.query.cursor
   logger.info(`cursor params ${req.query.cursor}, limit ${limit}`)
 
-  const resp = await req.store.list(`stream/`, cursor, limit)
+  const resp = await req.store.list({ prefix: `stream/`, cursor, limit })
   let output = resp.data
   res.status(200)
 
@@ -40,12 +40,12 @@ app.get('/user/:userId', authMiddleware({}), async (req, res) => {
     })
   }
 
-  const streamIds = await req.store.query(
-    'stream',
-    { userId: req.params.userId },
+  const streamIds = await req.store.query({
+    kind: 'stream',
+    query: { userId: req.params.userId },
     cursor,
-    limit,
-  )
+    limit
+  })
   const streams = []
   for (let i = 0; i < streamIds.length; i++) {
     const token = await req.store.get(`stream/${streamIds[i]}`, false)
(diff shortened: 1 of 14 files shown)


criticaltv:
There are 2 incomplete sections marked as  "required" in the PR description.

Are these no longer "required", and the template can be updated?

packages/api/src/store/model.js
```diff
@@ -47,40 +47,80 @@ export default class Model {
...
-      cleanWriteOnly,
     )
 
     const ids = []
     for (let i = 0; i < keys.length; i++) {
       ids.push(keys[i].split('/').pop())
     }
 
-    return ids
+    return { data: ids, cursor: cursorOut }
```
angyangie, line 92: Hmmm seems like we're not using this cursor returned anywhere, which is a bug from previous code iterations.

This method `.query()` was originally intended to just return the presence of a `user-email` object, which would indicate the existence of a user object with that email. In looking at which `.query()` DB calls are left in our code, they're mostly in `user.js` and used for that purpose. There never really should be more than 1 `user-email` object per email, so returning first page of results is okay.

However, there are two `.query()` calls, one in `object-stores.js` `app.get(/:userId/:id)` and in `user.js` in `app.post( '/password/reset')`. In both of these calls, we should actually be returning ALL objects, not just those in the first page of results. Since we're checking for the existence of an `objectStoreId` and a `passwordResetToken`, these could potentially be found on page 2 or 3, etc of results.

Seems like we could either:
1. Have the `query()` method return all results using the cursor
2. Combine `.query()` and `queryObjects()` into one. `user-email` calls could have a limit of 1, and return that entire user object (we always make a separate DB call to retrieve that user object anyway). For the remaining two calls I mention above, we could either be okay with making some extra queries and returning all `objectStore` or `passwordResetToken` objects, when only one is necessary, OR, add an extra param to `queryObjects` that would instruct it to only return id's.

packages/api/src/store/model.js
```diff
@@ -47,40 +47,80 @@ export default class Model {
...
+        prefix,
+        cursor,
+        limit,
+      )
+
+      const documents = []
+      for (let i = 0; i < keys.length; i++) {
+        const id = keys[i].split('/').pop()
+        const doc = await this.backend.get(`${kind}/${id}`)
+        if (doc && (typeof filter !== 'function' || filter(doc))) {
```
angyangie, line 114: Nice sneaky filtering in the if statement with `filter(doc)` 😆

packages/api/src/controllers/stream.js
```diff
@@ -40,41 +42,34 @@ app.get('/user/:userId', authMiddleware({}), async (req, res) => {
...
 
 app.get('/:id', authMiddleware({}), async (req, res) => {
-  const output = await req.store.get(`stream/${req.params.id}`)
-  if (
-    !output ||
-    (output.userId !== req.user.id &&
-      !(req.user.admin && req.authTokenType == 'JWT'))
-  ) {
+  const stream = await req.store.get(`stream/${req.params.id}`)
+  if (!stream || (stream.userId !== req.user.id || stream.deleted) && !req.isUIAdmin) {
```
angyangie, line 61: Do we need `req.isUIAdmin` here? I feel like we should allow users to search for their own user object no?

packages/api/src/controllers/stream.js
```diff
@@ -103,6 +99,21 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
   res.json(doc)
 })
 
+app.delete('/:id', authMiddleware({}), async (req, res) => {
+  const { id } = req.params
+  const stream = await req.store.get(`stream/${id}`, false)
+  if (!stream || stream.userId !== req.user.id && !(req.user.admin && req.authTokenType == 'JWT')) {
```
angyangie: Should we allow anyone to delete their own streams, regardless of whether or not they're an admin? Not sure what the product requirements are here

packages/api/src/controllers/stream.test.js
```diff
@@ -287,6 +331,15 @@ describe('controllers/stream', () => {
       expect(streams[0].userId).toEqual(nonAdminUser.id)
     })
 
+    it('should delete stream', async () => {
```
angyangie, line 334: Might have forgotten the delete!

packages/api/src/controllers/stream.js
angyangie: If not, we could check for `req.isUIAdmin`

packages/api/src/store/model.js
darkdarkdragon: Any Model method we have without set `limit` will return all the data. 
Both `app.get(/:userId/:id` and `app.post( '/password/reset')` do not set limit, so they will receive all data. I don't see any problems here.

packages/api/src/controllers/stream.js
darkdarkdragon: This will allow users to see own not deleted objects and will allow admin to see any object including deleted.

packages/api/src/controllers/stream.js
darkdarkdragon: this reads like `if StreamNotFound OR NotMyStream AND NotAdmin then forbid`
this condition allows user to delete own stream and allows admin to delete any stream

added new check to not delete already deleted stream

packages/api/src/controllers/stream.test.js
darkdarkdragon: Fixed

packages/api/src/store/model.js
angyangie: Ah I see, misunderstanding on my end.

angyangie (approved):

PR #724, opened 2020-05-18
fix subgraph poll tally when delegator switches to an orchestrator that voted

adamsoffer:
This PR addresses a discrepancy in the offchain poll tally subgraph results that occurs when a delegator switches stake to an orchestrator that voted in a poll. The discrepancy is too small to notice in the explorer UI right now due to precision rounding but can be confirmed running the [poll tally audit script](https://github.com/livepeer/poll-tally-audit).

The Livepeer Canary subgraph is currently syncing with this fix. After it’s done syncing I’ll run the audit again and once it passes we can merge this PR and point to the updated subgraph.

Added an integration test for when a delegator switches to an orchestrator that voted in a poll.

**Screenshots (optional):**
Current [poll tally audit](https://github.com/livepeer/poll-tally-audit) results:
Screen Shot 2020-05-18 at 7 18 32 PM

### DIFF
--- a/packages/subgraph/src/mappings/pollTallyHandlers.ts
+++ b/packages/subgraph/src/mappings/pollTallyHandlers.ts
@@ -102,9 +102,12 @@ export function updatePollTallyOnBond(event: BondEvent): void {
     } else {
       oldDelegateVote.registeredTranscoder = false
     }
-    if (isSwitchingDelegates && oldDelegateVote.choiceID != null) {
-      oldDelegateVote.voteStake = oldDelegate.totalStake
-
+    if (isSwitchingDelegates) {
+      // if old delegate voted, update its vote stake
+      if(oldDelegateVote.choiceID != null) {
+        oldDelegateVote.voteStake = oldDelegate.totalStake
+      }
+      
       // if caller is voter, remove its nonVoteStake from old delegate
       if (voterAddress == event.params.delegator.toHex()) {
         oldDelegateVote.nonVoteStake = oldDelegateVote.nonVoteStake.minus(
@@ -147,9 +150,18 @@ export function updatePollTallyOnBond(event: BondEvent): void {
     }
 
     newDelegateVote.voter = event.params.newDelegate.toHex()
-    newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
+
+    // if switching, add stake to new delegate's nonVoteStake, otherwise update
+    // new delegate's nonVoteStake
+    if(isSwitchingDelegates) {
+      newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
+      .plus(event.params.bondedAmount)
+    } else {
+      newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
       .minus(vote.voteStake as BigInt)
       .plus(event.params.bondedAmount)
+    }
+    
     newDelegateVote.save()
 
     vote.voteStake = event.params.bondedAmount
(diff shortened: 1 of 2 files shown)


packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -102,9 +102,12 @@ export function updatePollTallyOnBond(event: BondEvent): void {
     } else {
       oldDelegateVote.registeredTranscoder = false
     }
-    if (isSwitchingDelegates && oldDelegateVote.choiceID != null) {
-      oldDelegateVote.voteStake = oldDelegate.totalStake
-
+    if (isSwitchingDelegates) {
```
yondonfu, line 105: IIUC this change is required because otherwise when a voter switches delegates and its old delegate did *not* vote yet then the old delegate's nonVoteStake would not be updated properly?

packages/subgraph/src/mappings/pollTallyHandlers.ts
```diff
@@ -147,9 +150,18 @@ export function updatePollTallyOnBond(event: BondEvent): void {
     }
 
     newDelegateVote.voter = event.params.newDelegate.toHex()
-    newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
+
+    // if switching, add stake to new delegate's nonVoteStake, otherwise update
+    // new delegate's nonVoteStake
+    if(isSwitchingDelegates) {
+      newDelegateVote.nonVoteStake = newDelegateVote.nonVoteStake
```
yondonfu, line 157: IIUC this change is required because otherwise if a voter is switching delegates then the new delegate's nonVoteStake would be incremented by the difference between the voter's bondedAmount and voteStake which would be incorrect because the voter's voteStake was never added to new delegate's nonVoteStake in the first place?

packages/subgraph/test/integration/test.js
```diff
@@ -597,6 +597,21 @@ contract('Subgraph Integration Tests', accounts => {
     await tallyPollAndCheckResult()
   })
 
+  it('correctly tallies poll after delegator moves stake to a transcoder that voted', async () => {
```
yondonfu, line 600: Ah so [this test](https://github.com/livepeer/livepeerjs/pull/724/files#diff-8a3abbaaf2296f3605d523ab21bd1c1fR522) didn't cover this case because `transcoder2` had not voted right?

yondonfu (commented):
Looks good! Just a few questions for my own personal understanding.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Exactly. We need to update its nonVoteStake in this case regardless of whether it voted.

packages/subgraph/src/mappings/pollTallyHandlers.ts
adamsoffer: Correct! This would be the first time that its voteStake gets added to the new delegate's nonVoteStake.

packages/subgraph/test/integration/test.js
adamsoffer: Exactly. If `transcoder2` had voted, that test would fail.

yondonfu (approved):
Thanks for answering my questions - LGTM!

adamsoffer:
Subgraph is synced and poll tally audit is passing again 👍

image

PR #726, opened 2020-05-20
add transaction summary modal

### LINKED ISSUE #709: Block web3 interaction buttons while an async call to a web3 provider is pending
Currently most buttons that call a web3 provider (such as voting yes/no) don't block the button from user actions nor display a spinner to indicate that the call is pending. Users can currently click the same button multiple times resulting in multiple transaction popups in metamask.

adamsoffer:
This PR introduces a "transaction summary modal" as described by [Rimble](https://rimble.consensys.design/templates/tx-summary) effectively preventing users from accidentally triggering multiple wallet prompts. In a follow up PR we can introduce other important information to this modal such as estimated fees and time as described in the Rimble guide.

Fixes #709 

**Screenshots (optional):**
tx-summary

### DIFF
--- a/packages/explorer-2.0/pages/index.tsx
+++ b/packages/explorer-2.0/pages/index.tsx
@@ -16,30 +16,21 @@ const Home = () => {
   const orchestratorsViewQuery = require('../queries/orchestratorsView.gql')
   const accountQuery = require('../queries/account.gql')
   const context = useWeb3React()
-  const { data, loading, refetch } = useQuery(orchestratorsViewQuery, {
+  const { data, loading } = useQuery(orchestratorsViewQuery, {
     pollInterval: 10000,
     ssr: false,
   })
-  const {
-    data: dataMyAccount,
-    loading: loadingMyAccount,
-    refetch: refetchMyAccount,
-  } = useQuery(accountQuery, {
-    variables: {
-      account: context?.account?.toLowerCase(),
+  const { data: dataMyAccount, loading: loadingMyAccount } = useQuery(
+    accountQuery,
+    {
+      variables: {
+        account: context?.account?.toLowerCase(),
+      },
+      pollInterval: 10000,
+      skip: !context.active,
+      ssr: false,
     },
-    pollInterval: 10000,
-    skip: !context.active,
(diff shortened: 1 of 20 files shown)


packages/explorer-2.0/pages/index.tsx
```diff
@@ -16,30 +16,21 @@ const Home = () => {
...
+      ssr: false,
     },
-    pollInterval: 10000,
-    skip: !context.active,
-    ssr: false,
-  })
-
-  // Refetch data if we detect a network change
-  useEffect(() => {
-    refetch()
```
yondonfu, line 38: Ah so we don't need this anymore because the explorer no longer supports network changes since there is a separate explorer URL per network?

yondonfu (commented):
Looks like there is a common pattern for the onClick handler whenever the tx summary modal is shown:

1. Call `writeData` using the Apollo client with `open: true` to open the tx summary modal
2. Submit the tx
3. Call `writeData` using the Apollo client with `open: false` to close the tx summary modal
4. If an error is thrown, call `writeData` using the Apollo client with `error: true` to show an error message in the tx summary modal

With the exception of [this line](https://github.com/livepeer/livepeerjs/pull/726/files#diff-00dd433e5cec94df0990ad83eb3ea784R53) that also uses the `tourOpen: false` option when calling `writeData`, looks like this pattern is always used when showing the tx summary modal. Could a helper be used here so we don't have to add Apollo client `writeData` logic whenever we submit a tx? Not sure how to handle that case where we also need to use the `tourOpen: false` option though.

packages/explorer-2.0/pages/index.tsx
adamsoffer: Yeah exactly this would refetch data upon detection of a network change but we have a dedicated explorer URL per network now.

adamsoffer:
> Could a helper be used here so we don't have to add Apollo client `writeData` logic whenever we submit a tx? 

ah yeah totally, that would be much more DRY :) Will follow up with a helper inside `utils.tsx`.

yondonfu (approved):
LGTM! :shipit:

PR #727, opened 2020-05-21
fix poll end time display

adamsoffer:
The poll end date is incorrectly displaying the timestamp associated with the latest block rather than the end block. This PR fixes that.

Screen Shot 2020-05-21 at 11 40 17 AM

### DIFF
--- a/packages/explorer-2.0/lib/createSchema.tsx
+++ b/packages/explorer-2.0/lib/createSchema.tsx
@@ -146,9 +146,7 @@ export default async () => {
               _poll?.tally?.no ? _poll?.tally?.no : '0',
             ).add(Utils.toBN(_poll?.tally?.yes ? _poll?.tally?.yes : '0'))
 
-            return Utils.toBN(totalStake)
-              .sub(totalVoteStake)
-              .toString()
+            return Utils.toBN(totalStake).sub(totalVoteStake).toString()
           },
         },
         status: {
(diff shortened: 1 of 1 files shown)


packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -146,9 +146,7 @@ export default async () => {
               _poll?.tally?.no ? _poll?.tally?.no : '0',
             ).add(Utils.toBN(_poll?.tally?.yes ? _poll?.tally?.yes : '0'))
 
-            return Utils.toBN(totalStake)
-              .sub(totalVoteStake)
-              .toString()
+            return Utils.toBN(totalStake).sub(totalVoteStake).toString()
```
adamsoffer: FYI this change was just caused by an update to the `prettier` dependency and can be ignored.

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -210,10 +210,16 @@ export default async () => {
         },
         endTime: {
           async resolve(_poll, _args, _context, _info) {
-            const blockData = await _context.livepeer.rpc.getBlock('latest')
-            return parseInt(blockData.number) > parseInt(_poll.endBlock)
-              ? blockData.timestamp
-              : null
+            const currentBlockData = await _context.livepeer.rpc.getBlock(
```
yondonfu, line 213: Looks like we resolve `endTime` [here](https://github.com/livepeer/livepeerjs/blob/69031600dd5da6e30f9bb4f966bfca966be69b69/packages/explorer-2.0/pages/voting/index.tsx#L163) and [here](https://github.com/livepeer/livepeerjs/blob/bfaec96ce0fdfc1a40215dd85b0e68bdb569f57f/packages/explorer-2.0/pages/voting/%5Bpoll%5D.tsx#L152). In both cases, we check if the poll is active and only resolve `endTime`  if the poll is inactive. So, could we just always return the timestamp of the poll end block here?

i.e.

```javascript
async resolve(...) {
    const endBlockData = await _context.livepeer.rpc.getBlock(
        _poll.endBlock,
    )
    return endBlockData.timestamp
}
```

yondonfu (changes requested):

packages/explorer-2.0/lib/createSchema.tsx
adamsoffer: It's possible to query `endTime` before a poll is over in which case `getBlock()` would throw an error because you can't pass it a future block that doesn't exist yet. So I first check to see if the the latest block is greater than the end block and if so return null.

yondonfu: Gotcha so even though we only query `endTime` in the scenarios I described in my original comment, technically `endTime` could be queried in other scenarios so we'd prefer to make sure it doesn't throw an error in those scenarios.

yondonfu (approved):
LGTM!

PR #728, opened 2020-05-21
update infura api key

adamsoffer:
Infura requested we use the same API key for all our properties so our usage is all under one account. The explorer is currently using a different api key than our `go-livepeer` client and `sdk`. Per our discussions, eventually the plan is to tie this key to the whitelist origin `*.livepeer.org`.


yondonfu (approved):
LGTM

PR #731, opened 2020-05-26
RTMP Mist ingest

### LINKED ISSUE #673: Change stream to accommodate RTMP ingest
### Vocabulary:
`stream key` and `stream id`  used interchangeably - string id that is used to uniquely identify incoming RTMP stream (used as part of the RTMP ingest URL). Generated by Livepeer API.
`playback id` - string id that is used to identify outgoing stream. Used as a part of URL used to play transcoded stream.

### Implementation:
- add parentId, playbackId, streamKey fields to the stream object.
- workflow will work this way:
  - user creates new `stream` object using `/api/stream` endpoint. Fields `playbackId` and `streamKey` are generated by API, `parentId` is empty.
  - RTMP stream hits Mist server with URL like rtmp://mist.host/live/streamKey
  - Mist sends this RTMP url to `mist-api-connector`. Connector gets `stream` object from API (by `streamKey` and creates new 'stream' inside Mist instance, setting API URL for this stream to be `http://livepeeer.api/api/stream/streamObjectId`. That way, when Mist will start transcoding it will use `http//livepeer.api/api/stream/streamObjectId/stream` endpoint to create new stream object when transcoding starts. Handler for that endpoint will create new `stream` object with it `parentId` set to `streamObjectId` and `streamKey` and `playbackId` set to be empty.
  - The API will also need to handle requests for broadcasters at `http//livepeer.api/api/stream/streamObjectId/broadcaster`

In the UI we will show objects that has set `parentId` in the separate table called `Stream Sessions`. `lastSeen` field from the most recent `Stream Session` will also be shown on the parent `stream` object.

### Changes to Model:
We want to shows most recent sessions first, so we need a way to sort results we get from database. Currently our db returns results in ascending order, so we need to add option to sort in descending order.
So we will change
`Model.query(kind, queryObj, cursor, limit, cleanWriteOnly)` to

```
Model.query({
  kind: "user",
  query: {email: "eli@iame.li"},
  cursor,
  limit,
  desc: true
});
```

---
Old task, left here for historical purposes

Mist has two options to make separate URLs for streaming and playing back.
1. In stream configuration add 'password'. Without password urls looks like these:
    - rtmp://mist.host/live/stream_name
    - http://mist.host/hls/stream_name/index.m3u8
    When password added, URLs looks like:
    - rtmp://mist.host/password/stream_name
    - http://mist.host/hls/stream_name/index.m3u8
2. Use Mist trigger that allows us to check and rewrite (for internal purposes) RTMP url - allows us to write any logic around RTMP URLs we want

Will be implementing first option as it requires less work.

Currently, when `stream` object created through API, user gives name to the stream, and automatically generated ID assigned to the stream. `name` field currently not used anywhere and id is used as stream_name in URLs.
For example, user creates steam using this request:
```json
{ "name": "test_stream", "presets": ["P240p30fps16x9"] }
```
and in response gets:
```json
{ "id": "fa644ea8-fa66-4285-83af-7e030b92ee51"}
```
So now URLs will be:
    - rtmp://mist.host/live/fa644ea8-fa66-4285-83af-7e030b92ee51
    - http://mist.host/hls/fa644ea8-fa66-4285-83af-7e030b92ee51/index.m3u8

My proposal is to use user-provided stream name as playback ID and stream object's id as password, so URLs will be:

    - rtmp://mist.host/fa644ea8-fa66-4285-83af-7e030b92ee51/test_stream
    - http://mist.host/hls/test_steam/index.m3u8

Implementation tasks:
 - [ ] Check that stream name is unique upon creation
 - [ ] Show stream name in streams list in UI
 - [ ] Change mist-controller to create proper configuration in Mist when creating streams inside Mist server

darkdarkdragon:
Allows Mist Server to be used as RTMP ingest

Stream object add fields:
- parentId - Points to parent stream object
 - streamKey - Used to form RTMP ingest URL
 - playbackId - Used to form playback URL 
 - createdByTokenName - Name of the token used to create this object

UI:
StreamTable:
 1. Allow delete streams
 2. Show streams and streams sessions in separate tables	

Manually

Fixes #673 

**Screenshots (optional):**
image

**Checklist:**


darkdarkdragon:
@angyangie  This lacks unit tests (will add them tomorrow), but otherwise it is ready for review.

wohlner:
@darkdarkdragon in the UI, can you please remove the segments column. the way that information is displayed in the different lists isn't intuitive...yet. Let's leave that out for this iteration.

darkdarkdragon:
@wohlner done

image

wohlner:
Why for stream09 is there no frame rate? Where is that information coming from? @darkdarkdragon

darkdarkdragon:
@wohlner 'rendition' configuration in our db has such fields:
```
              - width
              - name
              - height
              - bitrate
              - fps
```
all these fields specified at stream creation. What you see in the row is the `name` field of each rendition. All other fields is displayed in the tooltip:

image

wohlner:
hmm, instead remove "name" from the tooltip and in the column show `height` + `p` + `framerate` so that it is consistent in the column.

darkdarkdragon:
@wohlner done

image

packages/api/src/controllers/api-token.js
```diff
@@ -70,14 +87,15 @@ app.post(
   validatePost('api-token'),
   async (req, res) => {
     const id = uuid()
-    const userId = req.user.id
+    const userId = req.body.userId && req.user.admin ? req.body.userId : req.user.id
```
angyangie, line 90: Are we always sending a `userId` with a request body if someone is an admin? If they're not admin, we don't send?

packages/api/src/controllers/stream.js
```diff
@@ -35,24 +41,58 @@ app.get('/', authMiddleware({ admin: true }), async (req, res) => {
   res.json(output)
 })
 
-app.get('/user/:userId', authMiddleware({}), async (req, res) => {
+app.get('/sessions/:id', authMiddleware({}), async (req, res) => {
```
angyangie: Perhaps rename `:id` to `:parentId`?

packages/api/src/controllers/stream.js
```diff
@@ -75,24 +115,131 @@ app.get('/:id', authMiddleware({}), async (req, res) => {
...
+      query: { streamId },
+    })
+    if (!qres.data.length && !otherKeys.includes(streamId)) {
+      return streamId
+    }
+  }
+}
+
+app.post(
+  '/:streamId/stream',
```
angyangie, line 157: Maybe `/:streamId/streamSession` or some variation of that ? It makes it super easy to read when stream refers to only stream, and session to only session :)

packages/api/src/controllers/stream.js
```diff
@@ -75,24 +115,131 @@ app.get('/:id', authMiddleware({}), async (req, res) => {
...
+      userId: stream.userId,
+      renditions: {},
+      objectStoreId: stream.objectStoreId,
+      id,
+      createdAt,
+      parentId: stream.id,
+    })
+
+    try {
+      await req.store.create(doc)
```
angyangie, line 194: 💪💪💪

packages/api/src/controllers/stream.js
```diff
@@ -108,6 +255,30 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
...
+    return res.json({ errors: ['not found'] })
+  }
+
+  stream.isActive = req.body.active
+  stream.lastSeen = +new Date()
+  await req.store.replace(stream)
+  if (stream.parentId) {
+    const pStream = await req.store.get(`stream/${id}`, false)
+    if (pStream && !pStream.deleted) {
+      pStream.isActive = req.body.active
```
angyangie, line 271: Will there generally only be one session active at a time per parent stream?

packages/api/src/middleware/auth.js
```diff
@@ -51,6 +51,9 @@ function authFactory(params) {
     req.user = user
     req.authTokenType = authToken.startsWith('Bearer') ? 'Bearer' : 'JWT'
     req.isUIAdmin = req.user.admin && req.authTokenType == 'JWT'
+    if (tokenObject && tokenObject.name) {
+      req.tokenName = tokenObject.name
```
angyangie, line 55: Why do we need to add name here?

packages/livepeer.com/www/components/DeleteStreamModal/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    streamName: string,
+    onClose: Function,
+    onDelete: Function,
+}
+
+export default ({ streamName, onClose, onDelete }: DeleteStreamModalProps) => {
+  return (
+    <Modal onClose={onClose}>
+      <h3>Are you sure?</h3>
+      <p>Are you sure you want to delete stream "{streamName}"?</p>
```
angyangie, line 14: Don't think we need this line! Unless the Figma wasn't updated and this is what we now want :) 

https://www.figma.com/file/a0Lf5NLNrDo4A585EHOj92/API-front-end-v1?node-id=0%3A1

packages/livepeer.com/www/pages/app/stream/[id].tsx
```diff
@@ -0,0 +1,273 @@
...
+                alignItems: "center",
+                gridTemplateColumns: "10em auto",
+                width: "100%",
+                fontSize: 0
+              }}
+            >
+              <Cell>Stream name</Cell>
+              <Cell>{stream.name}</Cell>
+              <Cell>Stream key</Cell>
+              <Cell>key is here</Cell>
```
Comment (angyangie) [inline packages/livepeer.com/www/pages/app/stream/[id].tsx]: This is just a placeholder for now?

angyangie:
Some general comments from running the UI:

- /app/admin/streams spacing issues: 

Screen Shot 2020-06-08 at 3 17 49 PM

- Is that bottom sessions / details / etc border at the bottom supposed to be there?

Screen Shot 2020-06-08 at 3 22 16 PM

- Create stream buttons aren't supposed to work right? That's where my PR comes in?

- When I copy RTMP ingest URL and Playback URL, I copy `http://localhost:3004/app/admin/streams`. When I click `show secret stream key`, I get a blank. I assume that's because streams don't have those properties yet. We'll hook that up to my PR?

packages/api/src/controllers/api-token.js
darkdarkdragon: This reads 'if userId is present in the body and user is admin - then use userId from the body, otherwise use user's id from req.user.id'
Anyone can send `userId` field in the request, but we will use it only if user is admin.

packages/api/src/controllers/stream.js
darkdarkdragon: done

packages/api/src/controllers/stream.js
darkdarkdragon: This is done this way for the Mist - we're setting 'http://livepeer.host/api/streamIdHere' url as API url in the Mist - so now when Mist creates new stream it sends POST to 'http://livepeer.host/api/streamIdHere/stream' - that way we now with which stream associate this 'session'

packages/api/src/controllers/stream.js
darkdarkdragon: It should be. Though with our planned setup it will be possible to have one session in the US region and one in the Europe - but I think we can live with it.

packages/api/src/middleware/auth.js
darkdarkdragon: I'm using this name to set `createdByTokenName` field on the stream object - to track what token created what stream.

packages/api/src/controllers/stream.js
angyangie: Got it, that makes sense.

packages/api/src/middleware/auth.js
angyangie: Got it. Is that preferable to using `streamId`?

darkdarkdragon:
@angyangie fixed spacing issue:
image

packages/livepeer.com/www/pages/app/stream/[id].tsx
Comment (darkdarkdragon) [reply packages/livepeer.com/www/pages/app/stream/[id].tsx]: it was a leftover, removed

packages/api/src/middleware/auth.js
darkdarkdragon: > Got it. Is that preferable to using streamId?

Do you mean preferable to using 'tokend.id'?
We've discussed this with Eli in Discord channel - in short, as token is basically user's password we don't want to store it in any place other than token's object. In that discussion we came to conclusion that it would be good to change our token's to have two parts public and private (like tokens on Amazon). But our token name is basically public part, so I've added this tracking.

angyangie: Ah yes, I was thinking `token.id`, but wrote `streamId`. Thanks for the explanation! I like the distinction between the public part of the token and private.

angyangie (approved):

PR #739, opened 2020-06-04
Reduce unneccesary polling

adamsoffer:
The explorer currently polls data on the orchestrators and accounts views every 10 seconds to keep data fresh. Each poll interval on the orchestrators view contains about 10 network requests to infura which means if 100 people have the explorer open for 24 hours that's ~8,640,000 calls to Infura. This PR removes polling from the orchestrator view in an effort to reduce the amount of calls.

In the future, we should look into replacing polling all together with websockets.


ericxtang (approved):
LGTM

PR #745, opened 2020-06-10
Reduce explorer poll interval

adamsoffer:
The explorer currently polls account data every 10 seconds to keep data fresh. Each poll interval 
makes ~10 network requests to Infura which means if 100 people have the explorer open for 24 hours that's ~8,640,000 calls to Infura which is around what we're seeing. This PR reduces polling from 10 seconds to every 20 seconds in an effort to reduce the amount of Infura requests.

In the future, we should look into replacing polling all together with websockets. Both the hosted subgraph and Infura offer websocket support.


yondonfu (approved):
LGTM

PR #748, opened 2020-06-10
ar/sendgrid-sales-emails-livepeer.com-only

angyangie:

Sendgrid sales verified emails should only be sent when user signs up with livepeer.com.

### DIFF
--- a/packages/api/src/controllers/user.js
+++ b/packages/api/src/controllers/user.js
@@ -178,7 +178,7 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
 
     if (req.headers.host.includes('livepeer.com')) {
       try {
-        // send email verification message to user using SendGrid
+        // send sales@livepeer.org user verification message using SendGrid
         await sendgridEmail({
           email: salesEmail,
           supportAddr,
@@ -194,12 +194,7 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
           ].join('\n\n'),
         })
       } catch (err) {
-        res.status(400)
-        return res.json({
-          errors: [
-            `error sending confirmation email to ${req.body.email}: error: ${err}`,
-          ],
-        })
+        console.error(`error sending email to ${salesEmail}: error: ${err}`,)
       }
     }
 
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/user.js
```diff
@@ -176,29 +176,31 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
...
-    } catch (err) {
-      res.status(400)
-      return res.json({
-        errors: [
-          `error sending confirmation email to ${req.body.email}: error: ${err}`,
-        ],
-      })
+    if (req.headers.host.includes('livepeer.com')) {
+      try {
+        // send email verification message to user using SendGrid
```
darkdarkdragon: this comment is misleading, can we fix it?

packages/api/src/controllers/user.js
```diff
@@ -176,29 +176,31 @@ app.post('/verify', validatePost('user-verification'), async (req, res) => {
...
+          preheader: 'We have a new verified user',
+          buttonText: 'Log into livepeer',
+          buttonUrl: buttonUrl,
+          unsubscribe: unsubscribeUrl,
+          text: [
+            `User ${user.email} has signed up and verified their email with Livepeer!`,
+          ].join('\n\n'),
+        })
+      } catch (err) {
+        res.status(400)
```
darkdarkdragon: Probably we shouldn't return error to the user if we can't send our internal email.

darkdarkdragon (changes requested):

packages/api/src/controllers/user.js
angyangie: You're right. Is a `console.log` enough at this point? Do we have a better way of alerting us internally when there is an error here?

packages/api/src/controllers/user.js
angyangie: Yes!

packages/api/src/controllers/user.js
darkdarkdragon: I think it will be better to do `console.error`.
No, we don't have a better way.
I think we can try to setup some alerts in the Papertrail (reacting to the 'error' words)

angyangie: Sounds good! I added the `console.error`. Official alerts don't seem like a priority right now for that specific sales email, but definitely important to keep on our radar in the future.

darkdarkdragon (approved):

Issue #744, opened 2020-06-10
Add orchestrator price per pixel to table view


adamsoffer:
_Note: to make room for a new column we will be removing the "active" label column and instead use the status indicator on account avatars to indicate active/inactive status._

yondonfu:
Closed by #749

PR #749, opened 2020-06-12
add orchestrator prices to table and campaign view

### LINKED ISSUE #740: Add orchestrator price per pixel to campaign view

adamsoffer:
This PR adds orchestrator pricing to the orchestrator table and campaign views using the [buidl api](https://github.com/buidl-labs/livepeer-pricing-tool/tree/master/api). It also introduces pagination to the orchestrator table view for better performance (dynamically updating prices for 100 rows was sluggish and pagination was long overdue).

Notes:
- Pagination is currently set to display 10 orchestrators per page. How do we feel about that? Do we want to increase to 20?
- With a greater emphasis on price and performance now, any thoughts on sorting by total generated fees as opposed to total stake by default?

Closes #740, #744


packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -76,20 +77,29 @@ export default async () => {
...
               context: _ctx,
               info: _info,
             })
             return threeBoxSpace
           },
         },
+        price: {
+          async resolve(_transcoder, _args, _context, _info) {
+            const response = await fetch(
+              `https://livepeer-pricing-tool.com/priceHistory/${_transcoder.id}`,
```
kyriediculous: Another reason why not using this endpoint is quite pressing: it has no pagination but does index price every hour so the request size (and thus it's relation to latency) grows linearly with time. 

https://github.com/buidl-labs/livepeer-pricing-tool/blob/c7ba5408b1bd8eed70652a85be549772312f3cff/api/dataservice/dataservice.go#L131

Let's just leave pricing outside of the subgraph for now and just request it in parallel with subgraph data and manage the state in the component itself (idk if we use a state management library ? ).

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -299,3 +299,10 @@ export const getEstimatedBlockCountdown = async number => {
   const { result } = await countdownRaw.json()
   return result
 }
+
+export const expandedPriceLabels = {
```
kyriediculous, line 303: What do you think of adding a USD toggle as well ?

kyriediculous (changes requested):

packages/explorer-2.0/lib/createSchema.tsx
adamsoffer: Yeah agreed, I'll add a separate query for fetching prices then merge results client side.

packages/explorer-2.0/lib/utils.tsx
adamsoffer: Yeah definitely, I was thinking we could add a global setting/toggle for displaying anything denominated in ETH in USD across the app (earned fees, price, etc). But, I think this could be a separate PR.

packages/explorer-2.0/lib/createSchema.tsx
adamsoffer: Refactored. I ended up using a package called [`graphql-middleware`](https://github.com/prisma-labs/graphql-middleware) for transforming the `transcoder` and `transcoders` query responses to include price data as long as `price` is included in the query's selection set.

packages/explorer-2.0/lib/utils.tsx
kyriediculous: Sounds good !

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -229,5 +227,50 @@ export default async () => {
...
+        const selectionSet = Object.keys(graphqlFields(info))
+        let transcoder = await resolve(parent, args, ctx, info)
+
+        // if selection set does not include 'price', return transcoder as is, otherwise fetch and merge price
+        if (!selectionSet.includes('price')) {
+          return transcoder
+        }
+
+        const response = await fetch(
+          `https://livepeer-pricing-tool.com/priceHistory/${args.id}`,
```
kyriediculous: I opened https://github.com/buidl-labs/livepeer-pricing-tool/issues/17 to allow date ranges. That should help. 

There's also [a bug ](https://github.com/buidl-labs/livepeer-pricing-tool/issues/18)whereby the orchestrator ethereum address needs to be lowercase. 

So for now let's use the `orchestratorStats` endpoint here as well until those two issues have been addressed.

`livepeer-pricing-tool.com/orchestratorStats` - 10.9 KB - 124ms 
`ivepeer-pricing-tool.com/priceHistory/0xf4e8ef0763bcb2b1af693f5970a00050a6ac7e1b`- 18.4 KB - 449 ms

Especially since under current circumstances the entire orchestrator list info is already nearly half the response size as the price history for one orchestrator and the difference will only increase. The latency is even more significant and we only care about the latest value for now. 

Once their API is more usable we can integrate it properly.

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -288,7 +288,7 @@ export const getBlockByNumber = async number => {
   return result
 }
 
-export const getEstimatedBlockCountdown = async number => {
+export const getEstimatedBlockCountdown = async (number) => {
```
kyriediculous, line 291: new linter rules?

kyriediculous (changes requested):

packages/explorer-2.0/lib/createSchema.tsx
adamsoffer: Sounds good. Will do.

packages/explorer-2.0/lib/utils.tsx
adamsoffer: Yeah, we updated prettier to v2.0.5 last month and I noticed they updated their default rule for parentheses and single argument functions. https://prettier.io/docs/en/options.html#arrow-function-parentheses.

Should really create another PR that reruns prettier on the entire repo to account for these.

kyriediculous: Should we be using [standard.js](https://standardjs.com/) instead of prettier defaults?

adamsoffer: Strong preference for Prettier. The defaults are nearly identical but Prettier has better tooling and adoption.

kyriediculous:
Changes look good 👍  perhaps let's rebase and force push already and that will also re-run CI so we can get a staging link and clean commit history in one go.

adamsoffer:
@kyriediculous rebased and all checks pass 👍 . I'll plan on squashing + merging on Monday along with release notes for the "what's new" section.

kyriediculous (approved):
LGTM 👍

PR #750, opened 2020-06-12
ar/adding yarn.lock changes

angyangie:
- Removing the yarn.lock file, running yarn and pushing the updated yarn.lock file

**Description of issue:**
Having some issues with yarn install on livepeer.monster. Deployment failed: https://vercel.com/livepeerorg/livepeer-monster/nnmfhb57j

One notable error: The engine "node" is incompatible with this module. Expected version "10.x". Got "12.16.2"

Found that specific error and tried the solution suggested here, yarn --ignore-engines, but didn't totally build on dev either. https://github.com/nats-io/stan.js/issues/43


adamsoffer:
Nice, looks like that got rid of the "node engine" issue, but the circleci build seems to be timing out now? Perhaps try rerunning circleci build.

PR #758, opened 2020-06-15
Geolocate /ingest endpoint

darkdarkdragon:
For `/ingest` endpoint find closes server. Uses same code as `/broadcast` endpoint.

- 
- 
- 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/broadcaster.js
+++ b/packages/api/src/controllers/broadcaster.js
@@ -19,7 +19,7 @@ export const amalgamate = async (req, endpoint) => {
   const { servers } = req.region
   let responses = await Promise.all(
     servers.map(async ({ server }) => {
-      const serverRes = await fetch(`http://${server}/api/${endpoint}`, {
+      const serverRes = await fetch(`${server}/api/${endpoint}`, {
         headers: {
           authorization: req.headers.authorization,
         },
--- a/packages/api/src/controllers/broadcaster.js
+++ b/packages/api/src/controllers/broadcaster.js
@@ -15,11 +15,11 @@ export const getBroadcasterStatuses = async req => {
   return statuses
 }
 
-export const amalgamate = async req => {
+export const amalgamate = async (req, endpoint) => {
   const { servers } = req.region
   let responses = await Promise.all(
     servers.map(async ({ server }) => {
-      const serverRes = await fetch(`https://${server}/api/broadcaster`, {
+      const serverRes = await fetch(`http://${server}/api/${endpoint}`, {
         headers: {
           authorization: req.headers.authorization,
         },
@@ -44,7 +44,7 @@ export const amalgamate = async req => {
 export async function getBroadcasterHandler(req, res) {
   let broadcasters
   if (req.region && req.region.servers) {
-    broadcasters = await amalgamate(req)
+    broadcasters = await amalgamate(req, 'broadcaster')
   } else {
     broadcasters = await req.getBroadcasters(req)
   }
--- a/packages/api/src/middleware/geolocate.js
+++ b/packages/api/src/middleware/geolocate.js
@@ -1,17 +1,24 @@
 'use strict'
 
+export function addProtocol(url, protocol = 'https') {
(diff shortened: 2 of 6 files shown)


iameli (changes requested):
So our transcoding servers are going to be located at `https://chi.livepeer.com` and `https://rkv.livepeer.com`. But our ingest servers are going to be located at `rtmp://mdw.livepeer.com` and `rtmp://fra.livepeer.com`. So we're going to need a new command-line option for these, perhaps `--ingest-region`?

darkdarkdragon:
@iameli so there will own API service in the `mdw.livepeer.com` and own in the `fra.livepeer.com`?

packages/api/src/controllers/broadcaster.js
```diff
@@ -15,11 +15,11 @@ export const getBroadcasterStatuses = async req => {
   return statuses
 }
 
-export const amalgamate = async req => {
+export const amalgamate = async (req, endpoint) => {
   const { servers } = req.region
   let responses = await Promise.all(
     servers.map(async ({ server }) => {
-      const serverRes = await fetch(`https://${server}/api/broadcaster`, {
+      const serverRes = await fetch(`http://${server}/api/${endpoint}`, {
```
iameli: I assume you did this for testing, but they're https-only in production

packages/api/src/middleware/geolocate.js
```diff
@@ -1,15 +1,14 @@
 'use strict'
 
-function geoLocateFactory(params) {
+function geoLocateFactory({ first = true, region = 'region' }) {
   return async (req, res, next) => {
-    let first = params.first || true
     if (req.query.first === 'false') {
       first = false
-    } else if ((req.query.first = 'true')) {
+    } else if ((req.query.first == 'true')) {
```
iameli: Whoa, good catch, we need some linting on here...

packages/api/src/controllers/broadcaster.js
iameli: Arguably we should specify these as `INGEST_REGION__0=https://mdw.livepeer.com` so we can allow both HTTP and HTTPS and ports and stuff. But it should at least default to `https` for backwards-compatibility.

packages/api/src/parse-cli.js
```diff
@@ -133,11 +144,17 @@ export default function parseCli(argv) {
...
+            'hardcoded list of ingest points to return from /api/ingest.',
+          type: 'string',
+          default: '[]',
+        },
         supportAddr: {
           describe:
             'email address where outgoing emails originate. should be of the form name/email@example.com',
           type: 'string',
-          coerce: supportAddr => {
+          coerce: (supportAddr) => {
```
iameli, line 157: It looks like your IDE has a different version of Prettier than we're using or something, there's a few of these format changes in this PR. Could you run this and see if it makes everything consistent?

```
git diff --name-only master | xargs npx prettier --write
```

iameli (changes requested):
Solid, just polish

packages/api/src/controllers/broadcaster.js
darkdarkdragon: I've added ability to specify `--region` and `region-ingest` with or without protocol, defaulting to `https`

packages/api/src/middleware/geolocate.js
darkdarkdragon: Yes, preferable it should be typescript compiler 😄

packages/api/src/parse-cli.js
darkdarkdragon: > Could you run this and see if it makes everything consistent?

and this added more parentheses around arrowed functions 😄 
I've looked into prettier docs - adding parentheses is default option, and we not overriding it anywhere - so why you have different results?

iameli: Hah, great question, super weird. No problem then.

iameli (approved):
LGTM!

PR #760, opened 2020-06-18
api: compile .ts files for production too

iameli:
Just a tiny little change so the thing actually gets compiled for the output container


darkdarkdragon (approved):
LGTM

PR #762, opened 2020-06-18
api: add fractional FPS profile support

iameli:
Some awkwardness around the fact that the API uses camelCase and the webhook uses `snake_case`. Not too hard to handle, but perhaps we should standardize for consistency across the Livepeer ecosystem?

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -111,7 +111,7 @@ components:
                 minimum: 400
               fps:
                 type: integer
-                minimum: 1
+                minimum: 0
               fpsDen:
                 type: integer
                 minimum: 1
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -111,7 +111,10 @@ components:
                 minimum: 400
               fps:
                 type: integer
-                minimum: 0
+                minimum: 1
+              fpsDen:
+                type: integer
+                minimum: 1
         objectStoreId:
           type: string
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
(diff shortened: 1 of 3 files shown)


j0sh:
> and the webhook uses snake_case

Hmm, it uses camelCase as well - see https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L119

Just noticed that the docs are typo'd with snake_case, was that the source of confusion? https://github.com/livepeer/go-livepeer/blob/master/doc/rtmpwebhookauth.md

If there's somewhere else that uses snake_case, let me know.

iameli:
Oh, cool — I don't remember why I saw that originally but that makes things easier.

iameli:
Fixed 2ff005a30d3cde4476a924f45f3755269e5da9b0

packages/api/src/schema/schema.yaml
```diff
@@ -111,7 +111,10 @@ components:
                 minimum: 400
               fps:
                 type: integer
-                minimum: 0
```
darkdarkdragon, line 114: We've put zero here for fps passthrough.

darkdarkdragon (changes requested):

packages/api/src/schema/schema.yaml
iameli: Fixed

angyangie (approved):
LGTM!

jailuthra:
> Just noticed that the docs are typo'd with snake_case, was that the source of confusion? https://github.com/livepeer/go-livepeer/blob/master/doc/rtmpwebhookauth.md

Ah my bad, will fix the docs with the H264 profiles changes!

PR #765, opened 2020-06-22
delete button

darkdarkdragon:
- show 'Reveal your stream key and the full URL via the button above' text right of the RTMP ingest link
- disable 'Delete' button after delete

- 
- 
- 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/livepeer.com/www/pages/app/stream/[id].tsx
+++ b/packages/livepeer.com/www/pages/app/stream/[id].tsx
@@ -131,9 +131,7 @@ export default () => {
 
   const getIngestURL = (stream: Stream, showKey: boolean): string => {
     const key = showKey ? stream.streamKey : "";
-    return ingest.length
-      ? pathJoin(ingest[0].ingest, key)
-      : stream.streamKey || "";
+    return ingest.length ? pathJoin(ingest[0].ingest, key) : key || "";
   };
   const getPlaybackURL = (stream: Stream): string => {
     return ingest.length
--- a/packages/livepeer.com/www/pages/app/stream/[id].tsx
+++ b/packages/livepeer.com/www/pages/app/stream/[id].tsx
@@ -1,5 +1,4 @@
 import Link from "next/link";
-import path from "path";
 import { Spinner, Box, Button, Flex, Heading } from "@theme-ui/components";
 import Layout from "../../../components/Layout";
 import useLoggedIn from "../../../hooks/use-logged-in";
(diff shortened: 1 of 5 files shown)


darkdarkdragon:
image

packages/livepeer.com/www/pages/app/stream/[id].tsx
```diff
@@ -199,11 +201,29 @@ export default () => {
...
+                    anchor={false}
+                  />
+                ) : (
+                  <Flex
+                    sx={{ justifyContent: "flex-start", alignItems: "center" }}
+                  >
+                    <Box
+                      sx={{ minWidth: 125, fontSize: 12, paddingRight: "1em" }}
+                    >
+                      {getIngestURL(stream, false)}
```
Comment (angyangie) [inline packages/livepeer.com/www/pages/app/stream/[id].tsx:216]: Think we might need to remove this line ? I'm seeing the stream key before I hit the `Show secret stream button`. Without this line I see correct expected behavior

Screen Shot 2020-06-22 at 12 00 06 PM

packages/livepeer.com/www/pages/app/stream/new-stream.tsx
```diff
@@ -100,6 +101,7 @@ export default () => {
             <Box>Stream name</Box>
             <Box>
               <Input
+                autoFocus={true}
```
angyangie, line 104: Cool! What does this do?

packages/livepeer.com/www/pages/app/stream/[id].tsx
Comment (darkdarkdragon) [reply packages/livepeer.com/www/pages/app/stream/[id].tsx:216]: It should be here, but there was bug in `getIngestURL` function. Fixed.

packages/livepeer.com/www/pages/app/stream/new-stream.tsx
darkdarkdragon: Put input focus to the text field when 'create stream' page opens.

angyangie: Niceee!

packages/livepeer.com/www/pages/app/stream/[id].tsx
```diff
@@ -128,17 +130,16 @@ export default () => {
...
+    return ingest.length ? pathJoin(ingest[0].ingest, key) : key || "";
   };
   const getPlaybackURL = (stream: Stream): string => {
     return ingest.length
-      ? `${ingest[0].playback}/${stream.playbackId}/index.m3u8`
+      ? pathJoin(ingest[0].playback, `${stream.playbackId}/index.m3u8`)
       : stream.playbackId || "";
   };
   const tabs = getTabs(0);
+  const backLink = query.admin ? "/app/admin/streams" : "/app/user";
```
Comment (angyangie) [inline packages/livepeer.com/www/pages/app/stream/[id].tsx:142]: Really minor, and only important for admin dashboard. Didn't notice this before. If we're getting to `stream/[id]` from admin, we should probably pull `getTabs` from "../admin"; and not from "../user", so we continue seeing the same subtab menu!

angyangie (approved):
Minor comment, but other than that, LGTM !

PR #767, opened 2020-06-22
Sort columns in Admin UI

darkdarkdragon:
Sort columns in Admin UI

- 
- 
- 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/livepeer.com/www/components/AdminTokenTable/index.tsx
+++ b/packages/livepeer.com/www/components/AdminTokenTable/index.tsx
@@ -9,7 +9,7 @@ import ReactTooltip from "react-tooltip";
 
 export const UserName = ({ id, users }: { id: string; users: Array<User> }) => {
   const user = users.find(o => o.id === id);
-const tid = `tooltip-name-${id}`;
+  const tid = `tooltip-name-${id}`;
   return (
     <Box
       sx={{
@@ -26,7 +26,13 @@ const tid = `tooltip-name-${id}`;
       >
         {user ? user.email : id}
       </ReactTooltip>
-      <span data-tip data-for={tid}>{user ? user.email.includes("@") ? user.email.split("@").join("@\u{200B}") : user.email : id}</span>
+      <span data-tip data-for={tid}>
+        {user
+          ? user.email.includes("@")
+            ? user.email.split("@").join("@\u{200B}")
+            : user.email
+          : id}
+      </span>
     </Box>
   );
 };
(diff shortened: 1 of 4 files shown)


adamsoffer:
Hey @darkdarkdragon - FYI the reason the vercel build is failing is due to a breaking change in one of the explorer's npm dependencies. I froze the dependency so if you rebase on master this build should go through.

packages/livepeer.com/www/components/AdminTokenTable/index.tsx
```diff
@@ -77,6 +83,24 @@ export default ({ id }: TokenTableProps) => {
     setNewToken(null);
     setCopyTime(null);
   };
+  const sortUser = () => {
+    if (tokens) {
+      tokens.sort((a, b) => a.userId.localeCompare(b.userId));
+      setTokens([...tokens]);
```
angyangie, line 105: I'm not sure this filter is working correctly ... this is what I am seeing on dev when clicking "User":

Screen Shot 2020-06-22 at 4 39 36 PM

angyangie (commented):
Thanks for doing this @darkdarkdragon ! I wonder if it wouldn't be too much work to add a small sorting arrow icon for columns that are sortable. I knew those columns were sortable from looking at the code, but others would likely not know that. Something like this: 

image

I think sorting the `Users` page might be something we talk about to do in the future, but also one we should keep in mind, especially sorting by `created_at` date (which the user object just doesn't have yet 🤔 😅  ). Super useful for reaching out to new users of Livepeer.

adamsoffer:
Hey guys - I noticed you're implementing a lot of the same table logic that I've recently implemented on the Livepeer Explorer. I had a really good experience using the [`react-table`](https://www.npmjs.com/package/react-table) library. It offers sorting, filtering, and pagination out of the box, and is not opinionated so it's easy to customize. Thought I'd share in case y'all find it helpful.

packages/livepeer.com/www/components/AdminTokenTable/index.tsx
darkdarkdragon: It works correctly - because it sorts by `userId` field 😄 But I've changed it to sort by email (if already loaded)

darkdarkdragon:
@angyangie added arrows to headers.
I've actually wanted to add `createdAt` field to the user, but forgot 😄 Will add later.

darkdarkdragon:
@adamsoffer thanks, I'll definitively will look at the react-table when will be doing this properly - for now it was just something very quick to improve usability.

angyangie (approved):
LGTM !

PR #776, opened 2020-06-24
Add support for calling staking functions with hints

### LINKED ISSUE #742: Add support for calling staking functions with hints

Allow users to provide optional list insertion hints when using calling a function that could update the transcoder pool

- [bondWithHint()](https://github.com/livepeer/protocol/blob/streamflow/contracts/bonding/BondingManager.sol#L460) -> counterpart to `bond()`
- [unbondWithHint()](https://github.com/livepeer/protocol/blob/streamflow/contracts/bonding/BondingManager.sol#L530) -> counterpart to `unbond()`
- [rebondWithHint()](https://github.com/livepeer/protocol/blob/streamflow/contracts/bonding/BondingManager.sol#L585) -> counterpart to `rebond()`
- [rebondFromUnbondedWithHint()](https://github.com/livepeer/protocol/blob/streamflow/contracts/bonding/BondingManager.sol#L612) -> counterpart to `rebondFromUnbonded()`

In the case where a delegate is already in or joining the transcoder pool, a client can calculate the list insertion hint off-chain and provide it when calling a function. The hint can reduce the gas cost of the transaction by reducing the number of iterations in a linear search to find the correct insertion position in the pool (in the best case, the hint is the correct insertion position and no search is needed).

A hint is structured as `(newPosPrev, newPosNext)` where `newPosPrev` is the address of the previous transcoder in the pool and `newPosNext` is the address of the next transcoder in the pool.

Note that a hint may be outdated because another tx updated the transcoder pool ordering. In this case, the contract will search for the correct position using the hint.

See this sheet for some gas cost projections when using hints vs. no hints https://docs.google.com/spreadsheets/d/1mZuD2Kp6PGFxgCFx4ndS7dJ-dPCGcUfQ2QhMPiSk62c/edit#gid=229822052

The explorer currently calls staking functions without hints.

adamsoffer:
This PR adds support for calling all staking functions with hints in an effort to reduce the gas costs associated with these functions.

- Adds `bondWithHint`, `unbondWithHint`, `rebondWithHint`, and `rebondFromUnbondedWithHint` methods to the sdk
- Updates the apollo mutations in the explorer to use these new methods

Closes #742

### DIFF
--- a/packages/explorer-2.0/apollo/resolvers/Mutation.tsx
+++ b/packages/explorer-2.0/apollo/resolvers/Mutation.tsx
@@ -1,4 +1,4 @@
-import { MAX_BATCH_CLAIM_ROUNDS } from '../../lib/utils'
+import { MAX_BATCH_CLAIM_ROUNDS, EMPTY_ADDRESS, getHint } from '../../lib/utils'
 
 /**
  * Approve an amount for an ERC20 token transfer
@@ -59,22 +59,49 @@ export async function approve(_obj, _args, _ctx) {
  * @return {Promise}
  */
 export async function bond(_obj, _args, _ctx) {
-  const { to, amount } = _args
-  const gas = await _ctx.livepeer.rpc.estimateGas('BondingManager', 'bond', [
-    amount,
-    to,
-  ])
-  const txHash = await _ctx.livepeer.rpc.bondApprovedTokenAmount(to, amount, {
-    gas: gas,
-    returnTxHash: true,
-  })
+  try {
+    const {
+      amount,
+      to,
+      oldDelegateNewPosPrev,
+      oldDelegateNewPosNext,
+      currDelegateNewPosPrev,
+      currDelegateNewPosNext,
+    } = _args
+    const gas = await _ctx.livepeer.rpc.estimateGas(
+      'BondingManager',
+      'bondWithHint',
+      [
+        amount,
+        to,
+        oldDelegateNewPosPrev,
+        oldDelegateNewPosNext,
+        currDelegateNewPosPrev,
+        currDelegateNewPosNext,
+      ],
+    )
+    const txHash = await _ctx.livepeer.rpc.bondWithHint(
+      amount,
+      to,
+      oldDelegateNewPosPrev,
+      oldDelegateNewPosNext,
+      currDelegateNewPosPrev,
+      currDelegateNewPosNext,
+      {
+        gas: gas,
+        returnTxHash: true,
+      },
+    )
 
-  return {
-    gas,
-    txHash,
-    inputData: {
-      ..._args,
-    },
+    return {
+      gas,
+      txHash,
+      inputData: {
+        ..._args,
+      },
+    }
+  } catch (e) {
+    console.log(e)
   }
 }
 
@@ -160,23 +187,34 @@ export async function batchClaimEarnings(_obj, _args, _ctx) {
  * @return {Promise}
  */
 export async function unbond(_obj, _args, _ctx) {
-  const { amount } = _args
-  const gas = await _ctx.livepeer.rpc.estimateGas('BondingManager', 'unbond', [
-    amount,
-  ])
+  try {
+    const { amount, newPosPrev, newPosNext } = _args
+    const gas = await _ctx.livepeer.rpc.estimateGas(
+      'BondingManager',
+      'unbondWithHint',
+      [amount, newPosPrev, newPosNext],
+    )
 
-  const txHash = await _ctx.livepeer.rpc.unbond(amount, {
-    ..._ctx.livepeer.config.defaultTx,
-    gas,
-    returnTxHash: true,
-  })
+    const txHash = await _ctx.livepeer.rpc.unbondWithHint(
+      amount,
+      newPosPrev,
+      newPosNext,
+      {
+        ..._ctx.livepeer.config.defaultTx,
+        gas,
+        returnTxHash: true,
+      },
+    )
 
-  return {
-    gas,
-    txHash,
-    inputData: {
-      ..._args,
-    },
+    return {
+      gas,
+      txHash,
+      inputData: {
+        ..._args,
+      },
+    }
+  } catch (e) {
+    console.log(e)
   }
 }
 
@@ -186,16 +224,23 @@ export async function unbond(_obj, _args, _ctx) {
  * @return {Promise}
  */
 export async function rebond(_obj, _args, _ctx) {
-  const { unbondingLockId } = _args
-  const gas = await _ctx.livepeer.rpc.estimateGas('BondingManager', 'rebond', [
-    unbondingLockId,
-  ])
+  const { unbondingLockId, newPosPrev, newPosNext } = _args
+  const gas = await _ctx.livepeer.rpc.estimateGas(
+    'BondingManager',
+    'rebondWithHint',
+    [unbondingLockId, newPosPrev, newPosNext],
+  )
 
-  const txHash = await _ctx.livepeer.rpc.rebond(unbondingLockId, {
-    ..._ctx.livepeer.config.defaultTx,
-    gas: gas,
-    returnTxHash: true,
-  })
+  const txHash = await _ctx.livepeer.rpc.rebondWithHint(
+    unbondingLockId,
+    newPosPrev,
+    newPosNext,
+    {
+      ..._ctx.livepeer.config.defaultTx,
+      gas: gas,
+      returnTxHash: true,
+    },
+  )
 
   return {
     gas,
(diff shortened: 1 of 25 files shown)


packages/explorer-2.0/apollo/resolvers/Mutation.tsx
```diff
@@ -59,15 +59,38 @@ export async function approve(_obj, _args, _ctx) {
...
-  const txHash = await _ctx.livepeer.rpc.bondApprovedTokenAmount(to, amount, {
-    gas: gas,
-    returnTxHash: true,
-  })
+    oldDelegateNewPosPrev,
+    oldDelegateNewPosNext,
+    currDelegateNewPosPrev,
+    currDelegateNewPosNext,
+  } = _args
+  const gas = await _ctx.livepeer.rpc.estimateGas(
```
kyriediculous, line 70: Unrelated to this PR , more of a question. But on the quest for reducing RPC calls , should we remove gas estimation since all web3 providers will dry run transactions before sending them? 

They already do the gas estimation before sending the actual transaction. In the case of metamask you would get the popup screen where you could adjust things. In case of a geth node it would dry run the tx , calculate the gas and then send it.

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -299,3 +301,27 @@ export const getEstimatedBlockCountdown = async number => {
...
+    newPosNext: EMPTY_ADDRESS,
+  }
+
+  const index = transcoders.indexOf(id.toLowerCase())
+
+  // if transcoder is not in active set return
+  if (index < 0) {
+    return hint
+  } else if (index === 0) {
+    // if transcoder is the first in the active set, only set posNex
```
kyriediculous, line 317: This should be the other way around if I'm not mistaken. 

> * Given a new node with a `key`, a hint is of the form `(prevId, nextId)` s.t. `prevId` and `nextId` are adjacent in the list.
 > * `prevId` is a node with a key >= `key` and `nextId` is a node with a key <= `key`

Where key is the amount of stake

Since the list is in descending order and we start at the head `next` would mean downward and `previous` would mean back up. 



So if a node is the head, then the previous would be null 

Here's the actual implementation : https://github.com/livepeer/protocol/blob/streamflow/contracts/libraries/SortedDoublyLL.sol#L74

kyriediculous (changes requested):
Nice ! This PR will hopefully bring down the cost of staking operations a little bit when there aren't too many rounds to auto-claim for. 

Quick question, do we recalculate the `transcoders` list we pass into `getHint(id, transcoders)` to reflect what the list would look like after the staking operation completes?

packages/explorer-2.0/apollo/resolvers/Mutation.tsx
adamsoffer: > Quick question, do we recalculate the `transcoders` list we pass into `getHint(id, transcoders)` to reflect what the list would look like after the staking operation completes?

ah good catch. We need to add a `pollInterval` to the transcoders query to make sure the transcoders list is up to date after a staking operation.

adamsoffer:
> Quick question, do we recalculate the `transcoders` list we pass into `getHint(id, transcoders)` to reflect what the list would look like after the staking operation completes?

ah good catch. We need to add a pollInterval to the transcoders query to make sure the transcoders list is up to date after a staking operation.

packages/explorer-2.0/apollo/resolvers/Mutation.tsx
adamsoffer: Yeah agreed we can probably remove gas estimation and rely on web3 providers for this. I don't see any downsides to that. It won't have a huge impact on the amount of RPC calls the app makes, but good to remove erroneous code.

kyriediculous:
> We need to add a pollInterval to the transcoders query to make sure the transcoders list is up to date after a staking operation.

Sorry I more like  meant that the transcoders list needs to be updated already (for what the state on-chain , ie stake, _would_ be if the transaction succeeds). 

So we would have to 

1. Get the transcoder stake
2. add/substract stake depending on the action
3. With updated stake calculate the _would be_  position in the list
4. Use those values as hints

So under the current implementation I guess we could make a copy of the transcoders list, adjust it for the staking action about to take place and then pass it into `getHint()`. 

Alternatively allow `getHint()` to take amount and 'action'  as well and then do all that logic inside of `getHint()` which might be cleaner. 

i.e `getHint(action, amount, id, transcoders): newID`  and then prev and next would be +1 and -1 the return value respectively

adamsoffer:
> the transcoders list needs to be updated already (for what the state on-chain , ie stake, _would_ be if the transaction succeeds).

ah I misunderstood. I didn't realize we need to pass hints based on the order *after* a staking operation. Thanks for the clarification.

> i.e `getHint(action, amount, id, transcoders): {nextPos, prevPos}`

I'll use this function signature 👍

kyriediculous:
> I'll use this function signature 👍

I updated it slightly so it doesn't have to return an object and the caller can calculate previous and next positions based on the new position returned.

kyriediculous:
Just to make a small note, it is possible due to race conditions that a hint might be invalid by the time a transaction is mined. 
This would cause the function to fallback to old loop behaviour. 

I don't think there's much we should do about this though, I do wonder if RPC providers will be able to differentiate between these cases when estimating the gas cost as they dry  run the tx.

adamsoffer:
> Just to make a small note, it is possible due to race conditions that a hint might be invalid by the time a transaction is mined.
> This would cause the function to fallback to old loop behaviour.
> 
> I don't think there's much we should do about this though, I do wonder if RPC providers will be able to differentiate between these cases when estimating the gas cost as they dry run the tx.

Yeah not much we can do about this, but shouldn't be a very common occurrence and worst case is that a delegator ends up paying slightly more gas.

adamsoffer:
@kyriediculous I ended up going about this a little differently. I created a method called `simulateNewActiveSetOrder` that returns a reordered `transcoders` array based on the delegator's action and amount, then pass that updated array into the `getHint` method.

kyriediculous:
I briefly looked over it, looks good to me at first sight but I'll take a closer look in a bit !

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -368,6 +368,6 @@ export const simulateNewActiveSetOrder = ({
 
   // reorder transcoders array
   return transcoders.sort((a, b) =>
-    +Utils.fromWei(a.totalStake) > +Utils.fromWei(b.totalStake) ? 1 : -1,
+    +Utils.fromWei(a.totalStake) < +Utils.fromWei(b.totalStake) ? 1 : -1,
```
kyriediculous: Let's use the `cmp()` (compare) BN prototype instead of converting these 

```
a.cmp(b) - compare numbers and return -1 (a < b), 0 (a == b), or 1 (a > b) depending on the comparison result (ucmp, cmpn)
```

Since this already returns -1, 1 or 0 we can just do `return a.totalStake.cmp(b.totalStake)` 


Assuming this is `BN.js` and not `bignumber.js` but that should have a similar method. 

Edit: see my other comment about using BN's instead of strings here

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -325,3 +325,49 @@ export const getHint = (id, transcoders) => {
...
+  )
+
+  if (index < 0) {
+    return transcoders
+  }
+
+  if (action === 'stake') {
+    transcoders[index].totalStake = Utils.toBN(transcoders[index].totalStake)
+      .add(Utils.toBN(amount))
+      .toString()
```
kyriediculous, line 367: I would generally not advise converting bignumbers to strings outside of templating for displaying on the UI. 

If this is a string in the schema it should be a bignumber in my opinion. I guess changing this would ripple all the way back up. But when you look at my other comment as well it actually makes things cleaner.  We should avoid type juggling when not necessary. 

tl;dr BN in the logic, string on the UI

kyriediculous (changes requested):

packages/explorer-2.0/lib/utils.tsx
adamsoffer: Yeah that makes sense. In this case, however, the reason I convert it back to a string is because the original value in the array (fetched from the graph endpoint) is a string. I just converted it to a big number to do the math. Does that make sense?

kyriediculous: Is that a thegraph thing? Because I'm quite sure web3 js libraries would return a bignumber for `uint256`

adamsoffer: Yeah, the graph returns BigInt numbers as strings in response to a query.

adamsoffer: Naturally, as JSON doesn't not support BigInt.

kyriediculous: Why can't we have nice things 😞 

I'm assuming this is a limitation of the graphQL type system itself, not so much thegraph specific ? 

I'm okay with converting to BN to do the math

adamsoffer: 😞 It's a limitation of the JSON spec. JSON schema types are limited to strings, numeric types, object, array, boolean and null.

packages/explorer-2.0/lib/utils.tsx
adamsoffer: Using `a.cmp(b)` instead now 👍

kyriediculous:
Changes look good, let's rebase and resolve the conflicts 👍

PR #799, opened 2020-07-15
WIP webhooks

### LINKED ISSUE #791: Webhooks MVP: "stream_session_started"
With the webhooks feature, we add the ability for the video developer's platform to hook into all the different events -- not just if the video is ready, but the progres, and more to be decided. This will make interactivity easier and also allow the video platform to feed in events during ingest and surface them in playback.

**For the MVP of the new webhooks feature, `stream_session_started is the only requirements.**

- These should be triggered when a session starts. 
- At minimum these need to be triggered for all stream ingested via RTMP, if not all streams.
- We can use the Mist Server _triggers_ panel to set up this up. See section 2.5 in the [Mist Server Manual](https://mistserver.org/guides/MistServer_Manual_2.15.pdf).
- If `stream_session_started` gets an error, disallow all sessions for that stream.

ya7ya:
API webhooks implementation based on spec defined in #791 

- 
- 
- 

Fixes #791 

**Screenshots (optional):**

**Checklist:**


packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,40 @@ tags:
...
+      required:
+        - id
+        - name
+        - userId
+        - timestamp
+        - eventType
+        - url
+      properties:
+        id:
+          pattern: ^webhook/
```
iameli: We moved away from this architecture a little — now ids are just UUIDs. We should also have a `kind` field on here that's required to be `webhook`.

packages/api/src/schema/schema.yaml
```diff
@@ -698,3 +729,106 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/api-token'
+
+  /webhook:
```
iameli, line 740: Looks good. We should also document the `trigger` endpoint.

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+    cursor,
+    limit,
+    filter: (o) => filter1(o) && filter2(o),
+  })
+  let output = resp.data
+  res.status(200)
+
+  if (output.length > 0) {
+    res.links({ next: makeNextHREF(req, resp.cursor) })
+  } // CF doesn't know what this means
```
iameli: I think that it does know now

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+app.post('/', authMiddleware({}), validatePost('webhook'), async (req, res) => {
+  // create a webhook
+  if (!req.body || !req.body.name) {
+    res.status(422)
+    return res.json({
+      errors: ['missing name'],
+    })
+  }
+
+  const webhook = await req.store.get(`webhook/${req.body.id}`)
```
iameli: Shouldn't need this section, `validatePost('webhook')` will reject anything that contains an ID

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+  const id = uuid()
+  const createdAt = Date.now()
+
+  const doc = {
+    id,
+    userId: req.user.id,
+    kind: 'webhook',
+    name: req.body.name,
+    timestamp: createdAt,
+    eventType: 'streamStarted', // TODO, remove hardcoding once we have mutliple webhook events
```
iameli: We know from validatePost this'll be streamStarted, this can jsut be `req.body.eventType`

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+    url: req.body.url, // TODO validate this.
```
iameli: 1. Let's at least do a `new URL(req.body.url)` and reject if the URL doesn't parse
2. Let's check that the url.protocol is `http` or `https`

At trigger time:

1. Get the IP from the domain. If it's already an IP, cool. https://nodejs.org/api/dns.html#dns_dns_resolve4_hostname_options_callback
2. Check that the resolved IP isn't local: https://www.npmjs.com/package/is-local-ip

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+  params.headers = {
+    'content-type': 'application/json',
+  }
+
+  params.body = JSON.stringify(payload)
+
+  let resp = await fetch(webhook.url, params)
+  
+  if (!resp || resp.statsCode !== 200) {
```
iameli: `status`?

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  let payload = {} //TODO get stream object, sanatize it and send it over.
+
+  let params = {
+    method: 'POST',
+
+  }
+  params.headers = {
+    'content-type': 'application/json',
```
iameli: `user-agent`: `livepeer.com` maybe

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  params.body = JSON.stringify(payload)
+
+  let resp = await fetch(webhook.url, params)
+  
+  if (!resp || resp.statsCode !== 200) {
+    // no 200, no stream
+    res.status(400)
+    return res.end()
```
iameli: Let's do a `res.json()`

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+    method: 'POST',
+
+  }
+  params.headers = {
+    'content-type': 'application/json',
+  }
+
+  params.body = JSON.stringify(payload)
+
+  let resp = await fetch(webhook.url, params)
```
iameli: Let's get a timeout here... 10 seconds?

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,151 @@
...
+  // delete a specific webhook
+  res.json({})
+})
+
+
+app.post('/trigger', authMiddleware({admin: true}), async (req, res) => {
+  // TODO move this logic to /setactive endpoint
+  // triggers a webhook
+  const { id } = req.body
+  const webhook = await req.store.get(`webhook/${id}`)
```
iameli: I think we'll have to get all the webhooks for a stream, so `req.store.query({streamId: req.body.streamId})`

iameli (changes requested):

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
+    
+    // basic sanitization.
+    let sanitized = stream
+    delete sanitized.streamKey
+
+    const all = false // TODO remove hardcoding here 
+    const limit = 10 // hard limit so we won't spam endpoints, TODO , have a better adjustable limit 
+    // get a list of user defined webhooks
+    const filter1 = all ? (o) => o : (o) => !o[Object.keys(o)[0]].deleted
```
iameli: Hmm, no fun. We should figure out a way to handle this soon in the model layer.

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+
+    const resp = await req.store.list({
+      prefix: `webhook/`,
+      limit,
+      filter: (o) => filter1(o) && filter2(o),
+    })
+    let output = resp.data
+    res.status(200)
+
+    output.forEach(async (webhook) => {
```
iameli: So this is where we can block on a failed webhook — instead of the `output.forEach` we can do an `await Promise.all(output.map(async (webhook) => {`.

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+      let ip = await dns.resolve4(webhook.url)
+      let isLocal = isLocalIP(ip)
+      if (isLocal) {
+        // don't fire this webhook.
```
iameli: Good, but let's log something here so we can track down when it happened.

packages/api/src/util.js
```diff
@@ -22,3 +22,23 @@ export const shuffle = (arr) => {
     })
     .map((idx) => arr[idx])
 }
+
+
+export const fetchWithTimeout = (url, options) => new Promise((resolve, reject) => {
```
iameli, line 26: Love having this function in util.js!

packages/api/src/util.js
```diff
@@ -22,3 +22,23 @@ export const shuffle = (arr) => {
...
+  const timeout = setTimeout(() => reject('timeout'), options.timeout || 10 * 1000);
+
+  return fetch(url)
+    .then(response => {
+      clearTimeout(timeout);
+
+      if (response.status === 200) {
```
iameli: I think `fetchWithTimeout` will be more useful if it does the same thing as `fetch` and returns the response even if it's an error code. That way the two functions are totally interchangeable. We can check for success codes back up  in the webhook code

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
   stream.isActive = req.body.active
   stream.lastSeen = +new Date()
   await req.store.replace(stream)
+  
+  if (req.body.active) {
```
iameli, line 319: This section all being here is cool for now but eventually we're going to want a `triggerWebhook` function so we can do it from lots of different contexts.

iameli (changes requested):
So I think we're going to want the ability to block on a failed webhook — this would allow UGC platform operators to return an error code from a webhook and disallow the stream from happening. Nice feature, and it'll be even more important when we get to "wildcard streams" in a second.

That implies this webhook will need to happen before we save the stream "active" state.

Other than that, just a few notes :+1:

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+      } else {
+        // go ahead
+        let params = {
+          method: 'POST',
+          headers: {
+            'content-type': 'application/json',
+            'user-agent': 'livepeer.com'
+          },
+          timeout: 10 * 1000, // 10 second timeout
+          body: JSON.stringify(sanitized),
```
iameli: We're going to want a little more in the webhook body: how about this?
```
{
  event: "streamStarted",
  stream: sanitized
}
```

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,56 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    let sanitized = stream
+    delete sanitized.streamKey
+
+    const all = false // TODO remove hardcoding here 
+    const limit = 10 // hard limit so we won't spam endpoints, TODO , have a better adjustable limit 
+    // get a list of user defined webhooks
+    const filter1 = all ? (o) => o : (o) => !o[Object.keys(o)[0]].deleted
+    let filter2 = (o) => o[Object.keys(o)[0]].userId
+
+    const resp = await req.store.list({
```
iameli: I think this will not work when there are more than 10 streams in the app globally? How about
```
const resp = await req.store.query({
  kind: "webhook",
  query: {
    userId: stream.userId,
    eventType: "streamStarted"
  }

File: packages/api/src/controllers/stream.js (hunk 17 above)
Comment (ya7ya) [reply packages/api/src/controllers/stream.js:319]: Agreed, I was gonna refactor it once I'm able to test it

File: packages/api/src/controllers/stream.js (hunk 19 above)
Comment (ya7ya) [reply packages/api/src/controllers/stream.js]: Oh, I didn't know about this limitation, I'll change it to use `query`

File: packages/api/src/controllers/stream.js [hunk 20]
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
   stream.lastSeen = +new Date()
   await req.store.replace(stream)
+  
+  if (req.body.active) {
+    // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
+    // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
+    
+    // basic sanitization.
+    let sanitized = stream
+    delete sanitized.streamKey
```
iameli, line 325: This will still modify `stream` which you're presumably trying to avoid. If you do `const sanitized = {...stream}`, I think that'll have the effect you want

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+            body: {
+              event: 'streamStarted',
+              stream: JSON.stringify(sanitized),
+            }
+          }
+    
+          try {
+            console.log(`webhook ${webhook.id} firing`)
+            let resp = await fetchWithTimeout(webhook.url, params)
+            if (resp.status === 200) {
```
iameli: I think 201 and 204 are probably fine... maybe `if (resp.status >= 200 && resp.status < 300)`?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+        } else {
+          // go ahead
+          let params = {
+            method: 'POST',
+            headers: {
+              'content-type': 'application/json',
+              'user-agent': 'livepeer.com'
+            },
+            timeout: 10 * 1000, // 10 second timeout
+            body: {
```
iameli: `body` needs to be a string, you'll need to move the `JSON.stringify` up around the entire thing

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    
+          try {
+            console.log(`webhook ${webhook.id} firing`)
+            let resp = await fetchWithTimeout(webhook.url, params)
+            if (resp.status === 200) {
+              // all is good
+              console.log(`webhook ${webhook.id} fired successfully`)
+            } else {
+              // block this
+              console.log(`webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`)
```
iameli: We'll want to throw an error here to block stream start

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+            }
+          } catch (e) {
+            console.log('webhook Trigger error ', e) // TODO better logs.
```
iameli: should `throw e` here too

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,197 @@
...
+    await req.store.replace(req.body.id, webhook)
+  } catch (e) {
+    console.error(e)
+    throw e
+  }
+  res.status(200)
+  res.json({})
+})
+
+// trigger logic is in stream.js , /setactive endpoint
```
iameli: We can clean this up now, stuff like this will move to `triggerWebhook` whenever that exists

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +309,78 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    // let filter2 = (o) => o[Object.keys(o)[0]].userId
+    // const resp = await req.store.list({
+    //   prefix: `webhook/`,
+    //   limit,
+    //   filter: (o) => filter1(o) && filter2(o),
+    // })
+
+    const resp = await req.store.query({
+      kind: "webhook",
+      query: {
```
iameli: Does this work okay? I'm not positive the existing system works with more than one field, sorry about that 😅

packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,46 @@ tags:
...
+          example: webhook
+        name:
+          type: string
+        userId:
+          type: string
+          readOnly: true
+        timestamp:
+          type: string
+          readOnly: true
+        eventType:
```
iameli: It's `event` in the webhook itself and `eventType` here... I kind of like standardizing on `event`?

iameli (changes requested):
I wrote a test webhook! Check it out at https://winter-darkness-88ea.livepeer.workers.dev. It rejects all streams unless the `name` field of the stream contains the string "eli_is_cool". Live at https://winter-darkness-88ea.livepeer.workers.dev

There are a few changes to make here but I kind of suspect that we can't `req.store.query` by multiple fields right now :( That'll change next week. We can either wait until then or go back to the `req.store.list` technique you had a second ago in the interim.

packages/api/src/controllers/stream.js
ya7ya: i don't know how, but I didn't know that :facepalm: :open_mouth:

iameli (approved):
LGTM if we can get CI passing and it works with that test webhook I wrote 👍

darkdarkdragon:
@ya7ya @iameli 
> So I think we're going to want the ability to block on a failed webhook 

Can we make it optional?
Add field `blocking` to the webhook object, and make it false by default?
For people who do not need  to reject streams it would make requirements for their webhook implementation less demanding.

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,140 @@
...
+})
+
+app.get('/:id', authMiddleware({}), async (req, res) => {
+  // get a specific webhook
+  logger.info(`webhook params ${req.params.id}`)
+
+  const webhook = await req.store.get(`webhook/${req.params.id}`)
+  // return res.json(webhook)
+  if ( !webhook || webhook.deleted ||
+    (webhook.userId !== req.user.id && !req.isUIAdmin)
```
darkdarkdragon:
```suggestion
  if ( !webhook || (webhook.deleted ||
    webhook.userId !== req.user.id) && !req.isUIAdmin)
```

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,140 @@
...
+  res.json(webhook)
+})
+
+app.put('/:id', authMiddleware({}), validatePost('webhook'), async (req, res) => {
+  // modify a specific webhook
+  const webhook = await req.store.get(`webhook/${req.body.id}`)
+  if (
+    !webhook ||
+    ((webhook.userId !== req.user.id || webhook.deleted) &&
+      !(req.user.admin && !stream.deleted))
```
darkdarkdragon:
```suggestion
    ((webhook.userId !== req.user.id || webhook.deleted) &&
      !req.user.admin)
```

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,140 @@
...
+  res.status(200)
+  res.json({id: req.body.id})
+})
+
+app.delete('/:id', authMiddleware({}), async (req, res) => {
+  // delete a specific webhook
+  const webhook = await req.store.get(`webhook/${req.params.id}`)
+  
+  if ( !webhook || webhook.deleted ||
+    (webhook.userId !== req.user.id && !req.isUIAdmin)
```
darkdarkdragon:
```suggestion
  if ( !webhook || (webhook.deleted ||
    webhook.userId !== req.user.id) && !req.isUIAdmin)
```

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,140 @@
...
+
+  webhook.deleted = true
+  try {
+    await req.store.replace(webhook)
+  } catch (e) {
+    console.error(e)
+    throw e
+  }
+  res.status(200)
+  res.json({})
```
darkdarkdragon:
```suggestion
  res.status(204)
  res.end()
```

packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,46 @@ tags:
...
     description: Changes user admin status
+  - name: webhook
+    description: Webhooks endpoints
 
 components:
   schemas:
+    webhook:
+      type: object
+      required:
+        - id
```
darkdarkdragon: shouldn't be required because `validatePost('webhook')` will reject without this field

packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,46 @@ tags:
...
+        - name
+        - timestamp
```
darkdarkdragon: shouldn't be required

packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,46 @@ tags:
...
+          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+        kind:
+          type: string
+          readOnly: true
+          example: webhook
+        name:
+          type: string
+        userId:
+          type: string
+          readOnly: true
```
darkdarkdragon, line 47:
```suggestion
          readOnly: true
          index: true
```

packages/api/src/schema/schema.yaml
```diff
@@ -20,9 +20,46 @@ tags:
...
+        timestamp:
```
darkdarkdragon: why not `createdAt`, like in other objects?

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,140 @@
...
+
+  const doc = {
+    id,
+    userId: req.user.id,
+    kind: 'webhook',
+    name: req.body.name,
+    timestamp: createdAt,
+    event: req.body.event,
+    url: req.body.url, // TODO validate this. 
+    deleted: false
```
darkdarkdragon: Don't add `deleted` field here - it is being exposed to the user, we don't want it. Or you should sanitise it from every place where object is returned to the user.

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +313,85 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+          console.log(`webhook ${webhook.id} resolved to a localIP, url: ${webhook.url}, resolved IP: ${ip}`)
+        } else {
+          // go ahead
+          let params = {
+            method: 'POST',
+            headers: {
+              'content-type': 'application/json',
+              'user-agent': 'livepeer.com'
+            },
+            timeout: 10 * 1000, // 10 second timeout
```
darkdarkdragon: I think 10 seconds is too much for blocking webhook at the stream start

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +313,85 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+            body: JSON.stringify({
+              event: 'streamStarted',
```
darkdarkdragon:
```suggestion
              event: webhook.event,
```

packages/api/src/util.js
```diff
@@ -22,3 +22,25 @@ export const shuffle = (arr) => {
     })
     .map((idx) => arr[idx])
 }
+
+
+export const fetchWithTimeout = (url, options) => new Promise((resolve, reject) => {
+  const timeout = setTimeout(() => reject('timeout'), options.timeout || 10 * 1000);
+
+  return fetch(url)
```
darkdarkdragon:
```suggestion
  return fetch(url, options)
```

packages/api/src/util.js
```diff
@@ -22,3 +22,25 @@ export const shuffle = (arr) => {
...
+    .then(response => {
+      clearTimeout(timeout);
+
+      return resolve(response)
```
darkdarkdragon, line 38: I think it would be good not to resolve promise if already was rejected

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +313,85 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+        } else {
+          // go ahead
+          let params = {
+            method: 'POST',
+            headers: {
+              'content-type': 'application/json',
+              'user-agent': 'livepeer.com'
+            },
+            timeout: 10 * 1000, // 10 second timeout
+            body: JSON.stringify({
```
darkdarkdragon: shouldn't we include webhook's id in the body?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +313,85 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+          urlObj = parseUrl(webhook.url)
+          ip = await resolver.resolve4(webhook.url)
+        } catch (e) {
+          console.error('error: ', e)
+        }
+        console.log('resolvedIP: ', ip)
+        // let isLocal = false
+        let isLocal = isLocalIP(ip)
+        if (isLocal) {
+          // don't fire this webhook.
```
darkdarkdragon: why? is it for debugging only?

packages/api/src/controllers/stream.js
```diff
@@ -5,13 +5,17 @@ import Router from 'express/lib/router'
...
 import wowzaHydrate from './wowza-hydrate'
+import { fetchWithTimeout } from '../util'
 import { makeNextHREF, trackAction } from './helpers'
 import { generateStreamKey } from './generate-stream-key'
 import { geolocateMiddleware } from '../middleware'
 import { getBroadcasterHandler } from './broadcaster'
+import { util } from 'prettier'
+const { Resolver } = require('dns').promises;
+const resolver = new Resolver();
+resolver.setServers(['8.8.8.8', '8.8.4.4']);
```
darkdarkdragon: this is for debugging, right?

packages/api/src/controllers/stream.js
```diff
@@ -5,13 +5,17 @@ import Router from 'express/lib/router'
 import logger from '../logger'
 import uuid from 'uuid/v4'
 import wowzaHydrate from './wowza-hydrate'
+import { fetchWithTimeout } from '../util'
 import { makeNextHREF, trackAction } from './helpers'
 import { generateStreamKey } from './generate-stream-key'
 import { geolocateMiddleware } from '../middleware'
 import { getBroadcasterHandler } from './broadcaster'
+import { util } from 'prettier'
```
darkdarkdragon: what is this?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +313,85 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    // basic sanitization.
+    let sanitized = {...stream}
+    delete sanitized.streamKey
+
+    const all = false // TODO remove hardcoding here 
+    const limit = 100 // hard limit so we won't spam endpoints, TODO , have a better adjustable limit 
+    const filter1 = all ? (o) => o : (o) => !o[Object.keys(o)[0]].deleted
+    let filter2 = (o) => o[Object.keys(o)[0]].userId
+    // TODO: change this to query once query can handle multiple query values.
+    const resp = await req.store.list({
```
darkdarkdragon: I think it would be better:
1. Refactor getting list of webhooks to separate function and make it something like this:
```js
function getWebhooks(store, userId, event) {
  let filter = (o) => !o.deleted && o.event === event
  const { data } = await store.queryObjects({
    kind: 'webhook',
    query: { userId },
    limit: 100,
    filter
  })
  return data
}
```

darkdarkdragon (changes requested):

packages/api/src/controllers/stream.js
ya7ya: this is to make sure the webhook's url doesn't resolve to a local IP

packages/api/src/controllers/stream.js
ya7ya: you mean the resolver ? it's to make sure the url is legit and accessible.

this isn't currently working, I'm actively debugging why.

packages/api/src/controllers/stream.js
darkdarkdragon: Ok, I've got it. But maybe we can allow to use local IPs for admins? It's much more convenient to test locally 😄

packages/api/src/controllers/stream.js
darkdarkdragon: > you mean the resolver ?

Yes, resolver. It's not good to change used dns servers, so - is resolver used only for debugging?

> this isn't currently working, 

what's not working?

ya7ya: all dns queries are returning ENOTFOUND error, or timing out.

packages/api/src/controllers/stream.js
iameli: Yeah, I was figuring there was an attack something like this:

1. Make a webhook for http://10.0.0.1:80
2. Stream in. Depending on the content/timing of the error, you know if there's a server there
3. Make a webhook for http://10.0.0.2:80 and repeat

You can do all of that really fast in parallel to do a scan of our entire internal network!

packages/api/src/controllers/stream.js
```diff
@@ -298,7 +304,7 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
   res.json(doc)
 })
 
-app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
+app.put('/:id/setactive', authMiddleware({admin: true}), async (req, res) => {
```
darkdarkdragon: `{admin: true}` requires user to be logged in throw JWT - which is not the case for the `mist-api-controller` calling this method. We really need to improve authMiddleware and change `admin` arg to `UIAdmin` and add new arg such as `justAdmin` 😃

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    // basic sanitization.
+    let sanitized = {...stream}
+    delete sanitized.streamKey
+
+    const all = false // TODO remove hardcoding here 
+    const limit = 100 // hard limit so we won't spam endpoints, TODO , have a better adjustable limit 
+    const filter1 = all ? (o) => o : (o) => !o[Object.keys(o)[0]].deleted
+    let filter2 = (o) => o[Object.keys(o)[0]].userId
+    // TODO: change this to query once query can handle multiple query values.
+    const resp = await req.store.list({
```
darkdarkdragon: can we use `getWebhooks` here?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    let output = resp.data
+    res.status(200)
+
+    await Promise.all(
+      output.map(async (webhookObj, key) => {
+        let webhook = webhookObj[Object.keys(webhookObj)[0]] // webhookObj has 1 key.
+        console.log('webhook: ', webhook)
+        let ips, urlObj, isLocal
+        try {
+          urlObj = parseUrl(webhook.url)
```
darkdarkdragon: `parseUrl` by itself can throw error. I think we should try to parse url in POST/PUT /api/webhook endpoint

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+          ips = await resolver.resolve4(urlObj.host)
```
darkdarkdragon: Can we move this check to POST/PUT /api/webhook endpoint?
@iameli what do you think?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+        } catch (e) {
+          console.error('error: ', e)
+        }
+
+        // This is mainly useful for local testing
+        if (req.query.allowLocalWebhooks) {
```
darkdarkdragon: This will allow anyone to skip local check?
Can we just skip local check for admin? Without introducing new query param?

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+          isLocal = false
+        } else {
+          try {
+            isLocal = isLocalIP(ips[0])
```
darkdarkdragon:
```suggestion
if (ips && ips.length) {
            isLocal = isLocalIP(ips[0])
}
```

`ips` can be empty, so without check it will throw useless indexing error

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+        } else {
+          console.log('preparing to fire webhook ', webhook.url)
+          // go ahead
+          let params = {
+            method: 'POST',
+            headers: {
+              'content-type': 'application/json',
+              'user-agent': 'livepeer.com'
+            },
+            timeout: WEBHOOK_TIMEOUT, // 10 second timeout
```
darkdarkdragon: it is not 10 seconds anymore, I think it would be better to just remove comment

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,92 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    const resp = await req.store.list({
+      prefix: `webhook/`,
+      limit,
+      filter: (o) => filter1(o) && filter2(o),
+    })
+    
+    let output = resp.data
+    res.status(200)
+
+    await Promise.all(
```
darkdarkdragon: Result of this call is not used anywhere, and hence `/setactive` endpoint will always return success, and stream can't be blocked by webhook

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,136 @@
...
+
+  if (output.length > 0) {
+    res.links({ next: makeNextHREF(req, resp.cursor) })
+  }
+  output = output.map((o) => o[Object.keys(o)[0]])
+  res.json(output)
+})
+
+app.post('/', authMiddleware({}), validatePost('webhook'), async (req, res) => {
+  // create a webhook
```
darkdarkdragon: comments not needed?

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,136 @@
+import { parse as parseUrl } from 'url'
+import { authMiddleware } from '../middleware'
+import { validatePost } from '../middleware'
+import Router from 'express/lib/router'
+import logger from '../logger'
+import uuid from 'uuid/v4'
+import { makeNextHREF, trackAction } from './helpers'
+import fetch from 'isomorphic-fetch'
+import dns from 'dns'
+import isLocalIP from 'is-local-ip'
```
darkdarkdragon: not used

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,136 @@
...
+  
+  let output = await getWebhooks(req.store, req.user.id, event)
+  res.status(200)
+
+  if (output.length > 0) {
+    res.links({ next: makeNextHREF(req, resp.cursor) })
+  }
+  output = output.map((o) => o[Object.keys(o)[0]])
+  res.json(output)
+})
```
darkdarkdragon:
```suggestion
async function getWebhooks(store, userId, event, limit = 100, cursor = undefined, includeDeleted = false) {
  let filter = (o) => (!o.deleted || includeDeleted) && (!event || o.event === event)
  const resp = await store.queryObjects({
    kind: 'webhook',
    query: { userId },
    limit,
    cursor,
    filter
  })
  return resp
}

const app = Router()

app.get('/', authMiddleware({}), async (req, res) => {
  let { limit, cursor, all, event } = req.query
  logger.info(`cursor params ${cursor}, limit ${limit} all ${all}`)
  
  const resp = await getWebhooks(req.store, req.user.id, event, limit, cursor, !!all && req.user.admin)
  res.status(200)

  if (output.length > 0) {
    res.links({ next: makeNextHREF(req, resp.cursor) })
  }
  res.json(resp.data)
})
```

something like this (not tested)

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,136 @@
...
+  const createdAt = Date.now()
+
+  const doc = {
+    id,
+    userId: req.user.id,
+    kind: 'webhook',
+    name: req.body.name,
+    createdAt: createdAt,
+    event: req.body.event,
+    url: req.body.url,
```
darkdarkdragon, line 49: at least do `url.parse`?
and it would be good to check that url has schema defined and it is http or https

darkdarkdragon (changes requested):

packages/api/src/controllers/stream.js
ya7ya: @darkdarkdragon we check every time to make sure the user didn't create a proper webhook then later changed the DNS record before triggering it.

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,98 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+    // basic sanitization.
+    let sanitized = {...stream}
+    delete sanitized.streamKey
+
+    const all = false // TODO remove hardcoding here 
+    const limit = 100 // hard limit so we won't spam endpoints, TODO , have a better adjustable limit 
+    
+    let webhooks = await getWebhooks(req.store, req.user.id, 'streamStarted')
+    let output = webhooks.data
+    console.log('output : ', output)
```
iameli: can clean this up

packages/api/src/controllers/stream.js
```diff
@@ -309,6 +315,98 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+              console.log(`webhook ${webhook.id} firing`)
+              let resp = await fetchWithTimeout(webhook.url, params)
+              if (resp.status >= 200 && resp.status < 300) { // 2xx requests are cool.
+                // all is good
+                console.log(`webhook ${webhook.id} fired successfully`)
+                return true
+              } else {
+                // block this
+                console.error(`webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`)
+                // throw new Error(`webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`)
```
iameli: Think this should throw, yeah?

packages/api/src/controllers/webhook.js
```diff
@@ -0,0 +1,142 @@
...
+import Router from 'express/lib/router'
+import logger from '../logger'
+import uuid from 'uuid/v4'
+import { makeNextHREF, trackAction, getWebhooks } from './helpers'
+
+const app = Router()
+
+app.get('/', authMiddleware({}), async (req, res) => {
+  let { limit, cursor, all, event } = req.query
+  logger.info(`cursor params ${cursor}, limit ${limit} all ${all}`)
```
iameli: cleanup

packages/api/src/util.js
```diff
@@ -22,3 +22,25 @@ export const shuffle = (arr) => {
...
+      //   return resolve(response);
+      // }
+
+      // return reject(response);
+    }, rejectReason => {
+      clearTimeout(timeout);
+
+      return reject(rejectReason);
+    });
+});
```
iameli: Just added a little check to see if it already timed out:
```suggestion


export const fetchWithTimeout = (url, options) => new Promise((resolve, reject) => {
  let timeout = setTimeout(() => {
    timeout = null;
    reject('timeout');
  }, options.timeout || 10 * 1000);

  return fetch(url, options)
    .then(response => {
      if (timeout === null) {
        // already timed out
        return;
      }
      clearTimeout(timeout);

      return resolve(response)
      // return a response regardless of the status code.
      // if (response.status === 200) {
      //   return resolve(response);
      // }

      // return reject(response);
    }, rejectReason => {
      if (timeout === null) {
        // already timed out
        return;
      }
      clearTimeout(timeout);

      return reject(rejectReason);
    });
});
```

iameli (changes requested):
A few one-line cleanups and I left a suggestion for `fetchWithTimeout`

Feel free to move over and merge when CI on the other repo passes :+1:

PR #821, opened 2020-08-18
fix pagination

### LINKED ISSUE #820: the orchestrators list view is displaying 11 pages instead of 10
There should only be 10 pages on the orchestrators view. Navigating to the 11th throws an error.

adamsoffer:
This PR fixes a bug in the the orchestrator pagination component caused by [the pricing api returning unregistered orchestrators](https://github.com/buidl-labs/livepeer-pricing-tool/issues/20). Until a fixed is deployed to that API, this PR discards those invalid orchestrators.

Closes #820

### DIFF
--- a/packages/explorer-2.0/pages/index.tsx
+++ b/packages/explorer-2.0/pages/index.tsx
@@ -21,7 +21,6 @@ const Home = () => {
     orderDirection: 'desc',
     where: {
       status: 'Registered',
-      id_not: '0x0000000000000000000000000000000000000000', // subgraph indexes this as a transcoder for some reason so need to filter it out
     },
   }
   const { data, loading } = useQuery(orchestratorsViewQuery, {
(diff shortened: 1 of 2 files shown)


packages/explorer-2.0/pages/index.tsx
```diff
@@ -21,7 +21,6 @@ const Home = () => {
     orderDirection: 'desc',
     where: {
       status: 'Registered',
-      id_not: '0x0000000000000000000000000000000000000000', // subgraph indexes this as a transcoder for some reason so need to filter it out
```
kyriediculous, line 24: I still see a null address transcoder in the subgraph query , shouldn't we keep this? 

LGTM Otherwise 👍

adamsoffer: @kyriediculous yeah that null address used to get indexed as a transcoder in the subgraph for some odd reason but that's no longer the case so I removed that filter.

kyriediculous: Are you sure? I just queried the subgraph and it's the first result: 

image

adamsoffer: ah yeah you're right it's still getting indexed but its status is null so its not returned by the above query.

kyriediculous: Okay no problem then ! 👍

PR #826, opened 2020-08-21
remove unnecessary contract call from reward handler

adamsoffer:
This PR removes an unnecessary contract call (`getDelegator`) from the subgraph's reward handler.


kyriediculous (approved):
:LGTM:

PR #840, opened 2020-10-09
add wallet connect

adamsoffer:
This PR adds WalletConnect as a wallet option so users can stake with Rainbow, Trust Wallet, Argent, etc.

**Screenshots (optional):**
image


packages/explorer-2.0/components/Approve/index.tsx
```diff
@@ -19,12 +19,16 @@ export default ({ account, banner = true }) => {
 
   const onClick = () => {
     initTransaction(client, async () => {
-      await approve({
-        variables: {
-          type: 'bond',
-          amount: MAXIUMUM_VALUE_UINT256,
-        },
-      })
+      try {
```
kyriediculous, line 22: I assume this try/catch is lingering here from testing the integration?

packages/explorer-2.0/lib/connectors.tsx
```diff
@@ -31,8 +32,15 @@ export const Fortmatic = new FortmaticConnector({
 })
 
 // mainnet only
-export const walletlink = new WalletLinkConnector({
+export const walletLink = new WalletLinkConnector({
```
kyriediculous, line 35: What do you think of in the future making fixups along the way a seperate PR so that feature improvements/additions diffs can be kept to the relevant changes?

kyriediculous (changes requested):
Not an expert on wallet connect but the integration looks good to me, just a few minor comments.

packages/explorer-2.0/components/Approve/index.tsx
adamsoffer: ah yes this was from testing the integration. I can remove 👍

packages/explorer-2.0/lib/connectors.tsx
adamsoffer: yeah this var name change is sort of unrelated to the PR. Agree in the future these types of fixups can be separated in another PR 👍

adamsoffer:
@kyriediculous thanks for the review. I'm testing this integration with Rainbow wallet and Trust wallet but noticed transactions are failing and the error message is not very helpful. I'll continue to test and see what I find.

PR #843, opened 2020-10-28
fix rebondFromUnbonded

kyriediculous:
Fixes a bug whereby we call rebondFromUnbonded with hints instead of rebondFromUnbondedWithHint , which makes the abi encoder throw an error


adamsoffer (approved):
LGTM

PR #847, opened 2020-11-09
UI updates

adamsoffer:
This PR implements UI updates/polish per [this figma design](https://www.figma.com/file/wpiGtVzSQn1mf4QiUdH7pB/Explorer?node-id=309%3A3).

- Removes the search bar page in favor of a search bar in the orchestrators page
- Removes the staking widget from the orchestrators view to avoid the optics around favoring the default top orchestrator by stake
- Updates the data table style
- Updates Apollo Client dependency to 3.0.0 and other misc dependency updates
- Refactors the way we handle data fetching on the home page


kyriediculous:
The table on the staking tab doesn't seem to respect the 775px breakpoint,  then again I'm not sure if we should omit certain columns on mobile or not, I guess we should because that's what the mobile view does. 

image

Looks good otherwise

adamsoffer:
> The table on the staking tab doesn't seem to respect the 775px breakpoint.

Good catch. Will push up a fix for that breakpoint.

kyriediculous (approved):
Really like this new look 👍

PR #850, opened 2020-11-18
add performance leaderboard view

adamsoffer:
This PR builds on the new "UI polish" PR and implements the orchestrator performance component, introducing it as a subview within the orchestrator list component. 

**Screenshots (optional):**
Screen Shot 2020-11-18 at 3 51 29 PM

### DIFF
--- a/packages/explorer-2.0/components/Orchestrators/PerformanceTable.tsx
+++ b/packages/explorer-2.0/components/Orchestrators/PerformanceTable.tsx
@@ -255,9 +255,6 @@ const PerformanceTable = ({ data: { currentRound, transcoders }, region }) => {
                       textTransform: 'uppercase',
                     }}
                     align="right"
-                    {...column.getHeaderProps(
-                      column.getSortByToggleProps({ title: '' }),
-                    )}
                     key={i}
                   >
                     <Flex
@@ -270,6 +267,9 @@ const PerformanceTable = ({ data: { currentRound, transcoders }, region }) => {
                         sx={{
                           fontSize: 10,
                         }}
+                        {...column.getHeaderProps(
+                          column.getSortByToggleProps({ title: '' }),
+                        )}
                       >
                         <span>
                           {column.isSorted ? (
@@ -405,7 +405,7 @@ const PerformanceTable = ({ data: { currentRound, transcoders }, region }) => {
   )
   function renderTooltip(title) {
     switch (title) {
-      case 'Success Rate':
+      case 'Success Rate (%)':
         return (
           <>
             <ReactTooltip
@@ -414,6 +414,8 @@ const PerformanceTable = ({ data: { currentRound, transcoders }, region }) => {
               place="bottom"
               type="dark"
               effect="solid"
+              delayHide={200}
+              delayUpdate={500}
             />
             <Help
               data-tip="The percentage of video segments sent by a broadcaster that are successfully transcoded. See the FAQ for more details on how this metric is calculated."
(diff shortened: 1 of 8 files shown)


kyriediculous:
If a column cell contains a value that is 0 , the value should still be displayed, it seems we currently display it as an empty cell. 

We should distinguish between a score of 0 (meaning the test was a complete failure) and no results for a particular O.

kyriediculous:
What do you think of putting some more focus on the "total score" column for example making it bold and putting it in the livepeer green color ? 

Also I'm wondering if we can do something to emphasise that a higher score is better.

kyriediculous:
The page has an added `OrchestratorsTable` query parameter in its URL , could we add that for region and the time period as well you think? This would allow to link to the daily and weekly leaderboards for each regional option as well. It might come in handy in the future, WDYT ?

kyriediculous:
The column that contains the latency scores currently is actually the average latency (in seconds), see t[his SQL query](https://github.com/livepeer/leaderboard-serverless/blob/f78e8c65ace4fa2b42119a35ac628341c8646981/postgres/postgres.go#L47).  

I'm wondering whether we should change this to be latency score or keep it as latency in seconds or milliseconds. Probably latency _score_ is better rather than latency itself due to it being scaled uniformly. Although latency in seconds or ms seems more informative overall. WDYT ? 

cc @yondonfu

yondonfu:
> I'm wondering whether we should change this to be latency score or keep it as latency in seconds or milliseconds. Probably latency score is better rather than latency itself due to it being scaled uniformly. Although latency in seconds or ms seems more informative overall. WDYT ?

I think we should use the latency score (defined as segment duration / round trip time) to emphasize its importance as a metric since it can be used as a standalone data point while absolute latency generally needs to be accompanied by the segment duration in order for it to be used as a data point.

kyriediculous:
@adamsoffer I added a commit, https://github.com/livepeer/livepeerjs/pull/850/commits/66e6823901d1eb28d99a6b901d0552989f78a327, to change `avg_latency` to `round_trip_score` in alignment with the [API changes](https://github.com/livepeer/leaderboard-serverless/commit/89ea23950570270d0f0f6d5dd245407cf2b91069).

kyriediculous:
Since the latest changes and the implementation of the [leaderboard scoring framework ](https://www.notion.so/livepeer/Leaderboard-Score-Framework-a420c0e9b6e4408b81cf0d9ffcd9d40e), the total score is now on a scale of 0-1. 

While for the API this looks okay, for a front-end it might be better to multiply this by 100 to have a range of 0-100 for better readability.

kyriediculous:
@adamsoffer I just redeployed a bunch of changes to the API and I think sorting of the table might be inconsistent

In this case Livepeer Studio has a total score of 0.05 and it's the only orchestrator to have a score higher than 0 (tests are still running) 

Yet it doesn't show up in first place, but it shows as 5th to last when sorting by ascending. 

Descending view: 
image


Ascending view: 
image

packages/explorer-2.0/components/Orchestrators/PerformanceTable.tsx
```diff
@@ -0,0 +1,539 @@
...
+        filter: 'fuzzyText',
+        Filter: DefaultColumnFilter,
+        mobile: true,
+        sortType: (rowA, rowB, columnID) => {
+          let a = getRowValueByColumnID(rowA, columnID)
+          let b = getRowValueByColumnID(rowB, columnID)
+          let aThreeBoxSpace = getRowValueByColumnID(rowA, 'threeBoxSpace')
+          let bThreeBoxSpace = getRowValueByColumnID(rowB, 'threeBoxSpace')
+          let rowAIdentity = aThreeBoxSpace?.name ? aThreeBoxSpace?.name : a
+          let rowBIdentity = bThreeBoxSpace?.name ? bThreeBoxSpace?.name : b
```
kyriediculous: I think we can combine L73, L75 and L77 into one line (and do the same for row B)

```suggestion
          let rowAIdentity =  getRowValueByColumnID(rowA, 'threeBoxSpace')?.name || getRowValueByColumnID(rowA, columnID)
          let rowBIdentity = getRowValueByColumnID(rowB, 'threeBoxSpace')?.name || getRowValueByColumnID(rowB, columnID)
```

adamsoffer:
> I just redeployed a bunch of changes to the API and I think sorting of the table might be inconsistent

Looking into this now.

kyriediculous (changes requested):

packages/explorer-2.0/components/Orchestrators/PerformanceTable.tsx
```diff
@@ -0,0 +1,539 @@
...
+    [region],
+  )
+
+  function getRowValueByColumnID(row, columnID) {
+    return row.values[columnID]
+  }
+
+  function compareBasic(a, b) {
+    return a === b ? 0 : a > b ? 1 : -1
+  }
```
kyriediculous, line 131:
```suggestion
  function compareBasic(a, b) {
   a = a || 0
   b = b || 0
    return a === b ? 0 : a > b ? 1 : -1
  }
```

I think this should fix inconsistent sorting

kyriediculous (changes requested):

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -303,8 +313,50 @@ const Index = async () => {
...
+                  mdw: metrics[key].MDW?.success_rate,
+                  sin: metrics[key].SIN?.success_rate,
+                },
+                roundTripScores: {
+                  global: avg(metrics[key], 'round_trip_score'),
+                  fra: metrics[key].FRA?.round_trip_score,
+                  mdw: metrics[key].MDW?.round_trip_score,
+                  sin: metrics[key].SIN?.round_trip_score,
+                },
+              })
```
kyriediculous: @adamsoffer I deployed the v1.0.0 leaderboard API as mentioned in an earlier comment

Some of the breaking changes include 
- score is now scaled between 0 and 1 
- round_trip_score is now scaled between 0 and 1
- success rate is no longer returned as the percentage but the decimal (also a range between 0 and 1) 

As per earlier suggestions we would have to make following adjustments:

- Multiply score, round_trip_score and success_rate by 100 for display. 

There might be a more optimal way to do this though

```suggestion
              performanceMetrics.push({
                id: key,
                scores: {
                  global: avg(metrics[key], 'score')*100,
                  fra: metrics[key].FRA?.score*100,
                  mdw: metrics[key].MDW?.score*100,
                  sin: metrics[key].SIN?.score*100,
                },
                successRates: {
                  global: avg(metrics[key], 'success_rate')*100,
                  fra: metrics[key].FRA?.success_rate*100,
                  mdw: metrics[key].MDW?.success_rate*100,
                  sin: metrics[key].SIN?.success_rate*100,
                },
                roundTripScores: {
                  global: avg(metrics[key], 'round_trip_score')*100,
                  fra: metrics[key].FRA?.round_trip_score*100,
                  mdw: metrics[key].MDW?.round_trip_score*100,
                  sin: metrics[key].SIN?.round_trip_score*100,
                },
              })
```

adamsoffer: Thanks. Just pushed up.

packages/explorer-2.0/components/Orchestrators/PerformanceTable.tsx
adamsoffer: done

kyriediculous:
@adamsoffer as discussed in discord 50f3bdc adds scales/denomination to the table headers for the performance table. If you feel there's a better way to do this feel free to drop this commit.

kyriediculous:
@adamsoffer "0" results are not shown explicitly in the table 

There are a few orchestrators with "0" results : https://leaderboard-serverless.vercel.app/api/aggregated_stats

These used to show though IIRC.

adamsoffer:
> @adamsoffer "0" results are not shown explicitly in the table
> 
> There are a few orchestrators with "0" results : https://leaderboard-serverless.vercel.app/api/aggregated_stats
> 
> These used to show though IIRC.

Ah yeah I know why. Pushing up a fix now.

yondonfu:
Made a small tooltip copy in 786dec4 to better align with the descriptions used in the [FAQ](https://github.com/livepeer/docs/pull/64/files#diff-a9952768ba85e60ad7bf52afeb0cd35451621f0911c64cca76be00f8cf4ea275R66) (for example noting that its an average).

kyriediculous (approved):
LGTM , a little cleanup of the commit history might be nice to get rid of the "fixes" commit messages.

PR #854, opened 2020-11-30
convert subgraph values from `BigInt` to `BigDecimal`

adamsoffer:
This PR converts values in the subgraph from `BigInt` to `BigDecimal` so that values are easier to read without having to move the decimal 18 places. Previously when querying a delegator's `bondedAmount` the value returned was in WEI. It now gets returned in ETH (`81.551884481024178446` as opposed to `81551884481024178446`). This also makes the data easier to consume in client side application since it avoid the client having to convert the value from a BigInt to a float. This change was inspired by the new v2 uniswap subgraph.

- Changed all instances of `BigInt` in `BigDecimal` in the subgraph
- Created a helper function to convert values to decimal format
- Updated the explorer to account for the change in values)

I performed each staking action to ensure nothing broke and looked over all the affected pages and components.

_Note: this is a breaking change for users of the Livepeer subgraph and will require we reach out to developers we know are relying on it and make an announcement in the Discord #dev channel. I propose we keep this on our `livepeer-canary` subgraph for a 2-3 weeks to give developers a chance to update their applications._


kyriediculous:
I suppose this change (and the reason why the uniswap v2 subgraph also deos this) is suitable for time series data to avoid having to loop through an entire time series of i.e. ETH values and have to do `utils.fromWei()` for each of the values ?

adamsoffer:
> I suppose this change (and the reason why the uniswap v2 subgraph also deos this) is suitable for time series data to avoid having to loop through an entire time series of i.e. ETH values and have to do `utils.fromWei()` for each of the values ?

Yeah exactly...better performance when generating the charts client side since we don't have to convert every value in the time series. I forgot to mention that.

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -312,9 +310,7 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
                     {totalActiveStake === 0
                       ? 0
                       : (
-                          (parseFloat(
-                            Utils.fromWei(delegator.delegate.totalStake),
-                          ) /
+                          (parseFloat(delegator.delegate.totalStake) /
```
kyriediculous: Line 310-316 can be replaced using the unary plus operator as well. 

```suggestion
                          (+delegator.delegate.totalStake / totalActiveStake * 100).toPrecision(4)}
```

Javascript will not error on division by zero, instead the result will be NaN which might be even more accurate than returning 0.

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -43,12 +43,10 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
   }
 
   const pendingStake = parseFloat(Utils.fromWei(delegator.pendingStake))
-  const unbonded = delegator.unbonded
-    ? parseFloat(Utils.fromWei(delegator.unbonded))
-    : 0
-  const principal = parseFloat(Utils.fromWei(delegator.principal))
+  const unbonded = delegator.unbonded ? parseFloat(delegator.unbonded) : 0
```
kyriediculous: Can use unary plus here, this is true for as good as all cases where we parseInt/parseFloat now because we no longer have to call `utils.fromWei`/ `utils.toWei`

```javascript
+undefined = NaN 
+"" = 0
```

```suggestion
  const unbonded = +delegator.unbonded
```

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -43,12 +43,10 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
...
+  const principal = parseFloat(delegator.principal)
```
kyriediculous:
```suggestion
  const principal = +delegator.principal
```

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -43,12 +43,10 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
...
   const rewards = pendingStake + (unbonded ? unbonded : 0) - principal
-  const totalActiveStake = parseFloat(Utils.fromWei(protocol.totalActiveStake))
+  const totalActiveStake = parseFloat(protocol.totalActiveStake)
```
kyriediculous:
```suggestion
  const totalActiveStake = +protocol.totalActiveStake
```

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -324,7 +320,7 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
                 <span>
                   <span sx={{ fontFamily: 'monospace' }}>
                     {abbreviateNumber(
-                      parseFloat(Utils.fromWei(delegator.delegate.totalStake)),
+                      parseFloat(delegator.delegate.totalStake),
```
kyriediculous:
```suggestion
                      +delegator.delegate.totalStake,
```

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -344,9 +340,7 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
                     {(totalActiveStake === 0
                       ? 0
                       : ((totalActiveStake -
-                          parseFloat(
-                            Utils.fromWei(delegator.delegate.totalStake),
-                          ) -
+                          parseFloat(delegator.delegate.totalStake) -
```
kyriediculous: Seems like we re-use this value quite a bit in this code block.

What do you think of putting it on the heap after L49 and re-using it instead?

packages/explorer-2.0/components/StakingView/index.tsx
```diff
@@ -360,9 +354,7 @@ const Index = ({ delegator, transcoders, protocol, currentRound }) => {
                       totalActiveStake === 0
                         ? 0
                         : totalActiveStake -
-                            parseFloat(
-                              Utils.fromWei(delegator.delegate.totalStake),
-                            ) -
+                            parseFloat(delegator.delegate.totalStake) -
```
kyriediculous: See comment above about avoiding parsing this value multiple times

Should also be more memory efficient because I think what's happening now is 

1. Read delegator from heap
2. Read delegate from heap 
3. Read totalStake from heap 
4. Add totalStake to call stack
5. add parseFloat(totalStake) to callstack
6. Add result of (5) to callstack
...

compared to 

1. read totalStake from heap 
2. add totalStake to callstack
... 

In turn I believe the this would allow `delegator` to be GC'd earlier as well. 

Not sure on the JS compiler details here for nested objects but I think it's roughly accurate. 

Anyway don't mind my rambling here, just trying to practice thinking and being mindful about how code behaves on a low level a bit more.

packages/explorer-2.0/components/StakingWidget/Input.tsx
```diff
@@ -9,17 +9,16 @@ let roundsPerYear = hoursPerYear / averageHoursPerRound
 const Input = ({ transcoder, value = '', onChange, protocol, ...props }) => {
   const client = useApolloClient()
   const { width } = useWindowSize()
-  const totalSupply = Number(Utils.fromWei(protocol.totalTokenSupply))
-  const totalStaked = Number(Utils.fromWei(protocol.totalActiveStake))
+  const totalSupply = Number(protocol.totalTokenSupply)
+  const totalStaked = Number(protocol.totalActiveStake)
```
kyriediculous: Can use unary plus 

```suggestion
  const totalSupply = +protocol.totalTokenSupply
  const totalStaked =  +protocol.totalActiveStake
```

packages/explorer-2.0/components/StakingWidget/Input.tsx
```diff
@@ -114,5 +113,9 @@ function calculateAnnualROI({
         : inflation + inflationChange
   }
 
-  return roi ? roi : 0
+  return roi
+    ? Utils.fromWei(
+        Math.round(roi).toLocaleString('fullwide', { useGrouping: false }),
+      )
+    : 0
```
kyriediculous:
```suggestion
       return Utils.fromWei((Math.round(+roi) || 0).toString()),
```

The compiler is smart enough to avoid scientific notation when converting to string 

i.e. 

```javascript
1e5.toString() = "500000"
```

```javascript
const roi = NaN
(Math.round(+roi) || 0).toString()
// result: "0"
```

packages/explorer-2.0/components/StakingWidget/InputBox.tsx
```diff
@@ -17,12 +17,12 @@ const InputBox = ({
 
   let stake = '0'
   if (
-    Utils.fromWei(delegator?.bondedAmount ? delegator?.bondedAmount : '0') >
-    Utils.fromWei(delegator?.pendingStake ? delegator?.pendingStake : '0')
+    delegator?.bondedAmount
```
kyriediculous: I find `(delegator?.bondedAmount || '0') to be more legible , but I'll leave it up to you

packages/explorer-2.0/components/StakingWidget/InputBox.tsx
```diff
@@ -17,12 +17,12 @@ const InputBox = ({
...
+      ? delegator?.bondedAmount
+      : '0' >
```
kyriediculous: I find `(delegator?.bondedAmount || '0')` to be more legible , but I'll leave it up to you 

```suggestion
   (delegator?.bondedAmount || '0') >
```

packages/explorer-2.0/components/StakingWidget/InputBox.tsx
```diff
@@ -17,12 +17,12 @@ const InputBox = ({
...
+        Utils.fromWei(delegator?.pendingStake ? delegator?.pendingStake : '0')
```
kyriediculous:
```suggestion
        Utils.fromWei(delegator?.pendingStake || '0')
```

packages/explorer-2.0/components/StakingWidget/InputBox.tsx
```diff
@@ -17,12 +17,12 @@ const InputBox = ({
...
   ) {
-    stake = Utils.fromWei(
-      delegator?.bondedAmount ? delegator?.bondedAmount : '0',
-    )
+    stake = delegator?.bondedAmount ? delegator?.bondedAmount : '0'
```
kyriediculous: We are already evaluating this expression at L20 , we can re-use it

packages/explorer-2.0/components/Tokenholders/index.tsx
```diff
@@ -142,8 +142,7 @@ function renderSwitch(cell, protocol) {
     case 'Equity':
       return (
         <span sx={{ fontFamily: 'monospace' }}>{`${(
-          (parseFloat(Utils.fromWei(cell.value)) /
-            parseFloat(Utils.fromWei(protocol.totalActiveStake))) *
+          (parseFloat(cell.value) / parseFloat(protocol.totalActiveStake)) *
           100
```
kyriediculous:
```suggestion
         +cell?.value / +protocol?.totalActiveStake * 100
```

We can use unary plus and the order of operations doesn't matter here ( the multiplication is evaluated first regardless) so no need for extra parenthesis.

packages/explorer-2.0/components/VotingWidget/index.tsx
```diff
@@ -29,13 +29,13 @@ const Index = ({ data }) => {
   }, [copied])
 
   let noVoteStake = parseFloat(
-    Utils.fromWei(data.poll?.tally?.no ? data.poll?.tally?.no : '0'),
+    data.poll?.tally?.no ? data.poll?.tally?.no : '0',
```
kyriediculous:
```suggestion
    data.poll?.tally?.no || '0',
```

packages/explorer-2.0/components/VotingWidget/index.tsx
```diff
@@ -29,13 +29,13 @@ const Index = ({ data }) => {
...
   )
   let yesVoteStake = parseFloat(
-    Utils.fromWei(data.poll?.tally?.yes ? data.poll?.tally?.yes : '0'),
+    data.poll?.tally?.yes ? data.poll?.tally?.yes : '0',
```
kyriediculous:
```suggestion
    data.poll?.tally?.yes || '0',
```

although I believe at this point this will always be defined because we impose similar checks here https://github.com/livepeer/livepeerjs/blob/3d41e72236a7359df16a0f85508acf08f2429f56/packages/explorer-2.0/lib/createSchema.tsx#L145

packages/explorer-2.0/components/VotingWidget/index.tsx
```diff
@@ -29,13 +29,13 @@ const Index = ({ data }) => {
...
   )
   let totalVoteStake = noVoteStake + yesVoteStake
-  let totalNonVoteStake = parseFloat(Utils.fromWei(data.poll.totalNonVoteStake))
+  let totalNonVoteStake = parseFloat(data.poll.totalNonVoteStake)
```
kyriediculous: Can use unary plus, sorry if I'm overdoing it just discovered this feature. 

Also just a good idea to have a ubiquitous method of parsing in here I find. 

```suggestion
  let totalNonVoteStake = +data?.poll?.totalNonVoteStake
```

packages/explorer-2.0/components/StakingGuide/Step4.tsx
```diff
@@ -83,9 +83,7 @@ const Step4 = ({ goTo, nextStep }) => {
         ETH Balance:{' '}
         <span sx={{ fontWeight: 'bold' }}>
           {dataMyAccount?.account &&
-            parseFloat(Utils.fromWei(dataMyAccount.account.ethBalance)).toFixed(
-              2,
-            )}
+            parseFloat(dataMyAccount.account.ethBalance).toFixed(2)}
```
kyriediculous: What do you think of using `abbreviateNumber` here as well and getting rid of the `AND` statement since we can use null propagation operator ? 

```suggestion
abbreviateNumber(+dataMyAccount?.account?.ethBalance, 2)
```

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -203,17 +199,18 @@ const Index = async () => {
               isActive ? blockNumber : _poll.endBlock,
             )
             let noVoteStake = parseFloat(
-              Utils.fromWei(_poll?.tally?.no ? _poll?.tally?.no : '0'),
+              _poll?.tally?.no ? _poll?.tally?.no : '0',
```
kyriediculous:
```suggestion
              let noVoteStake = +_poll?.tally?.no || 0
```

packages/explorer-2.0/lib/createSchema.tsx
```diff
@@ -203,17 +199,18 @@ const Index = async () => {
...
             )
             let yesVoteStake = parseFloat(
-              Utils.fromWei(_poll?.tally?.yes ? _poll?.tally?.yes : '0'),
+              _poll?.tally?.yes ? _poll?.tally?.yes : '0',
```
kyriediculous:
```suggestion
              let yesVoteStake = +_poll?.tally?.yes || 0
```

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -45,7 +45,7 @@ export const getDelegatorStatus = (
   if (
     !delegator ||
     !delegator.bondedAmount ||
-    parseFloat(Utils.fromWei(delegator.bondedAmount)) === 0
+    parseFloat(delegator.bondedAmount) === 0
```
kyriediculous: This whole expression (L46-L48) can be shortened I think

```suggestion
    !+delegator?.bondedAmount
```

```javascript
let a = {x:0}

!+a?.x
// true
!+a?.y
// true
```

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -393,9 +393,7 @@ export const simulateNewActiveSetOrder = ({
   }
 
   if (action === 'stake') {
-    transcoders[index].totalStake = Utils.toBN(transcoders[index].totalStake)
-      .add(Utils.toBN(amount))
-      .toString()
+    transcoders[index].totalStake = +transcoders[index].totalStake + +amount
```
kyriediculous, line 395:
```suggestion
    transcoders[index].totalStake += +amount
```

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -406,22 +404,15 @@ export const simulateNewActiveSetOrder = ({
       const oldDelegateIndex = transcoders.findIndex(
         (t) => t.id.toLowerCase() === oldDelegate.toLowerCase(),
       )
-      transcoders[oldDelegateIndex].totalStake = Utils.toBN(
-        transcoders[oldDelegateIndex].totalStake,
-      )
-        .sub(Utils.toBN(amount))
-        .toString()
+      transcoders[oldDelegateIndex].totalStake =
+        +transcoders[oldDelegateIndex].totalStake - +amount
```
kyriediculous, line 407:
```suggestion
      transcoders[oldDelegateIndex].totalStake -= +amount
```

packages/explorer-2.0/lib/utils.tsx
```diff
@@ -406,22 +404,15 @@ export const simulateNewActiveSetOrder = ({
...
     }
   } else {
-    transcoders[index].totalStake = Utils.toBN(transcoders[index].totalStake)
-      .sub(Utils.toBN(amount))
-      .toString()
+    transcoders[index].totalStake = +transcoders[index].totalStake - +amount
```
kyriediculous, line 410:
```suggestion
    transcoders[index].totalStake -= +amount
```

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -138,11 +138,9 @@ const Poll = () => {
     )
   }
   console.log(pollData)
-  let noVoteStake = parseFloat(
-    Utils.fromWei(pollData?.tally?.no ? pollData?.tally?.no : '0'),
-  )
+  let noVoteStake = parseFloat(pollData?.tally?.no ? pollData?.tally?.no : '0')
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]:
```suggestion
  let noVoteStake = +pollData?.tally?.no || 0
```

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -138,11 +138,9 @@ const Poll = () => {
...
   let yesVoteStake = parseFloat(
-    Utils.fromWei(pollData?.tally?.yes ? pollData?.tally?.yes : '0'),
+    pollData?.tally?.yes ? pollData?.tally?.yes : '0',
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]: let yesVoteStake = +pollData?.tally?.yes || 0

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -440,14 +438,10 @@ const Poll = () => {
...
 async function transformData({ poll }) {
-  let noVoteStake = parseFloat(
-    Utils.fromWei(poll?.tally?.no ? poll?.tally?.no : '0'),
-  )
-  let yesVoteStake = parseFloat(
-    Utils.fromWei(poll?.tally?.yes ? poll?.tally?.yes : '0'),
-  )
-  let totalVoteStake = parseFloat(Utils.fromWei(poll.totalVoteStake))
-  let totalNonVoteStake = parseFloat(Utils.fromWei(poll.totalNonVoteStake))
+  let noVoteStake = parseFloat(poll?.tally?.no ? poll?.tally?.no : '0')
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]:
```suggestion
  let noVoteStake = +poll?.tally?.no || 0
```

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -440,14 +438,10 @@ const Poll = () => {
...
+  let yesVoteStake = parseFloat(poll?.tally?.yes ? poll?.tally?.yes : '0')
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]:
```suggestion
  let yesVoteStake = +poll?.tally?.yes || 0
```

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -440,14 +438,10 @@ const Poll = () => {
...
+  let totalVoteStake = parseFloat(poll.totalVoteStake)
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]:
```suggestion
  let totalVoteStake = +poll?.totalVoteStake
```

packages/explorer-2.0/pages/voting/[poll].tsx
```diff
@@ -440,14 +438,10 @@ const Poll = () => {
...
+  let totalNonVoteStake = parseFloat(poll.totalNonVoteStake)
```
Comment (kyriediculous) [inline packages/explorer-2.0/pages/voting/[poll].tsx]:
```suggestion
  let totalNonVoteStake = +poll?.totalNonVoteStake
```

Instead of using the null propagation operator on each of the values here we could also simply add a check at the beginning of the function; this will save us some operations. 

```
if (!poll) return;
```

packages/subgraph/src/mappings/bondingManager.ts
```diff
@@ -215,15 +216,15 @@ export function bond(call: BondCall): void {
         Address.fromString(oldTranscoder.id),
       )
 
-      oldTranscoder.totalStake = delegateData.value3
-      oldDelegate.delegatedAmount = delegateData.value3
+      oldTranscoder.totalStake = convertToDecimal(delegateData.value3)
```
kyriediculous, line 219: Is it possible to used named accessors here instead of things like `value3` or is that not allowed ?

kyriediculous (changes requested):
Overall looks good ! 

Just a bunch of nitpicky style / readability comments

packages/subgraph/src/mappings/bondingManager.ts
adamsoffer: not allowed/available unfortunately :/

PR #855, opened 2020-11-30
add daily fee volume to subgraph

adamsoffer:
This PR introduces a new entity to the Livepeer subgraph for indexing daily volume in ETH and USD.

- Indexes eth price based on the uniswap v2 eth/dai pair
- Introduces a new entity type to the schema called `DayData` for indexing daily volume in ETH and USD.
- Renames field `totalGeneratedFees` to `totalVolumeETH`

I deployed the subgraph to my own test subgraph `https://thegraph.com/explorer/subgraph/adamsoffer/livepeer-canary` and ensured the values returned are accurate.


packages/subgraph/schema.graphql
```diff
@@ -220,6 +227,14 @@ type Vote @entity {
   registeredTranscoder: Boolean
 }
 
+# Data accumulated and condensed into day stats for protocol fee volume
+type DayData @entity {
```
kyriediculous, line 231: What do you think of doing this on a per round basis, as rounds provide a timing framework that's engrained within the protocol or is this what the `Protocol` entity already does ? 

Is the idea to have a time series available for each day rather than each round, so more product related reasoning? 

```
type RoundData @entity {
id: ID! # round number
volumeETH: BigDecimal!
volumUSD: BigDecimal!
}

```

adamsoffer: I actually added it per round as well :) https://github.com/livepeer/livepeerjs/blob/as/subgraph-daily-data/packages/subgraph/src/mappings/ticketBroker.ts#L92

But, yeah, the reason for adding it per day was a product decision. Most users are used to seeing time series charts in days/weeks/months. Also, displaying fees over days will probably give users more insight. For example, are there certain days of the week that see more fee volume than others? I'm thinking we could give users the option to toggle between either days or rounds though.

kyriediculous: Oh that's great, I like the product oriented thinking there actually and having access to both is a win-win either way! 👍

PR #869, opened 2020-12-16
add overview page

### LINKED ISSUE #862: explorer: Protocol overview page
Contains network participation rate chart and fee volume chart.

adamsoffer:
This PR introduces an overview page featuring network stats (fee volume and participation over time) and an overview of the top orchestrators.

TODOs:

- [x] Add and default to weekly view for fee volume chart
- [x] Rename "Volume (24hr)" to "Fee Volume (7d)"
- [x] Add percentage change to charts
- [x] Improve overview page speed performance using next.js SSG (static site generation)
- [x] Fix past trailing 7 week fee volume total

- Introduces a historical fee volume chart
- Introduces a historical participation rate chart
- Introduces a dedicated orchestrators page featuring 30 orchestrators per page
- Refactors pages to use SSG for improved performance

Close #862 

**Screenshots**
overview


yondonfu:
Preview URL is looking good!

> Add and default to weekly view for fee volume chart

I do suspect that the weekly view will be a better default for the fee volume chart especially since a longer time frame will result in a smoother data distribution that would otherwise look really spiky, as is the case with the current daily view in the preview URL, given that there may be 0 winning tickets redeemed on some days.

I also wonder if the daily view should be included for the fee volume chart at all right now since days with 0 winning tickets are fairly common right now. As a result, on those days, the daily view would just show a big 0 for the fees earned on the current day which can be misleading for the user. This will become less of a problem as the overall fees earned increase such that the number of days with 0 winning tickets will be minimized so it makes more sense to exclude a daily view until then.

> Rename "Fee volume" to "Fees Earned"

👍 

> Add percentage change to charts

Would this reflect the +/- percentage change since the last time period (i.e. last week) similar to the +/- percentage shown in the "Liquidity" and "Volume" charts at uniswap.info? If so, 👍 .

adamsoffer:
> I also wonder if the daily view should be included for the fee volume chart at all right now since days with 0 winning tickets are fairly common right now. 

Yeah Doug brought this up as well. I think it makes sense to remove the daily view altogether until fee volume increases. 

> Would this reflect the +/- percentage change since the last time period (i.e. last week) similar to the +/- percentage shown in the "Liquidity" and "Volume" charts at uniswap.info? 

Yeah exactly 👍

yondonfu:
Side note: I was going to ask about when the fee volume is converted to USD, but I noticed that as of #855 the subgraph is using the Uniswap ETH/DAI pair to convert the daily fee volume to USD on the day that the fee volume is recorded. Nice!

yondonfu:
A few minor comments:

Screen Shot 2020-12-21 at 12 34 48 PM

- The 0 in the last decimal place for $1000.00 on the y-axis of the fee volume chart is cut off
- Should we be consistent with either including 0s in the decimal places of all y-axis values of the fee volume chart? Right now all of the values include trailing 0s except for $1400 and $1200. Maybe trailing 0s can just be omitted?

adamsoffer:
> Should we be consistent with either including 0s in the decimal places of all y-axis values of the fee volume chart? 

@yondonfu ah yeah I'll remove trailing zeros there 👍

PR #884, opened 2021-01-03
Make subgraph improvements

adamsoffer:
This PR includes a bunch of subgraph improvements. Details below.

- Refactors transaction type for better visibility into transaction history and events. Now you can view multiple events associated with a single transaction entity
- Denotes event types by suffixing `Event` to all types that implement the `Event` interface.
- Adds missing descriptive comments to graphql types for autogenerated docs
- Refactors the way entities get created by abstracting into helper methods (inspired by the Graph's own subgraph that powers their network page)
- Makes schema.yaml entity fields required
- Renames ABI files to correspond with version names used by [deprecated contracts](https://github.com/livepeer/wiki/blob/master/Deployed-Contract-Addresses.md#mainnet-deprecated) (ie BondingManagerV1).
- Consolidates BondingManger and RoundsManager mappings into a single file and moved deprecated mappings into dedicated files
- Adds face value in USD to winning ticket events
- Renames `DayData` entity to `Day`
- Adds `endBlock` to `Round` entity
- Adds eth price from uniswap V1 so now we can track missing volume from January to May.
- Refactors the explorer history view to account for changes to `Transaction` entity 

Deployed updates to my personal [test subgraph ](https://thegraph.com/explorer/subgraph/adamsoffer/livepeer)

**Additional Context**
The time to sync now on the Graph's hosted service is now a reasonable 4.5 hours.


packages/explorer-2.0/components/HistoryView/index.tsx
```diff
@@ -56,6 +54,11 @@ const Index = () => {
     return <div sx={{ pt: 5 }}>No history</div>
   }
 
+  const events = data.transactions.reduce(
+    (res, { events }) => res.concat(...events),
```
kyriediculous: Small nit: `Array.prototype.concat` returns a new array evertime. This means for every iteration of `reduce` a new array will be allocated. Could we use `Array.prototype.push` instead ? 

Also if sticking with `concat` I believe we don't need to deconstruct the array.

kyriediculous (commented):
LGTM overall 👍

kyriediculous (approved):

PR #888, opened 2021-01-06
add transcoder day data

adamsoffer:
This PR adds a `TranscoderDay` entity for tracking daily fees earned by transcoders. This can be used for estimating minutes of video transcoded per orchestrator in combination with off chain pricing data.

### DIFF
--- a/packages/subgraph/utils/helpers.ts
+++ b/packages/subgraph/utils/helpers.ts
@@ -3,10 +3,10 @@ import { integer } from '@protofire/subgraph-toolkit'
 import {
   Day,
   Delegator,
-  Pool,
   Protocol,
   Round,
   Transcoder,
+  TranscoderDay,
 } from '../src/types/schema'
 
 let x = BigInt.fromI32(2)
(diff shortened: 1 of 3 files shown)


packages/subgraph/utils/helpers.ts
```diff
@@ -212,6 +212,28 @@ export function createOrLoadDay(timestamp: i32): Day {
   return day as Day
 }
 
+export function createOrLoadTranscoderDay(
+  timestamp: i32,
+  transcoderAddress: string,
+): TranscoderDay {
+  let dayID = timestamp / 86400
+  let dayStartTimestamp = dayID * 86400
+  let transcoderDayID = transcoderAddress
```
kyriediculous, line 221: This doesn't seem to be in line with the documentation above 

```
  "Timestamp rounded to current day by dividing by 86400"
  id: ID!
  ```

in the mapping the id is defined as the timestamp rounded to the current day, but in this code snippet the id is constructed as a combined value. I assume the combined value is correct though, so let's adjust the docs/comment

kyriediculous (changes requested):
LGTM just a small comment

packages/subgraph/utils/helpers.ts
adamsoffer: yeah good spot! Just adjusted the docs/comment.

kyriediculous (approved):
LGTM 👍

PR #901, opened 2021-01-14
add back subgraph tests

adamsoffer:
This PR adds back the subgraph tests.

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -17,7 +17,10 @@
     "prepare:mainnet": "NETWORK_NAME=mainnet node ./templatify.js",
     "prepare:rinkeby": "NETWORK_NAME=rinkeby node ./templatify.js",
     "prepare:development": "NETWORK_NAME=development node ./templatify.js",
-    "docker:build": "docker build --cache-from livepeer/subgraph:latest -t livepeer/subgraph:latest ."
+    "docker:build": "docker build --cache-from livepeer/subgraph:latest -t livepeer/subgraph:latest .",
+    "test:docker": "DOCKER=true npx truffle test --network=docker",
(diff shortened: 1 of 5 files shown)


adamsoffer:
@yondonfu the subgraph tests are passing again :)

adamsoffer:
Actually doesn't appear they ran...

adamsoffer:
Okay it's running again albeit most tests are failing. Guessing we have to update the docker image again with the latest contract fix.

yondonfu:
@adamsoffer I updated the `livepeer/ganache-with-livepeer-protocol:streamflow` image to include the latest contract fix.

I also pushed a few updates. The commit history describes the changes, but a few things worth noting:

- I had to add 3cf4aa3 because the graph-node logs complained about storing the block hash as a UTF-8 string instead of a hex string.
- I had to add 62ff6a2 because the `lastRoundLengthUpdateRound` in the RoundsManager is not necessarily the current round when the `roundLength` parameter is updated. I added a contract call to fetch the latest `lastRoundLengthUpdateRound` instead.
- Disabling the earnings claimed poll tally handler exposed a bug in the bond poll tally handler which is fixed in 031e9ca
- 4649c01 updates the earnings claimed handler to set the correct end round (it was previously using the start round for the end round)

adamsoffer:
Nice! Fixes look great. Glad we're able to [return early inside the `earningsClaimed` handler](https://github.com/livepeer/livepeerjs/pull/901/files#diff-5a2c8a7ae0a6b9d85279bb5c122e162d178119b1a604cfaa6bf357b7f602e4b4R203) now - will speed up time to sync.

adamsoffer:
@yondonfu ready to merge?

yondonfu:
> @yondonfu ready to merge?

Yep!

PR #909, opened 2021-01-20
add usage chart

adamsoffer:
TODOs:
- [x] mobile 
- [x] fix chart slider "flickering" issue on page load
- [x] add banner that links to what's new section


packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -203,9 +240,63 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+      toTime: twoWeeksAgo.getTime(),
+    });
+
+    // merge in Livepeer.com usage data
+    dayData = dayData.map((item) => {
+      const found = livepeerComDayData.find(
+        (element) => item.date == element.date
+      );
+      // combine Livepeer.com minutes with minutes calculated via fee volume
+      const minutes =
```
yondonfu: Looks like averagePrice is 6000 wei converted to USD based on a specific fixed USD/ETH exchange rate.

Since the volumeUSD per day is calculated based on the DAI/ETH exchange rate on the day that the fees were generated, I think it could make sense to also determine the USD average price based on the DAI/ETH exchange rate on the day that the fees were generated. I think we should be able to derive the DAI/ETH exchange rate based on the volumeUSD and volumeETH fields included in the Day entity.

For example:

```javascript
// 6000 wei in ETH terms
const averagePrice = 0.000000000000006
...

dayData = dayData.map((item) => {
    ...

    const ethDaiRate = +item.volumeETH / +item.volumeUSD
    const usdAveragePrice = averagePrice * ethDaiRate
    const minutes =
        (found?.sourceSegmentsDuration ?? 0) / 60 +
        +item.volumeUSD / usdAveragePrice / pixelsPerMinute
    ...
})
```

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -147,6 +150,9 @@ export async function getChartData(_obj, _args, _ctx, _info) {
     participationRate: 0,
   };
 
+  const averagePrice = 0.0000000000094;
+  const pixelsPerMinute = 2995488000;
```
yondonfu, line 160: Might be helpful to add a comment here for future reference indicating how this value is calculated i.e. the # of pixels in a minute of 240p30fps, 360p30fps, 480p30fps, 720p30fps transcoded renditions.

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -203,9 +240,63 @@ export async function getChartData(_obj, _args, _ctx, _info) {
     let dayDataResult = await getDayData();
     dayData = dayDataResult.data.days;
 
+    let oneWeekAgo = new Date();
+    oneWeekAgo.setDate(oneWeekAgo.getDate() - 7);
+    oneWeekAgo.setUTCHours(0, 0, 0, 0);
+
+    let twoWeeksAgo = new Date();
+    twoWeeksAgo.setDate(twoWeeksAgo.getDate() - 7);
```
yondonfu: Looks like we're setting twoWeeksAgo to the same value as oneWeekAgo here?

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -203,9 +240,63 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+      // combine Livepeer.com minutes with minutes calculated via fee volume
+      const minutes =
+        (found?.sourceSegmentsDuration ?? 0) / 60 +
+        +item.volumeUSD / averagePrice / pixelsPerMinute;
+      return { ...item, ...found, minutes };
+    });
+
+    // get total Livepeer.com aggregate usage
+    let totalLivepeerComUsage = livepeerComDayData.reduce((x, y) => {
+      return x + y.sourceSegmentsDuration / 60;
```
yondonfu, line 262: I believe sourceSegmentsDuration will track source streams ingested by the .com API while transcodedSegmentsDuration will track transcoded streams for the .com API. Since the network's main responsibility is transcoding right now as opposed to ingest, I think using transcodedSegmentsDuration instead of sourceSegmentsDuration would make sense here.

yondonfu (changes requested):

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: Oops yeah that's supposed to be `twoWeeksAgo.getDate() - 14`

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: Took a stab at this. Let me know if it looks right to you.

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: Done.

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: Fixed.

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: Done.

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -203,9 +215,56 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+      toTime: utcTwoWeeksBack * 1000, // Livepeer.com api uses milliseconds
+    });
+
+    // merge in Livepeer.com usage data
+    dayData = dayData.map((item) => {
+      let found = livepeerComDayData.find(
+        (element) => item.date == element.date
+      );
+
+      let ethDaiRate = +item.volumeETH / +item.volumeUSD;
```
yondonfu: Could getTotalFeeDerivedMinutes be used here?

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -216,57 +275,93 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
-          currentWeek = week;
-          startIndexWeekly++;
-        }
-        weeklyData[startIndexWeekly] = weeklyData[startIndexWeekly] || {};
-        weeklyData[startIndexWeekly].date = weeklySizedChunk.date;
-        weeklyData[startIndexWeekly].weeklyVolumeUSD =
-          (weeklyData[startIndexWeekly].weeklyVolumeUSD ?? 0) +
-          +weeklySizedChunk.volumeUSD;
-      }
+    let [oneWeekVolumeETH, _weeklyVolumeChangeETH] = get2DayPercentChange(
```
yondonfu: Is _weeklyVolumeChangeETH used anywhere?

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -216,57 +275,93 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+      oneDayData.totalVolumeUSD ? oneDayData.totalVolumeUSD : 0,
+      twoDayData.totalVolumeUSD ? twoDayData.totalVolumeUSD : 0
+    );
 
-      // format the total participation change
-      const participationRateChange = getPercentChange(
-        data.participationRate,
-        oneDayData.participationRate
-      );
+    let [oneWeekVolumeUSD, weeklyVolumeChangeUSD] = get2DayPercentChange(
```
yondonfu: Just noting for clarification - get2DayPercentChange() does not necessarily need to accept param values that represent data between 24 hours ago and 48 hours ago right? Assuming so based on glancing at the implementation.

adamsoffer: yeah that's correct. I should rename that function...

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: yeah just realized that. Done. https://github.com/livepeer/livepeerjs/pull/909/commits/1bc07f9ca0da3574480d620ebda0f8d6d5de7cc7

packages/explorer-2.0/apollo/resolvers/Query.tsx
adamsoffer: It is not actually. Just removed 👍

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -203,9 +215,58 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+
+      // combine Livepeer.com minutes with minutes calculated via fee volume
+      let minutes =
+        (found?.transcodedSegmentsDuration ?? 0) / 60 + feeDerivedMinutes;
+      return { ...item, ...found, minutes };
+    });
+
+    // get total Livepeer.com aggregate usage
+    let totalLivepeerComUsage = livepeerComDayData.reduce((x, y) => {
+      return x + y.transcodedSegmentsDuration / 60;
```
yondonfu: Okay after thinking about this some more I now think my original suggestion to use transcodedSegmentsDuration might have been misplaced. transcodedSegmentsDuration will include the duration of each of the transcoded renditions for a source video segment. So, given let's say 4 transcoded renditions, the transcoded segments duration tracked for each source video segment will be the source duration multiplied by 4. I don't think it makes sense to track usage in this manner right now because:

- feeDerivedMinutes represents the # of minutes of source video given the # of pixels in an assumed set of transcoded renditions. So, if we use transcodedSegmentsDuration for the .com API usage, there would be a mismatch between what the .com API usage represents and what feeDerivedMinutes represents
- Focusing the definition of usage on the duration of source video that is transcoded as opposed to the combined duration of transcoded renditions feels simpler right now and easier to understand (i.e. one just needs to consider the amount of video that is being sent into the network)

adamsoffer: This makes sense to me. This would also be consistent with how we're displaying usage back to Livepeer.com customers (using sourceSegmentsDuration).

yondonfu (approved):
LGTM after completing the banner text TODO

PR #931, opened 2021-02-20
subgraph: fix active field for Transcoder entity

### LINKED ISSUE #929: subgraph: active field for Transcoder entity should not be set in TranscoderActivated/Deactivated event handlers
**Describe the bug (required)** A clear and concise description of what the bug
is.

At the moment, for the Transcoder entity, the active field field is set to true when the [TranscoderActivated](https://github.com/livepeer/livepeerjs/blob/bbe439701dea7b798ab08d41ce4b5f739d2b7653/packages/subgraph/src/mappings/bondingManager.ts#L526) event is emitted and to false when the [TranscoderDeactivated](https://github.com/livepeer/livepeerjs/blob/bbe439701dea7b798ab08d41ce4b5f739d2b7653/packages/subgraph/src/mappings/bondingManager.ts#L558) event is emitted. However, a transcoder only becomes active in the round *after* the TranscoderActivated event is emitted and inactive in the round *after* the TranscoderDeactivated event is emitted (this is a bit confusing...I know :/). As a result, if someone sends a GraphQL query with `where: { active: true }` expecting to get only active transcoders, they might receive transcoders that are actually still inactive (i.e. pending activation).

I think this could be resolved by setting the active field in the [NewRound](https://github.com/livepeer/livepeerjs/blob/bbe439701dea7b798ab08d41ce4b5f739d2b7653/packages/subgraph/src/mappings/roundsManager.ts#L54) handler. As the handler loops through the on-chain transcoder pool, it can check if the transcoder's activation round <= currentRound < transcoder's deactivation round and if so set active = true, else set active = false.

adamsoffer:
This PR fixes the active field for the Transcoder entity.

Deployed the fix to the [livepeer-canary](https://thegraph.com/explorer/subgraph/livepeer/livepeer-canary) subgraph.

Fixes #929

### DIFF
--- a/packages/subgraph/src/mappings/roundsManager.ts
+++ b/packages/subgraph/src/mappings/roundsManager.ts
@@ -66,9 +66,19 @@ export function newRound(event: NewRound): void {
     pool.totalStake = transcoder.totalStake;
     pool.rewardCut = transcoder.rewardCut as BigInt;
     pool.feeShare = transcoder.feeShare as BigInt;
-
-    // Apply store updates
     pool.save();
+
+    // Update transcoder active state
+    if (
+      event.params.round.ge(transcoder.activationRound) &&
+      event.params.round.lt(transcoder.deactivationRound)
+    ) {
+      transcoder.active = true;
+    } else {
+      transcoder.active = false;
+    }
+
+    transcoder.save();
 
     currentTranscoder = bondingManager.getNextTranscoderInPool(
       currentTranscoder
--- a/packages/subgraph/src/mappings/roundsManager.ts
+++ b/packages/subgraph/src/mappings/roundsManager.ts
@@ -49,6 +49,29 @@ export function newRound(event: NewRound): void {
 
   let poolId: string;
   let pool: Pool;
+  let protocol = Protocol.load("0");
+
+  // Activate all transcoders pending activation
+  let pendingActivation = protocol.pendingActivation;
+  if (pendingActivation.length) {
+    for (let index = 0; index < pendingActivation.length; index++) {
+      let t = Transcoder.load(pendingActivation[index]);
+      t.active = true;
+      t.save();
+    }
+    protocol.pendingActivation = [];
+  }
+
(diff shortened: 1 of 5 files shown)


packages/subgraph/src/mappings/roundsManager.ts
```diff
@@ -66,10 +66,20 @@ export function newRound(event: NewRound): void {
     pool.totalStake = transcoder.totalStake;
     pool.rewardCut = transcoder.rewardCut as BigInt;
     pool.feeShare = transcoder.feeShare as BigInt;
-
-    // Apply store updates
     pool.save();
 
+    // Update transcoder active state
+    if (
+      event.params.round.ge(transcoder.activationRound) &&
```
yondonfu: `ge` -> `gte`

yondonfu (changes requested):
Realized that looping through the on-chain pool in the newRound handler won't quite cover all cases because if a transcoder is pending deactivation in round N, it will be inactive in round N + 1 and will not be in the on-chain pool. So, when the newRound handler fires in round N + 1, we won't be able to set the active field for that transcoder to false when we loop through the on-chain pool. We need a way to set the active field for that transcoder to false when round N + 1 begins. 

Maybe one of the following?

- Iterate through all registered transcoders in the newRound handler
- Maintain a global pending activation map and pending deactivation map. Add to former for TranscoderActivated events and add to latter for TranscoderDeactivated events. In the newRound handler, for each pending activation entry set active = true for the transcoder and for each pending deactivation entry set active = false for the transcoder. Then, clear both maps.

yondonfu (approved):
LGTM!

adamsoffer:
@yondonfu Just remembered I'll need to update the subgraph API reference in the docs with these new fields. I'll create a new issues inside livepeerjs for that.

yondonfu:
> @yondonfu Just remembered I'll need to update the subgraph API reference in the docs with these new fields. I'll create a new issues inside livepeerjs for that.

Good call!

PR #951, opened 2021-03-22
Support price updates when calculating usage

### LINKED ISSUE #950: Support price updates when calculating usage

adamsoffer:
This PR adds support for Livepeer.com price updates for more precise usage data displayed in the chart.

Closes #950

### DIFF
--- a/packages/explorer-2.0/apollo/resolvers/Query.tsx
+++ b/packages/explorer-2.0/apollo/resolvers/Query.tsx
@@ -246,7 +246,7 @@ export async function getChartData(_obj, _args, _ctx, _info) {
       // if Livepeer.com's broadcaster changed max price, use updated price
       if (
         pricePerPixelIndex &&
-        item.date <= pricePerPixel[pricePerPixelIndex].startDate
+        item.date < pricePerPixel[pricePerPixelIndex].startDate
       ) {
         pricePerPixelIndex--;
       }
--- a/packages/explorer-2.0/apollo/resolvers/Query.tsx
+++ b/packages/explorer-2.0/apollo/resolvers/Query.tsx
@@ -235,7 +235,7 @@ export async function getChartData(_obj, _args, _ctx, _info) {
     let totalFeeDerivedMinutes = 0;
     let totalFeeDerivedMinutesOneWeekAgo = 0;
     let totalFeeDerivedMinutesTwoWeeksAgo = 0;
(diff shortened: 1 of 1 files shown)


adamsoffer:
@yondonfu I updated the price to 3000 wei and the start date to March 23rd. Let me if I should change those to anything else.

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -225,15 +235,21 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
+    let pricePerPixelIndex = 0;
 
     // merge in Livepeer.com usage data
     dayData = dayData.map((item) => {
       const found = livepeerComDayData.find(
         (element) => item.date === element.date
       );
 
+      // if Livepeer.com's broadcaster changed max price, use updated price
+      if (item.date >= pricePerPixel[pricePerPixelIndex].endDate) {
```
yondonfu: Just to double check - `dayData` should already be sorted right? So, once this conditional evaluates to true once all following dates will be after `item.date`?

adamsoffer: Ah good catch! It's sorted, but in descending order. I'll push up a fix.

adamsoffer: I fixed the conditional to account for the fact days are in desc order.

packages/explorer-2.0/apollo/resolvers/Query.tsx
```diff
@@ -225,15 +235,24 @@ export async function getChartData(_obj, _args, _ctx, _info) {
...
     // merge in Livepeer.com usage data
     dayData = dayData.map((item) => {
       const found = livepeerComDayData.find(
         (element) => item.date === element.date
       );
 
+      // if Livepeer.com's broadcaster changed max price, use updated price
+      if (
+        pricePerPixelIndex &&
+        item.date <= pricePerPixel[pricePerPixelIndex].startDate
```
yondonfu: Minor: Should <= be < ? Otherwise, if the date is exactly 1616457600 then we would decrement the counter and use the previous price despite date of the current item being within the date range for the current price.

yondonfu (approved):
Left one minor comment, but feel free to address and merge without another review.

Issue #114, opened 2018-06-18
Removing Transfer Allowance Pop Up


RaffiSapire:
**Describe the bug (required)**
I personally felt very irritated that I had to pay to change my allowance. Requiring people to change their transfer allowance takes both time and gas, and therefore causes people people drop off and have a negative experience with Livepeer. Please can you lay out the technical challenge this solves and can we brainstorm how to solve without using the token holder? Here's feedback from a user interview: 

"Having the initial transfer allowance at 0 means I have to spend gas to update it straight away... why not start at a moderate amount initially so people don't have to jump through so many hoops to get started?"

**Expected behavior (required)**
When I click bond, I just want to bond. I don't want to have to adjust my limits and pay gas and then pay gas again to bond.

jozanza:
Yeah it’s an ERC20 thing unfortunately. You need to set one, and there’s no way around that :/

jozanza:
I'm going to close this issue, but it's definitely something to think about when user testing. There will absolutely be ways to continue to improve the first-time bonding experience. Unfortunately, transfer approval is a requirement at this point in time.

PR #959, opened 2021-03-31
Remove earnings claimed event from history view

adamsoffer:
This PR removes the claimed earnings event from the history view. Since users no longer have to explicitly/manually claim earnings from the explorer this event is no longer relevant to a user of the explorer.

Closes [#114](https://github.com/livepeer/internal-project-tracking/issues/114)


yondonfu (approved):
LGTM

Could be a good idea to add a note in What's New noting why this information was removed in case some users are confused why that information no longer shows up in the history view.

adamsoffer:
> LGTM
> 
> Could be a good idea to add a note in What's New noting why this information was removed in case some users are confused why that information no longer shows up in the history view.

Good call. I'll create a release note for the what's new section.

PR #978, opened 2021-04-22
add approval step to staking widget

adamsoffer:
This PR adds some UX improvements to the staking widget. Namely:
- It adds the approval step to the staking widget itself as opposed to a banner 
- It adds an explicit call to action to "Enter an Amount" if no amount has been inputted 

Closes #978


yondonfu:
Looks great! Just one question:

I entered a non-zero amount in the staking widget and I noticed that the buttons for the 2-step flow were labeled with "Approve" and "Move Stake":

Screen Shot 2021-04-26 at 9 58 14 PM

In this case, my account has never staked before so it feels a bit weird to me for the 2nd button to say "Move Stake" since the account doesn't have any stake to begin with. Should the 2nd button just say "Stake" instead?

The one scenario where I could see the 2nd button saying "Move Stake" is if the account is already staked, needs to approve an additional amount, and the orchestrator in the staking widget is different than the account's current orchestrator. That being said, I believe the explorer currently uses infinite approvals and in that case I don't think any account would end up in this scenario (unless it approved and staked during a time before the explorer supported infinite approvals).

adamsoffer:
> In this case, my account has never staked before so it feels a bit weird to me for the 2nd button to say "Move Stake" since the account doesn't have any stake to begin with. Should the 2nd button just say "Stake" instead?

Ah nice catch. Yes this is a mistake. It should only say "Move Stake" if you're already staked and staking to a different orch. Will fix. Thanks!

adamsoffer:
@yondonfu fixed 👍

yondonfu (approved):
LGTM

PR #1026, opened 2021-11-01
change default ordering to total fees earned

adamsoffer:
As proposed in [this forum post](https://forum.livepeer.org/t/livepeer-explorer-rewarding-top-performers-through-ui-changes/1479), this PR changes the default ordering of the orchestrator leaderboard from total stake to total fees earned.

**Screenshots (optional):**
image

### DIFF
--- a/packages/explorer-2.0/components/AccountCell/index.tsx
+++ b/packages/explorer-2.0/components/AccountCell/index.tsx
@@ -47,8 +47,8 @@ const Index = ({ threeBoxSpace, active, address }) => {
           <QRCode
             style={{
               borderRadius: 1000,
-              width: 30,
-              height: 30,
+              width: 24,
+              height: 24,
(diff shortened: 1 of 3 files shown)


papabear99:
With this change the "staking" tab now lists orchestrators by fees earned so I propose either adding a "fees earned" tab to allow sorting by fees earned, stake or performance, or if the intent is to not have the ability to list orchestrators by stake then change the "staking" tab to "fees earned" to reflect what the tab is actually showing.

Franck-UltimaRatio:
Just an idea,
It would be interesting to add a tab indicating the date when the node started to work (like on https://top-performers.vercel.app/)
To know the efficiency of a node, it is necessary to make a link between earns / Stake / start date.
That could also help future delegators to know the reliability over time of a node.

PR #1034, opened 2021-11-19
subgraph: fix incorrect transcoder active states

### LINKED ISSUE #1027: subgraph: TranscoderDeactivated followed by TranscoderActivated in same round results in inactive transcoder in next round
The newRound handler [processes pending activations from the previous round and then pending deactivations from the previous round](https://github.com/livepeer/livepeerjs/blob/a16770afa8a6f6cc461af14b48a2c64438ad250c/packages/subgraph/src/mappings/roundsManager.ts#L57). If a TranscoderDeactivated event is emitted in the previous round and then a TranscoderActivated event is emitted in the previous round for the same transcoder address then the pending activation will be processed before the pending deactivation resulting in the `active` field of the transcoder being set to false. In this case, the `active` field of the transcoder should be set to true, because the TranscoderDeactivated event was emitted *before* the TranscoderActivated even in the previous round.

For an example of this bug in the wild, run the following query at https://thegraph.com/hosted-service/subgraph/livepeer/livepeer

```graphql
{
  transcoders(where:{ id:"0xda43d85b8d419a9c51bbf0089c9bd5169c23f2f9" }) {
    active
    activationRound
    deactivationRound
  }
  transcoderActivatedEvents(where: { delegate: "0xda43d85b8d419a9c51bbf0089c9bd5169c23f2f9" }) {id
    transaction {
      blockNumber
    }
    activationRound
    round {
      id
    }
  }
  transcoderDeactivatedEvents(where: { delegate: "0xda43d85b8d419a9c51bbf0089c9bd5169c23f2f9" }) {
    round {
      id
    }
    transaction {
      blockNumber
    }
    deactivationRound
  }
}
```

Notice that the `active` field for the transcoder is set to false when it should be set to true.

The sequence of events that results in the data returned from the above query:

- TranscoderActivated
    - block = 9276341
    - round = 1610
    - activationRound = 1611
- TranscoderDeactivated
    - block = 9325929
    - round = 1619
    - deactivationRound = 1620
- TranscoderActivated
    - block = 9326035
    - round = 1619
    - activationRound = 1620

adamsoffer:

Fixes #1027

### DIFF
--- a/packages/subgraph/src/mappings/bondingManager.ts
+++ b/packages/subgraph/src/mappings/bondingManager.ts
@@ -532,6 +532,14 @@ export function transcoderActivated(event: TranscoderActivated): void {
   transcoder.deactivationRound = MAXIMUM_VALUE_UINT256;
   transcoder.save();
 
+  // if transcoder was pending deactivation, remove it from deactivation array
+  let pendingDeactivation = protocol.pendingDeactivation;
+  if (pendingDeactivation.includes(event.params.transcoder.toHex())) {
+    let index = pendingDeactivation.indexOf(event.params.transcoder.toHex());
+    pendingDeactivation.splice(index, 1);
+    protocol.pendingDeactivation = pendingDeactivation;
+  }
+
   // Add transcoder to list of transcoders pending activation
   let pendingActivation = protocol.pendingActivation;
   pendingActivation.push(event.params.transcoder.toHex());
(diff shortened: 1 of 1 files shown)


yondonfu (approved):
LGTM

@adamsoffer can we test this change on the livepeer-canary subgraph after this is merged?

Would be great to test the following:

- Before deploying the change, querying for all cases of this bug i.e. by checking for transcoders that have `active: false` with `activationRound <= currentRound` and `deactivationRound > currentRound`
- Make sure all the reported cases in https://github.com/livepeer/livepeerjs/issues/1027 are resolved
- Make sure all cases from the first query are resolved as well

adamsoffer:
@yondonfu Yeah I deployed last night to `livepeer-canary`. It's 90% synced right now. When it's done syncing I'll make sure those tests are covered before deploying 👍

adamsoffer:
The subgraph has finished syncing and I've [tested the fix](https://codesandbox.io/s/subgraph-fix-test-u0zvt?file=/src/index.js) to confirm it works 👍 

The [main livepeer subgraph](https://thegraph.com/hosted-service/subgraph/livepeer/livepeer) on the graph's hosted explorer is now synced with the fix as well.

We'll want to update the subgraph on the graph's decentralized network at some point as well but we can address that in another issue.

PR #1048, opened 2022-02-01
fix: voting page

adamsoffer:
This PR fixes an issue preventing the voting page from rendering by making sure `poll.tally` is not null and if it is, initializing the the total vote stake to zero.


yondonfu (approved):
LGTM!

